PR #7, opened 2025-07-31
add a simple openai agent example

samhita-alla:


samhita-alla:
@kumare3 here’s a simple agent example with just tool calls, and it uses the openai agent sdk. this is an extension to your pb & j example. :)

examples/genai/openai_agent.py
```diff
@@ -0,0 +1,151 @@
...
+    return await get_jelly()
+
+
+@function_tool
+async def spread_peanut_butter_tool(bread: str, peanut_butter: str) -> str:
+    """Spread peanut butter on bread."""
+    return await spread_peanut_butter(bread, peanut_butter)
+
+
+@function_tool
```
kumare3, line 92: why do we have to do this?

kumare3: I mean create a shadow function, what do we need to do to make it seamless?

samhita-alla: @cosmicBboy and i were talking about this. `flyte.trace` would work seamlessly here, but a task wouldn’t, mainly because both the task and the OpenAI tool decorators materialize the function into a dataclass, so stacking them in either order breaks things.

so we can propose two approaches: one using a task and the other using a trace. ultimately, it’ll be up to the user to choose, depending on whether the tool needs to run in a separate runtime. i know this puts the decision on the user, but i think it’s the most flexible option, unless we want to enforce the use of traces only.

cosmicbboy: I think we should consider creating light-weight utility functions in the `flyte.ai` namespace that serve as drop-in replacements for popular frameworks, e.g.

```python
from flyte.ai.openai.agents import function_tool

@function_tool
@env.task
def my_tool(...): ...
```

kumare3: should this be a follow up?

PR #17, opened 2025-08-01
[WIP] Recursively deploying Environments.

kumare3:
Currently fails, as it cannot determine the `module` for the recursively discovered enviroments.

Example usage:
```bash
flyte deploy --all --recursive examples/basics
```

### DIFF
--- a/src/flyte/_deploy.py
+++ b/src/flyte/_deploy.py
@@ -4,6 +4,7 @@ import asyncio
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Dict, List, Optional, Tuple
 
+import grpc.aio
 import rich.repr
 
 import flyte.errors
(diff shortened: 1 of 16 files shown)


src/flyte/_utils/module_loader.py
```diff
@@ -76,6 +78,9 @@ def _load_module_from_file(file_path: Path) -> str | None:
 
         # Create and execute the module
         module = importlib.util.module_from_spec(spec)
+        sys.modules[module_name] = module
```
kumare3, line 81: interesting

src/flyte/_deploy.py
```diff
@@ -90,27 +92,39 @@ async def _deploy_task(
...
-        msg += f" from {spec.task_template.container.args[-3]}.{spec.task_template.container.args[-1]}"
-    logger.info(msg)
-    task_id = task_definition_pb2.TaskIdentifier(
-        org=spec.task_template.id.org,
-        project=spec.task_template.id.project,
-        domain=spec.task_template.id.domain,
-        version=spec.task_template.id.version,
-        name=spec.task_template.id.name,
-    )
+    try:
```
kumare3, line 95: Why are we catching this exception? we already catch rpc errors

PR #19, opened 2025-08-01
Updated Agent sim example and formatting

kumare3:


examples/genai/agent_simulation_loadtest.py
```diff
@@ -7,56 +7,58 @@
...
 
+tool_env = coordinator_env.clone_with(name="tool_env", reusable=flyte.ReusePolicy(replicas=8, idle_ttl=300))
+
 
 # Mock tools that research agents can use
-@flyte.trace
+@tool_env.task
 async def search_web(query: str) -> str:
-    await asyncio.sleep(1.0)  # Simulate API call
+    # await asyncio.sleep(1.0)  # Simulate API call
```
cosmicbboy, line 37: should we delete these sleep calls?

PR #24, opened 2025-08-04
[wip] Implement drop-in replacement for `@function_tool`

cosmicBboy:
This PR implements a `@function_tools` drop-in decorator for the openai `agents` sdk.

### DIFF
--- a/src/flyte/ai/openai/agents/_function_tools.py
+++ b/src/flyte/ai/openai/agents/_function_tools.py
@@ -1,0 +1,23 @@
+import typing
+
+from flyte._task import AsyncFunctionTaskTemplate
+from flyte._utils import lazy_module
+
+
+if typing.TYPE_CHECKING:
+    import agents
+    from agents import function_tool as openai_function_tool, FunctionTool
+else:
+    agents = lazy_module("agents")
+
+
+def function_tool(
+    func: AsyncFunctionTaskTemplate,
+) -> agents.FunctionTool:
+    import ipdb; ipdb.set_trace()
+    return agents.FunctionTool(
+        name=func.name,
+        description=func.description,
+        params_json_schema=func.params_json_schema,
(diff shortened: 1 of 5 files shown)


src/flyte/ai/openai/agents/_function_tools.py
```diff
@@ -0,0 +1,66 @@
...
+from flyte._utils import lazy_module
+
+
+if typing.TYPE_CHECKING:
+    import agents
+    from agents import function_tool as openai_function_tool, FunctionTool as OpenAIFunctionTool
+    from agents.function_schema import function_schema
+    from agents.tool_context import ToolContext
+else:
+    agents = lazy_module("agents")
```
kumare3: if agents is not installed does it raise an exception?

src/flyte/ai/openai/agents/_function_tools.py
```diff
@@ -0,0 +1,66 @@
...
+    openai_function_tool = agents.function_tool
+    OpenAIFunctionTool = agents.FunctionTool
+    function_schema = lazy_module("agents.function_schema").function_schema
+    ToolContext = lazy_module("agents.tool_context").ToolContext
+
+
+@dataclass
+class FlyteFunctionTool(OpenAIFunctionTool):
```
kumare3: nit can we rename it to FunctionTool? I do not like Flyte in it, as this is already inside Flyte. 
Also should this be a plugin or we make openai and anthropic first class

kumare3: and maybe gemini

src/flyte/ai/openai/agents/_function_tools.py
cosmicbboy: I think it should? not sure how we enforce lower/upper bound versions for soft dependencies

PR #26, opened 2025-08-04
Traces are throttled and are run on the background thread

kumare3:


src/flyte/_internal/controllers/remote/_controller.py
```diff
@@ -168,7 +168,7 @@ async def _submit(self, _task_call_seq: int, _task: TaskTemplate, *args, **kwarg
         code_bundle = tctx.code_bundle
 
         if code_bundle and code_bundle.pkl:
-            logger.debug(f"Building new pkl bundle for task {_task.name}")
+            logger.debug("Building new pkl bundle for task {_task.name}")
```
wild-endeavor: add back f?

wild-endeavor (approved):

PR #39, opened 2025-08-07
Plumb through trace spec interface

squiishyy:
Plumb through the trace spec information through the entire system


wild-endeavor (dismissed):
this is a breaking change on the backend right?

squiishyy:
> this is a breaking change on the backend right?

No. Just ran with most recent release (not including these changes) and BE works fine. Traces without IO do not get recorded but that is expected as it is broken there already. This addresses that as well

examples/advanced/local_tasks.py
```diff
@@ -64,7 +68,52 @@ async def parallel_main(q: str) -> list[str]:
     return r
 
 
+@flyte.trace
```
kumare3, line 71: can you also try 
`flyte get io`
or 
```
import flyte.remote as remote
remote.Action.listall(run=...)
```
And then 
```
action.details()
```
It will probably break, now that there are some additional information in action details

kumare3:
LGTM, one comment
and then can you merge from main

wild-endeavor (approved):

kumare3 (approved):

PR #44, opened 2025-08-07
adding log handler for file if stream is false for audit and compliance

### LINKED ISSUE #43: [Feature]: Adds file handler to the current logging interface.
Hi,

Currently there is no way to create log files and logs are only streamed.

We would like an approach where the user can decide whether they want to stream the logs or store the logs in a file for audit and compliance.

There also is some redundancy is the logging code which needs to be removed to make it clean.

Regards,

debasisdwivedy:
Fixes #43

### DIFF
--- a/src/flyte/_initialize.py
+++ b/src/flyte/_initialize.py
@@ -180,11 +180,13 @@ async def init(
     from flyte._utils import get_cwd_editable_install, org_from_endpoint, sanitize_endpoint
 
     interactive_mode = ipython_check()
-    
+
     if log_level and log_file_dir:
         initialize_logger(log_level=log_level, enable_rich=interactive_mode, stream_logs=stream_logs,log_file_dir=log_file_dir)
     elif log_level:
         initialize_logger(log_level=log_level, enable_rich=interactive_mode, stream_logs=stream_logs)
+    elif log_file_dir:
+        initialize_logger(enable_rich=interactive_mode, stream_logs=stream_logs,log_file_dir=log_file_dir)
     else:
         initialize_logger(enable_rich=interactive_mode, stream_logs=stream_logs)
 


src/flyte/_logging.py
```diff
@@ -82,27 +85,39 @@ def get_default_handler(log_level: int) -> logging.Handler:
...
     """
     logger = logging.getLogger(name)
     logger.setLevel(log_level)
     handler = None
     logger.handlers = []
-    if enable_rich:
-        handler = get_rich_handler(log_level)
-    if handler is None:
-        handler = get_default_handler(log_level)
+    if stream_logs:
```
kumare3, line 116: rich is disabled when running remotely in a container. you want to stream the logs to stdout/stderr in the container. Logging to file does not work. So please keep enable_rich. Becasue in the container, we still want to stream logs, but no rich, as most log systems need json

src/flyte/_initialize.py
```diff
@@ -176,11 +180,10 @@ async def init(
     from flyte._utils import get_cwd_editable_install, org_from_endpoint, sanitize_endpoint
 
     interactive_mode = ipython_check()
-
-    initialize_logger(enable_rich=interactive_mode)
-    if log_level:
-        initialize_logger(log_level=log_level, enable_rich=interactive_mode)
-
+    log_level = log_level if log_level else get_env_log_level()
```
kumare3: i would just defer all this logic to initialize logger. That way all logic is in one place

src/flyte/_logging.py
debasisdwivedy: It is the same code. Just a more pythonic way todo it.

`handler = get_rich_handler(log_level) if enable_rich else get_default_handler(log_level)`

We have `if enable_rich` in it, which basically checks if it is set and only then we call `get_rich_handler` else we call `get_default_handler`

src/flyte/_initialize.py
debasisdwivedy: Sure, makes sense.

I will make those changes and push

src/flyte/_logging.py
debasisdwivedy: Also one quich question related to this: Even in container we can provide a mount file location for file based logging. Isn't it a possiblity.

The container will have a local disk mounted , so even if its a container, we will have the file.

PR #52, opened 2025-08-08
implement json output format for flyte get commands

cosmicBboy:
With these changes, you could do something like:

```bash
flyte get run --output-format json --limit 3 | jq '.[].action.id.run.name' | xargs -I {} flyte abort run {}
```

Which will abort the last three runs.

This adds the `--output-format` flag to the following commands:
- `flyte get run`
- `flyte get task`
- `flyte get action`
- `flyte get secret`

### DIFF
--- a/src/flyte/cli/_get.py
+++ b/src/flyte/cli/_get.py
@@ -1,4 +1,6 @@
 import asyncio
+import json
+from dataclasses import asdict
 from typing import Tuple, Union
 
 import rich_click as click
@@ -8,6 +10,9 @@ from rich.pretty import pretty_repr
 import flyte.remote._action
 
 from . import _common as common
+
+
+OUTPUT_FORMATS = click.Choice(["table", "json"])
 
 
 @click.group(name="get")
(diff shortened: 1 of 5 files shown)


src/flyte/cli/_get.py
```diff
@@ -69,21 +83,31 @@ def run(cfg: common.CLIConfig, name: str | None = None, project: str | None = No
...
+        else:
+            console.print(pretty_repr(details))
     else:
-        console.print(common.get_table("Runs", Run.listall(), simple=cfg.simple))
+        details = list(Run.listall(limit=limit))
+        if output_format == "json":
+            details_json = [d.to_json_dict() for d in details]
+            console.print(json.dumps(details_json, indent=2))
+        else:
+            console.print(common.get_table("Runs", details, simple=cfg.simple))
```
kumare3, line 96: I think we should make the entire printing to be a common thing

src/flyte/remote/_run.py
```diff
@@ -57,9 +65,10 @@ async def listall(
             direction=(list_pb2.Sort.ASCENDING if sort_by[1] == "asc" else list_pb2.Sort.DESCENDING),
         )
         cfg = get_common_config()
+        i = 0
         while True:
             req = list_pb2.ListRequest(
-                limit=100,
+                limit=limit,
```
pingsutw, line 71: I think we can use `min(limit, 100)` here. If we try to fetch too many runs (1000+) in a single gRPC call, it might exceed the gRPC size limit.

PR #58, opened 2025-08-11
Example of loading memory

kumare3:


examples/reuse/reusable_with_mem.py
```diff
@@ -0,0 +1,71 @@
...
+
+from async_lru import alru_cache
+
+import flyte
+
+env_reuse = flyte.TaskEnvironment(
+    name="reusable_with_mem",
+    resources=flyte.Resources(memory="500Mi", cpu=1),
+    reusable=flyte.ReusePolicy(
+        replicas=1,  # Min of 2 replicas are needed to ensure no-starvation of tasks.
```
cosmicbboy: should this be `replicas=2` to match the comment?

kumare3: aah i wanted to test only one, let me change the comment

examples/reuse/reusable_with_mem.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        Simulate a long-running computation.
+        """
+        result = x * self.v
+        self.v = self.v + 1
+        return result
+
+
+# Create a singleton instance for the reusable environment
+_calculator = ValueCalculator()
```
cosmicbboy, line 45: there will be one of these per actor replica right? so the state will be independent across replicas?

kumare3: that is true, so the comment can be improved

PR #59, opened 2025-08-11
Triggers (alpha - sdk only)

kumare3:


cosmicBboy:
SDK design LGTM!

examples/triggers/named_triggers.py
```diff
@@ -0,0 +1,31 @@
...
+import flyte.trigger
+
+env = flyte.TaskEnvironment(
+    "named_triggers",
+)
+
+daily_trigger = flyte.trigger.Cron("daily", "0 * * * *")
+weekly_trigger = flyte.trigger.Cron("weekly", "0 0 * * 0", auto_activate=False)  # Every Sunday at midnight
+monthly_trigger = flyte.trigger.Cron("monthly", "0 0 1 * *", {"x": 10})  # First day of every month with x=10
+webhook_trigger = flyte.trigger.Webhook(
```
iaroslav-ciupin: will I be able to view all triggers on some UI page, select the one I'm interested in and explicitly invoke it manually(neither via Cron, neither via Webhook)?

kumare3: I dont think you can invoke a trigger? cc @jpvotta

iaroslav-ciupin: I think according to https://www.notion.so/PRD-Triggers-in-Flyte-v2-2468cc06513d800ea5b7e66f9273c274?source=copy_link#24d8cc06513d8050886eeab8cf7f1b00 we can

jpvotta: Added a section here to be super explicit: https://www.notion.so/PRD-Triggers-in-Flyte-v2-2468cc06513d800ea5b7e66f9273c274?source=copy_link#24f8cc06513d8017a0a7cbfb01492d41

src/flyte/trigger/_trigger.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def new(
+    name: str,
+    automation: AutomationType,
+    *,
+    description: str = "",
+    inputs: Dict[str, Any] | None = None,
+    env: Dict[str, str] | None = None,
+    interruptable: bool | None = None,
```
iaroslav-ciupin: [This doc](https://www.notion.so/PRD-Triggers-in-Flyte-v2-2468cc06513d800ea5b7e66f9273c274) suggests we can also override any field from RunSpec(so this includes `labels`, `annotations` and, in future, `service_account`, `security_context`, etc)

src/flyte/cli/_create.py
```diff
@@ -191,3 +191,83 @@ def config(
...
+    "--task-version",
+    type=str,
+    default="latest",
+    help="Version of the task to create a trigger for. Defaults to 'latest'.",
+    show_default=True,
+)
+@click.option(
+    "--schedule",
+    type=str,
+    default="* * * * *",
```
iaroslav-ciupin: I am not sure if "every minute" is a reasonable default.
Also, today we may have triggers without any automation(without a schedule) and in future we may have triggers with other kinds of automation. I would say "if schedule is specified, then create a trigger with automation of type schedule. Otherwise, create a trigger with no automation"

src/flyte/remote/_trigger.py
```diff
@@ -0,0 +1,166 @@
...
+        :param task_version: Optional version of the task to associate with the trigger.
+        """
+        ensure_client()
+        cfg = get_common_config()
+        triggers = []
+        for t in trigger:
+            triggers.append(
+                run_definition_pb2.TriggerSpec(
+                    name=trigger.name,
+                    automation=run_definition_pb2.CronAutomation(  # this should be a oneof the automation types
```
iaroslav-ciupin:
```suggestion
                    automation=run_definition_pb2.CronAutomation(  # this should be an optional oneof the automation types
```
❗

src/flyte/cli/_create.py
```diff
@@ -191,3 +191,83 @@ def config(
...
+)
+@click.pass_obj
+def trigger(
+    cfg: common.CLIConfig,
+    task_name: str,
+    name: str,
+    task_version: str = "latest",
+    trigger_time_var: str = "trigger_time",
+    auto_activate: bool = True,
+    schedule: str = "* * * * *",
```
iaroslav-ciupin: alternatively, instead of `schedule` user can specify `rate` as amount and units, e.g. "every 90 minutes" or "every 24 hours", etc. ?

src/flyte/cli/_create.py
kumare3: I agree. I am just going to drop the default

wild-endeavor (approved):

PR #65, opened 2025-08-12
Wheel changes

wild-endeavor:
Updated version of https://github.com/flyteorg/flyte-sdk/pull/36

This changes how python wheels are handled... currently we're just installing *whl, but this doesn't work for multi-arch wheels. pip knows how to find the correct architecture with the --find-links arg, but we need to tweak the command.  However, we want
* if there's a local wheel, we absolutely want to use it, even if there's a newer version on pypi.  So the first command doesn't install any dependencies and uses --no-index.
* Since we need the dependencies, we need a second command to install the rest of the deps from pypi.

Will need to update the remote builder as well.

> [!NOTE]
> Had to update the image identifier in the unit test because the layer definition string representation actually changed.


src/flyte/_internal/imagebuild/docker_builder.py
```diff
@@ -162,12 +162,32 @@ class PythonWheelHandler:
...
-        pip_install_args.extend(["/dist/*.whl"])
         secret_mounts = _get_secret_mounts_layer(layer.secret_mounts)
-        delta = UV_WHEEL_INSTALL_COMMAND_TEMPLATE.substitute(
-            PIP_INSTALL_ARGS=" ".join(pip_install_args), SECRET_MOUNT=secret_mounts
+
+        # First install: Install the wheel without dependencies using --no-deps
+        pip_install_args_no_deps = [
+            *pip_install_args,
+            *[
+                "--pre",
```
pingsutw: `--pre` should already in `pip_install_args`
```suggestion
```

PR #66, opened 2025-08-12
Translate pydantic to dataframe

wild-endeavor:
This turns the DataFrame object from a dataclass to a pydantic model, to match the File/Dir classes.

Changes:
* Update the base class to Pydantic
  * The class only has two args in the constructor uri and format.  The rest of the fields have been marked as private fields and are for internal use only.
  * Direct instantiation is still allowed but the examples don't show it and it's not used internally either.  Instead there's a wrapper function called `create_from` that among other things takes a native dataframe (like pandas) object.
* I tried to get rid of the `create_from` and implement a `from_local` similar to the File/Dir experience, and I think we should but it felt a bit trickier so I don't want to do that now.  The issue is that while we can do the upload, it's harder to construct the entire literal.  The DataFrame type (or any native dataframe type) takes additional arguments in the form of an `Annotated` annotation on the type.  This can add column information as well as the storage format, the latter of which can affect the encoder that's used (which in turn affects how the upload happens).  I don't think we should we should make the user declare the annotations twice - we could alternatively remove it from the type signature, and put everything in the `from_local` call, but this also doesn't work well when we're only working with types (as in, setting the interface for a task, which includes [column and format](https://github.com/flyteorg/flyte/blob/3b621c141fe8d0582e099a38e4eeb86881a1a4d5/flyteidl/protos/flyteidl/core/types.proto#L46) information).

Needs more cleanup following this PR
* Add a better example.
* Finalize constructor methods (remove metadata proto dependency and remove all usage of raw constructor).
* Update tests.


pingsutw (dismissed):

examples/basics/dataframe_usage.py
```diff
@@ -0,0 +1,171 @@
...
+async def create_raw_dataframe() -> pd.DataFrame:
+    """
+    Task returns RAW dataframe (pd.DataFrame).
+    UPLOAD happens at the end of the task, the Flyte Type Engine will do the upload from memory to blob store.
+    """
+    return pd.DataFrame(SAMPLE_DATA)
+
+
+@env.task
+async def create_wrapper_dataframe() -> DataFrame:
```
kumare3: i want ot stick to using flyte.io.DataFrame, that way it can be understood quickly

src/flyte/io/_dataframe/dataframe.py
```diff
@@ -48,15 +45,23 @@
...
-class DataFrame(SerializableType, DataClassJSONMixin):
+class DataFrame(BaseModel, SerializableType):
     """
     This is the user facing DataFrame class. Please don't confuse it with the literals.StructuredDataset
     class (that is just a model, a Python class representation of the protobuf).
     """
 
-    uri: typing.Optional[str] = field(default=None)
-    file_format: typing.Optional[str] = field(default=GENERIC_FORMAT)
+    uri: typing.Optional[str] = Field(default=None)
```
kumare3, line 54: @squiishyy / @wild-endeavor we should probably sync the work of the jsonification of literals with Pydantic models in the sdk?

src/flyte/io/_dataframe/dataframe.py
```diff
@@ -134,30 +139,48 @@ def columns(cls) -> typing.Dict[str, typing.Type]:
...
-        self._dataframe_type: Optional[DF] = None  # type: ignore
-        self._already_uploaded = False
+    ) -> DataFrame:
+        """
+        Internal constructor for type engine usage.
+        For user-facing APIs, use from_val(), from_existing_remote(), or new_remote().
+        """
+        instance = cls(uri=uri, format=format or GENERIC_FORMAT, **kwargs)
+        instance._raw_df = val
+        instance._metadata = metadata
```
kumare3, line 157: do we need to set this in other methods?

kumare3 (dismissed):

src/flyte/io/_dataframe/dataframe.py
wild-endeavor: i think we'll need to but don't want to add right now - want to play with the patterns here a bit more today.  i don't like that metadata is a pb object!

PR #70, opened 2025-08-13
Logs now has syncify and handles unavailable logs longer

kumare3:


src/flyte/cli/_get.py
```diff
@@ -201,7 +201,7 @@ def logs(
 
     async def _run_log_view(_obj):
         task = asyncio.create_task(
-            _obj.show_logs(
+            _obj.show_logs.aio(
```
wild-endeavor, line 204: await

wild-endeavor (approved):

src/flyte/cli/_get.py
kumare3: this is create task

PR #79, opened 2025-08-15
Check untyped functions in mypy

kumare3:
NOT YET READY, need to fix all the breaks


kumare3:
Not going to land in this shape — closing after a year. Worth reopening as a scoped mypy strictness pass if we want it.

PR #89, opened 2025-08-18
Add resource to env.task decorator

cosmicBboy:
This use case has come up in several Flyte 2.0 demo conversations so far: as a user, I want to be able to specify resources at a task-level, overriding the resources specified in the `TaskEnvironment`. This is effectively the same as `task.override`, where overriding resources is already available.

### DIFF
--- a/src/flyte/_task_environment.py
+++ b/src/flyte/_task_environment.py
@@ -134,6 +134,7 @@ class TaskEnvironment(Environment):
         _func=None,
         *,
         name: Optional[str] = None,
+        resources: Optional[Resources] = None,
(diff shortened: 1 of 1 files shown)


kumare3:
@cosmicBboy i am not in favor of this one.

Can we please talk about this

cosmicBboy:
> @cosmicBboy i am not in favor of this one.
> 
> Can we please talk about this

Sure let's chat

cosmicBboy:
Closing for now, as we need to solidify the behavior of regular TaskEnvironments vs reusable TaskEnvironments

PR #105, opened 2025-08-21
Updated HashMethod approach

wild-endeavor:
New approach to the `HashMethod` concept from flytekit.

## HashMethod
In `flytekit`, users could annotate output types of tasks with `HashMethod(...)` where the `...` was a lambda function that would be computed and then added into the `hash` field of the Flyte Literal message.  This `hash` field was then used in _downstream_ tasks to compute the cache key for discovery (i.e. whether to run the task or read from cache).

### New approach
A couple things:
* The fact that hashes were computed during outputs makes a lot less sense in the dynamic v2 approach.  Users can take the output of a Task, and modify the contents completely, before submitting it again to the next task.  There's no easy way to know if an object has been touched and if so meaningfully so the system can't reasonably know if a hash value computed or provided at output time is still relevant.
* In many instances, the inputs to cache-enabled Tasks will not be outputs of other tasks.  A parent task can generate inputs, and immediately pass them downstream.

so we're moving hash computation to the input side now.  However, since this concept was really only ever useful in flytekit for FlyteFile, FlyteDirectory, and StructuredDataset types, this pr adds this functionality directly on the types instead of relying on the `typing.Annotated` approach.  Also this only contains the change for File and Dir so far - DataFrame to be added later.

### Changes
* Removed the old HashMethod object and removed the calculate hash function in the type engine.
* Added a new HashMethod protocol with a standard PrecomputedValue implementation if you just want to provide a value.
* Updated constructors for File/Dir to take this object, either as is, or as a str in cases where I/O isn't relevant (e.g. `from_existing_remote`).
* Added I/O classes that capture and scan bytes as they are written/read.
* Updated the File class to have both hash and hash_method and Dir to just have hash.  The reason for not adding hash_method yet to Dir is just because the storage layer would need a much bigger change to support the scan and hash updating approach used for Files.  We will add this to Dir in the future if there is demand, or maybe remove from File if we find it's never used.
* Update cache key generation to respect Literals with a hash value, including cases where the Literal is a container type that has literals underneath that have hash values.  (Also, this change to only look at the literals rather than the whole Inputs object is needed as we added more fields to the Inputs object (like context)).


src/flyte/io/_dir.py
```diff
@@ -248,27 +250,62 @@ def download_sync(self, local_path: Optional[Union[str, Path]] = None) -> str:
         raise NotImplementedError("Sync download is not implemented for remote paths")
 
     @classmethod
-    async def from_local(cls, local_path: Union[str, Path], remote_path: Optional[str] = None) -> Dir[T]:
+    async def from_local(
+        cls, local_path: Union[str, Path], remote_path: Optional[str] = None, hash_method: Optional[HashMethod] = None
```
kumare3: can we also pass hashvalue?

src/flyte/io/_dir.py
```diff
@@ -248,27 +250,62 @@ def download_sync(self, local_path: Optional[Union[str, Path]] = None) -> str:
...
 
         Returns:
             A new Dir instance pointing to the uploaded directory
 
         Example:
             ```python
             remote_dir = await Dir[DataFrame].from_local('/tmp/data_dir/', 's3://bucket/data/')
+            # With a precomputed hash
+            from flyte.io._hashing_io import PrecomputedValue
+            hash_method = PrecomputedValue("abc123")
```
kumare3: why this instead of simple str?

src/flyte/io/_dir.py
```diff
@@ -248,27 +250,62 @@ def download_sync(self, local_path: Optional[Union[str, Path]] = None) -> str:
...
+            remote_dir = await Dir[DataFrame].from_local('/tmp/data_dir/', 's3://bucket/data/', hash_method=hash_method)
             ```
         """
         local_path_str = str(local_path)
         dirname = os.path.basename(os.path.normpath(local_path_str))
 
+        # Validate hash_method is PrecomputedValue if provided
+        hash_value = None
+        if hash_method is not None:
+            if not isinstance(hash_method, PrecomputedValue):
```
kumare3: i would just make this str and do the wrapping on your own

src/flyte/io/_dir.py
```diff
@@ -248,27 +250,62 @@ def download_sync(self, local_path: Optional[Union[str, Path]] = None) -> str:
...
+                raise ValueError("For Dir objects, only PrecomputedValue hash method is currently supported")
+            hash_value = hash_method.result()
+
         output_path = await storage.put(from_path=local_path_str, to_path=remote_path, recursive=True)
-        return cls(path=output_path, name=dirname)
+        return cls(path=output_path, name=dirname, hash=hash_value)
+
+    @classmethod
+    def from_existing_remote(cls, remote_path: str, known_hash_value: Optional[str] = None) -> Dir[T]:
```
kumare3:
```suggestion
    def from_existing_remote(cls, remote_path: str, dir_hash: Optional[str] = None) -> Dir[T]:
```

tests/flyte/io_types/test_files.py
```diff
@@ -94,13 +143,111 @@ async def my_task() -> File[pd.DataFrame]:
...
+        result = await File.from_local(local_path, remote_path)
+
+        assert result.path == remote_path
+        async with result.open() as f:
+            content = f.read()
+        content = content.decode("utf-8")
+        assert content == test_content
+
+
+@pytest.mark.sandbox
```
kumare3, line 179: why does it need to be sandbox, you could also use local right?

kumare3: i gues you have a local example, nm then

src/flyte/io/_dir.py
wild-endeavor: how about "dir_cache_key" and "file_cache_key"

tests/flyte/io_types/test_files.py
wild-endeavor: i have some tests marked as sandbox.  one day we'll get a v2 sandbox that we can run integration tests on.

examples/caching/content_based_caching.py
```diff
@@ -0,0 +1,121 @@
...
+    result1_dir = await process_directory_with_preset_hash(dir1)
+
+    # Create second file with DIFFERENT content but SAME hash
+    content2 = "Different content!\nThis is file version 2 with totally different text"
+
+    with tempfile.NamedTemporaryFile(mode="w", delete=False, suffix=".txt") as temp_file:
+        temp_file.write(content2)
+        temp_path2 = temp_file.name
+
+    file2 = await File.from_local(temp_path2, hash_method=PrecomputedValue(hash_value))  # Same hash!
```
kumare3: Why not just str?

src/flyte/io/_file.py
```diff
@@ -139,7 +144,7 @@ def schema_match(cls, incoming: dict):
 
     @classmethod
     @requires_initialization
-    def new_remote(cls) -> File[T]:
+    def new_remote(cls, file_cache_key: Optional[str] = None, hash_method: Optional[HashMethod] = None) -> File[T]:
```
kumare3:
```suggestion
    def new_remote(cls, file_cache_key: Optional[str] = None, hash_method: Optional[HashMethod | str] = None) -> File[T]:
```

pingsutw (approved):

PR #106, opened 2025-08-21
Rename `task.friendly_name` -> `task.short_name` May break for some

kumare3:


src/flyte/_initialize.py
```diff
@@ -260,8 +259,7 @@ async def init_from_config(
                     "user",
                     f"Configuration file '{path_or_config}' does not exist., current working directory is {Path.cwd()}",
                 )
-        else:
-            cfg = config.auto(path_or_config)
+        cfg = config.auto(path_or_config)
```
wild-endeavor: so we'll overwrite the cfg from the first if block?

kumare3: oops ya, good catch let me put back the else

wild-endeavor (approved):

PR #112, opened 2025-08-21
Add WanDB plugin with example

kevinliao852:
## Summary

This pull request introduces a WandB plugin that integrates Weights & Biases logging into the project.
The changes include adding the new plugin, updating dependencies, and providing an example demonstrating how to configure and use the plugin.

## Changes

Added wandb plugin for experiment tracking.
Integrated support for logging metrics and model artifacts.
Provided a working example to showcase usage.
Updated documentation with setup and usage instructions.

Add these dependencies:
examples/plugins/wandb_example.py: add flyteplugins-wandb to pip packages
plugins/wandb/pyproject.toml: add flyte as dependency for WandB plugin
plugins/wandb/README.md


examples/plugins/wandb_example.py
```diff
@@ -0,0 +1,73 @@
...
+from flyteplugins.wandb.tracking import WandbConfig
+
+import flyte
+
+WANDB_API_KEY = os.getenv("WANDB_API_KEY")
+
+task_env = flyte.TaskEnvironment(name="hello_wandb")
+
+
+wandb_env = flyte.TaskEnvironment(
```
kumare3, line 12: lets say that the environment wants an api-key as secret

examples/plugins/wandb_example.py
```diff
@@ -0,0 +1,73 @@
...
+    name="wandb_env",
+    plugin_config=WandbConfig(
+        wandb_api_key=WANDB_API_KEY,
```
kumare3, line 15: lets not add api key to the plugin config, as this will be in plain text, you can just take a secret name if you want

examples/plugins/wandb_example.py
```diff
@@ -0,0 +1,73 @@
...
+    y_pred = model.predict(X_test)
+    rmse = mean_squared_error(y_test, y_pred)
+
+    # Log custom metrics
+    wandb.log({"rmse": rmse})
+
+    return rmse
+
+
+@wandb_env.task
```
kumare3, line 59: I am confused with this example?

examples/plugins/wandb_example.py
```diff
@@ -0,0 +1,73 @@
...
+        y_train,
+        eval_set=[(X_train, y_train), (X_test, y_test)],
+    )
+
+    # Evaluate
+    y_pred = model.predict(X_test)
+    rmse = mean_squared_error(y_test, y_pred)
+
+    # Log custom metrics
+    wandb.log({"rmse": rmse})
```
kumare3, line 54: this task uses wandb, how do we get the api key and init,

examples/plugins/wandb_example.py
```diff
@@ -0,0 +1,73 @@
...
+import flyte
+
+WANDB_API_KEY = os.getenv("WANDB_API_KEY")
+
+task_env = flyte.TaskEnvironment(name="hello_wandb")
+
+
+wandb_env = flyte.TaskEnvironment(
+    name="wandb_env",
+    plugin_config=WandbConfig(
```
kumare3, line 14: I actually do not think wandb should be a plugin, it should be a decorator or a helper

plugins/wandb/src/flyteplugins/wandb/tracking.py
```diff
@@ -0,0 +1,77 @@
...
+    Actual Plugin that track
+    Attributes:
+        plugin_config (Wandb): Configuration for Weights & Biases tracking.
+        task_type (str): Type of the task, It should be "wandb".
+    """
+
+    plugin_config: WandbConfig
+    task_type: str = "wandb"
+
+    async def execute(self, *args, **kwargs) -> Any:
```
kumare3, line 43: I do not think it should be a plugin, example, folks can use wandb with ray, spark, distributed training etc

examples/plugins/wandb_example.py
kevinliao852: Thanks for pointing this out! You’re right.

examples/plugins/wandb_example.py
kevinliao852: I was inspired by the [flytekit-wandb plugin](https://github.com/flyteorg/flytekit/tree/master/plugins/flytekit-wandb)
, I modeled this approach similarly. I may have misunderstood the intended role of plugins here, and I’m happy to discuss if a decorator or helper would be a better fit.

PR #121, opened 2025-08-23
fix: Handle auto image build

pingsutw:
Automatically build the image if the image name is equal to `auto` in the editable mode

```python
env = flyte.TaskEnvironment(
    name="hello_world",
    resources=flyte.Resources(cpu=1, memory="1Gi"),
)
```

### DIFF
--- a/tests/flyte/test_union_run_basic.py
+++ b/tests/flyte/test_union_run_basic.py
@@ -39,9 +39,12 @@ async def test_task1_local_union_async():
 
 
 @pytest.mark.asyncio
+@mock.patch("flyte._deploy._build_image_bg")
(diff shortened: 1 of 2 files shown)


src/flyte/_initialize.py
```diff
@@ -248,21 +248,22 @@ async def init_from_config(
     import flyte.config as config
 
     cfg: config.Config
-    if path_or_config is None or isinstance(path_or_config, str):
-        # If a string is passed, treat it as a path to the config file
-        if root_dir and path_or_config:
-            cfg = config.auto(str(root_dir / path_or_config))
-        elif path_or_config and not Path(path_or_config).exists():
+    if path_or_config is None:
```
pingsutw, line 251: Simplfy the logic here

PR #126, opened 2025-08-27
Not copying ignore group file into docker image

popojk:
Currently when the user defines image spec with source folder like bellow example, the SDK will copy all files into temp dir and docker image:
```
Image.from_debian_base(install_flyte=False)
    .with_env_vars({"hello": "world1"})
    .with_source_folder(Path(__file__).parent / "copy_files")
```

However, files such as `.gitignore` files and those under `.cache/*` should be excluded.

Referring to the current code bundle default [IgnoreGroup](https://github.com/flyteorg/flyte-sdk/blob/main/src/flyte/_code_bundle/bundle.py#L137-L138) practice, the SDK should excludes files in the default IgnoreGroup instead of copying them into the image.

## How was this patch tested?
Unit tests

### DIFF
--- a/src/flyte/_internal/imagebuild/docker_builder.py
+++ b/src/flyte/_internal/imagebuild/docker_builder.py
@@ -6,7 +6,7 @@ import tempfile
 import typing
 from pathlib import Path
 from string import Template
-from typing import ClassVar, List, Optional, Protocol, Tuple, cast
+from typing import ClassVar, List, Optional, Protocol, cast
 
 import aiofiles
 import click
@@ -255,16 +255,14 @@ class DockerIgnoreHandler:
 
 class CopyConfigHandler:
     @staticmethod
-    def list_dockerignore(root_path: Optional[Path], layers: Tuple[Layer, ...]) -> List[str]:
+    def list_dockerignore(root_path: Optional[Path], docker_ignore_file_path: Optional[Path]) -> List[str]:
         patterns: List[str] = []
         dockerignore_path: Optional[Path] = None
         if root_path:
             dockerignore_path = root_path / ".dockerignore"
         # DockerIgnore layer should be first priority
-        for layer in layers:
-            if isinstance(layer, DockerIgnore):
-                dockerignore_path = Path(layer.path)
-                break
+        if docker_ignore_file_path:
+            dockerignore_path = docker_ignore_file_path
 
         # Return empty list if no .dockerignore file found
         if not dockerignore_path or not dockerignore_path.exists() or not dockerignore_path.is_file():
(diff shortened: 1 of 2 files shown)


src/flyte/_internal/imagebuild/docker_builder.py
```diff
@@ -255,7 +257,73 @@ async def handle(layer: DockerIgnore, context_path: Path, _: str):
...
+        """
+        copied_files = []
+        
+        # Check source path type and handle accordingly
+        if path_type == 0 and src_path.is_file():
+            # Handle single file copy
+            if not ignore_group or not ignore_group.is_ignored(src_path.relative_to(src_path.parent)):
+                dst_path.mkdir(parents=True, exist_ok=True)
+                shutil.copy(src_path, dst_path)
+                copied_files.append(dst_path.name)
```
kumare3: can you put an else and logger.debug? That way we can debug in case of issues

popojk: Hi @kumare3 . I ended up removing the entire code block you mentioned, since we can simply use the ignore parameter in shutil.copytree to skip files. This makes the code much cleaner.

src/flyte/_internal/imagebuild/docker_builder.py
```diff
@@ -264,18 +296,22 @@ async def handle(layer: CopyConfig, context_path: Path, dockerfile: str) -> str:
...
         dst_path.parent.mkdir(parents=True, exist_ok=True)
         abs_path = layer.src.absolute()
+
         if layer.src.is_file():
             # Copy the file
             shutil.copy(abs_path, dst_path)
         elif layer.src.is_dir():
             # Copy the entire directory
-            shutil.copytree(abs_path, dst_path, dirs_exist_ok=True)
+            docker_ignore_patterns = CopyConfigHandler.list_dockerignore(abs_path)
```
pingsutw: Instead of `abs_path`, can we use `_get_init_config().root_dir` here?

src/flyte/_internal/imagebuild/docker_builder.py
```diff
@@ -538,7 +577,7 @@ async def _build_image(self, image: Image, *, push: bool = True, dry_run: bool =
             )
 
             for layer in image._layers:
-                dockerfile = await _process_layer(layer, tmp_path, dockerfile)
+                dockerfile = await _process_layer(layer, tmp_path, dockerfile, image._layers)
```
pingsutw: Instead of passing all the layers to `_process_layer` to get the ignore file in the copy handler, can we get the dockerignore path at line 578

popojk: Hi @pingsutw . Thanks for the recommendation. It is fixed.

popojk: cc @machichima

PR #128, opened 2025-08-27
init_from_config looks for .flyte/config.yaml files from cwd to home

cosmicBboy:
As a user, I don't want to always think about config file paths. Often times I just want to make a single config for my repo, and be guaranteed that my flyte scripts use that config. If I need to override it, I can create a config in the current working directory of where I'm running the example from.

This PR makes the following changes:

1. `flyte create config` creates a config file in `./.flyte/config.yaml` by default
2. The `flyte get config` and `flyte.init_from_config` behavior is the same, for the most part, but adds the following logic: instead of looking at `./flyte/config.yaml` first, it will look for a `.flyte` directory in the current directory, then iteratively look at: `.`, `..`, etc..., until it reaches the home directory `~`.

### DIFF
--- a/examples/image/base_image.py
+++ b/examples/image/base_image.py
@@ -1,4 +1,5 @@
 from pathlib import Path
+import logging
 
 import flyte
 from flyte import Image
@@ -21,7 +22,7 @@ async def t1(data: str = "hello") -> str:
 
 
 if __name__ == "__main__":
-    flyte.init_from_config("../../config.yaml")
+    flyte.init_from_config(log_level=logging.INFO)
     run = flyte.run(t1, data="world")
     print(run.name)
     print(run.url)
--- a/src/flyte/cli/_create.py
+++ b/src/flyte/cli/_create.py
@@ -98,7 +98,7 @@ def secret(
     "-o",
     "--output",
     type=click.Path(exists=False, writable=True),
-    default=Path.cwd() / "config.yaml",
+    default=Path.cwd() / ".flyte" / "config.yaml",
     help="Path to the output directory where the configuration will be saved. Defaults to current directory.",
     show_default=True,
 )
@@ -146,6 +146,9 @@ def config(
     from flyte._utils import org_from_endpoint, sanitize_endpoint
 
     output_path = Path(output)
+
+    if not output_path.parent.exists():
+        output_path.parent.mkdir(parents=True)
 
     if output_path.exists() and not force:
         force = click.confirm(f"Overwrite [{output_path}]?", default=False)
--- a/src/flyte/config/_reader.py
+++ b/src/flyte/config/_reader.py
@@ -135,6 +135,23 @@ class ConfigFile(object):
         return self._yaml_config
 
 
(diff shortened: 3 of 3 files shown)


examples/image/base_image.py
```diff
@@ -21,7 +22,7 @@ async def t1(data: str = "hello") -> str:
 
 
 if __name__ == "__main__":
-    flyte.init_from_config("../../config.yaml")
+    flyte.init_from_config(log_level=logging.INFO)
```
kumare3, line 25: Honestly, I do not love this magic.
How about
```python
flyte.config.auto() or
cfg_file = flyte.config.find_config()
flyte.init_from_config(cfg_file)
```

kumare3: I like the idea of `.flyte` we could use it store local cache data and other data for runs. But, then we dont just want up - we look at current folder or provide path

cosmicbboy: > But, then we dont just want up - we look at current folder or provide path

At the same time I think the current behavior is kinda janky imo. You have to be aware of the current working directory and the relative path of the flyte script to the config file to get it to work properly... just look at our examples repo. you have a bunch of `../../../../config.yaml` paths that look pretty ugly.

cosmicbboy: > flyte.config.auto() or
> cfg_file = flyte.config.find_config()
> flyte.init_from_config(cfg_file)

This isn't gonna do anything. `flyte.init_from_config` already calls `config.auto()` with the current behavior.

kumare3: @cosmicBboy but, we can simply use `~/.flyte/config.yaml` pretty standard right

kumare3: i know, what i was saying was modify something, but you are right, dont want the magic to be behind the scenes

cosmicbboy: Okay, let's get rid of the magic.

The default auto behavior I'd like to see:
```
      2. If not specified, the config file is searched in the default locations.
            a. ./config.yaml if it exists  (current working directory)
            b. ./.flyte/config.yaml if it exists (current working directory) 👈 new
            c: `FLYTE_CONFIG` environment variable 👈 new
            b. `UCTL_CONFIG` environment variable ⏮️ I consider this legacy, for backward compat?
            c. `FLYTECTL_CONFIG` environment variable ⏮️ I consider this legacy, for backward compat?
            d. ~/.union/config.yaml if it exists
            e. ~/.flyte/config.yaml if it exists
```

Then we just assume that users run their tasks from some standard directory (e.g. the repo root)

Another idea is:

```
cfg = flyte.config.from_git_root()
flyte.init_from_config(cfg)
```

Where it just does a subprocess call `git rev-parse --show-toplevel` to get the top-level directory of the repo if it exists or None if it doesn't or if the `git` command is not available. This way, we can run all of our examples regardless of what the current working directory is.

kumare3: I like it. I actually like more of the git helpers, like `flyte.git.config_from_root` or `flyte.git.version_from_sha`, `flyte.git.path`

PR #131, opened 2025-08-28
Remove timestamp from the context.tar.gz

pingsutw:
It will make sure the context is only uploaded once if it doesn’t change.


wild-endeavor (approved):

kumare3:
Would this affect, layer caching? i hope it does not use the time to comper

pingsutw:
this is for the dataproxy.

PR #135, opened 2025-08-29
Fixed issues and tested

kumare3:


src/flyte/_environment.py
```diff
@@ -44,7 +42,7 @@ class Environment:
 
     name: str
     depends_on: List[Environment] = field(default_factory=list)
-    pod_template: Optional[Union[str, "V1PodTemplate"]] = None
+    pod_template: Optional[Union[str, "PodTemplate"]] = None
```
wild-endeavor: can remove the quotes if you want right?

wild-endeavor (approved):

PR #143, opened 2025-09-03
init_from_config supports pathlib.Path

cosmicBboy:


src/flyte/config/_reader.py
```diff
@@ -92,11 +92,11 @@ def read(self, cfg: typing.Optional["ConfigFile"] = None) -> typing.Optional[typ
 
 
 class ConfigFile(object):
-    def __init__(self, location: str):
+    def __init__(self, location: str | pathlib.Path):
```
wild-endeavor: i think we should leave this as str actually... this isn't user facing right?  if it's not user facing we should be prescriptive.

src/flyte/config/_reader.py
```diff
@@ -96,7 +96,7 @@ def __init__(self, location: str):
         """
         Load the config from this location
         """
-        self._location = location
+        self._location = str(location)
```
wild-endeavor:
```suggestion
        self._location = location
```

src/flyte/config/_reader.py
```diff
@@ -139,16 +139,22 @@ def resolve_config_path() -> pathlib.Path | None:
     """
     Config is read from the following locations in order of precedence:
     1. ./config.yaml if it exists
-    2. `UCTL_CONFIG` environment variable
-    3. `FLYTECTL_CONFIG` environment variable
-    4. ~/.union/config.yaml if it exists
-    5. ~/.flyte/config.yaml if it exists
+    2. ./.flyte/config.yaml if it exists
```
wild-endeavor: should we delete this block and move to config.auto()?  that comment block is now out of date right?

cosmicbboy: done

wild-endeavor (approved):

PR #144, opened 2025-09-03
Added example for using an existing ray cluster

chrismatteson:


kumare3 (approved):

kumare3:
There is another way of doing this - https://github.com/flyteorg/flytekit/blob/44af03fca6d2d6d85b0d94e2b8a017ce0d627bec/plugins/flytekit-ray/tests/test_ray.py#L122

cc @pingsutw is this supported in v2 as well i assume

examples/plugins/ray_existing_example.py
```diff
@@ -0,0 +1,61 @@
...
+@task_env.task()
+async def hello_ray():
+    await asyncio.sleep(20)
+    print("Hello from the Ray task!")
+
+
+@task_env.task
+async def hello_ray_nested(n: int = 3, cluster_address: str = "ray://localhost:10001") -> typing.List[int]:
+    # Get cluster address from environment variable
+    ray.init(address=cluster_address)
```
pingsutw, line 43: yes, it should just work

pingsutw (approved):

PR #145, opened 2025-09-03
DataFrame cleanup

wild-endeavor:
Further clean up on the flyte.io.DataFrame class.
* Added a simple user facing example.  Still need additional more complicated examples.
* Updated csv/pandas handler - better filename, working around an issue with obstore and csvs.
* Removed `metadata` and `format` from the `create_from` constructor - metadata shouldn't be used by the user in general, and format belongs in the type, not in the constructor.  If we have it in both places it gets confusing.
* Renamed `create_from` to `from_df`
* Split out tests into multiple files.


src/flyte/io/_dataframe/dataframe.py
```diff
@@ -786,7 +784,8 @@ async def to_literal(
             structured_dataset_type=expected.structured_dataset_type if expected else None
         )
 
-        fdf = DataFrame.create_from(val=python_val, metadata=meta)
+        fdf = DataFrame.create_from(val=python_val)
```
kumare3:
```suggestion
        fdf = DataFrame.from(val=python_val)
```

PR #147, opened 2025-09-03
Update default config path to ./.flyte/config.yaml, add `flyte.git` namespace to provide git utilities

cosmicBboy:
This PR:

- Changes the default path of `flyte create config` to `./.flyte/config.yaml`
- Adds a `flyte.git.config_from_root` utility function to get config files relative to a git project's root directory
- Expands string config paths with `~` to the user's home directory.

### DIFF
--- a/tests/user_api/test_initialize.py
+++ b/tests/user_api/test_initialize.py
@@ -236,6 +236,33 @@ task:
         call_kwargs = mock_init.aio.call_args[1]
         assert call_kwargs["root_dir"] == symlinked_root
 
+    @pytest.mark.parametrize(
+        "path_arg, expected_path",
+        [
+            ("~/.flyte/config.yaml", Path.home() / ".flyte" / "config.yaml"),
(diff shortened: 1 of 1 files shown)


kumare3:
Does `../../../config.yaml` also work?

src/flyte/config/_reader.py
```diff
@@ -139,6 +139,7 @@ def resolve_config_path() -> pathlib.Path | None:
     """
     Config is read from the following locations in order of precedence:
     1. ./config.yaml if it exists
+    2. ./.flyte/config.yaml if it exists
```
kumare3, line 142: do we not try ~/.flyte

kumare3 (approved):

src/flyte/config/_reader.py
cosmicbboy: we do check on line 153. The intent is that we change the default `config.yaml` location to `./.flyte/config.yaml` instead of `./config.yaml`

PR #160, opened 2025-09-09
create all types example

squiishyy:


cosmicBboy:
hey @squiishyy do you know if we support `Literal` types? This was a specific customer ask from EIDU

squiishyy:
How do you mean?

kumare3:
We support literal types, i will improve this example and check it in

wild-endeavor (approved):

PR #162, opened 2025-09-10
add get git root config to config auto

cosmicBboy:
This PR changes the `config.auto()` behavior to try to find the `<git_root>/.flyte/config.yaml` file, which makes the behavior between the SDK and CLI the same: the config file resolution will apply in both `flyte get config` and `flyte.init_from_config`

### DIFF
--- a/examples/advanced/cancel_tasks.py
+++ b/examples/advanced/cancel_tasks.py
@@ -34,7 +34,5 @@ async def main(n: int, f: float):
 
 
 if __name__ == "__main__":
-    import flyte.git
-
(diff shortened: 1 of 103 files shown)


src/flyte/_initialize.py
```diff
@@ -266,7 +266,7 @@ async def init_from_config(
     else:
         cfg = path_or_config
 
-    logger.debug(f"Flyte config initialized as {cfg}")
+    logger.warning(f"Flyte config initialized as {cfg}")
```
wild-endeavor: info?

PR #163, opened 2025-09-10
[Fix] Image Cache bug when python version different between local and remote

machichima:
## Description

When the image has the version that's different from local env (e.g. use python 3.10 image but use 3.12 in local env), the remote execution will fail as it cannot find the image.

## Changes

Exclude python version when creating image identifier hash. Change the structure of look up table into nested. Below is the example:

```sh
{'rRtF_Uo8Wabpz3pMlhjELA': {'3.10': '356633062068.dkr.ecr.us-east-2.amazonaws.com/union/playground:flyte-7a4ec1d4c17a82d32086f976606d13ab'}}
```

If the running env use different environment (e.g. 3.11), which is not in the look-up table, it will fall back to the first image, i.e. "3.10", in the dict


src/flyte/_image.py
```diff
@@ -435,7 +435,10 @@ def identifier(self) -> str:
         # Only get the non-None values in the Image to ensure the hash is consistent
         # across different SDK versions.
         # Layers can specify a _compute_identifier optionally, but the default will just stringify
-        image_dict = asdict(self, dict_factory=lambda x: {k: v for (k, v) in x if v is not None and k != "_layers"})
+        image_dict = asdict(
+            self,
+            dict_factory=lambda x: {k: v for (k, v) in x if v is not None and k not in ("_layers", "python_version")},
```
machichima, line 441: Exclude the python version when creating the identifier hash

src/flyte/_internal/runtime/task_serde.py
```diff
@@ -226,7 +227,13 @@ def _get_urun_container(
...
         )
     else:
-        img_uri = serialize_context.image_cache.image_lookup[image_id]
+        python_version_str = "{}.{}".format(sys.version_info.major, sys.version_info.minor)
+        version_lookup = serialize_context.image_cache.image_lookup[image_id]
+        if python_version_str in version_lookup:
+            img_uri = version_lookup[python_version_str]
+        else:
+            # Fallback: try to get any available version
+            img_uri = next(iter(version_lookup.values())) if version_lookup else task_template.image.uri
```
machichima: If the python version in execution environment is not found, we will get into here and fall back to use the first available image for other python version

machichima:
This will not work as expected in following example:

If there's already version 3.10 and 3.11 for the same image in the cache, then the user create image with python version 3.12 in version 3.11 env, then this build will be put into {3.11: "image_uri"}. When running, as 3.12 is not in the dict, our method will fall back to "the first one" in the dict, which is "3.10", but this is the wrong one

tests/flyte/test_image.py
```diff
@@ -163,4 +163,5 @@ def test_image_uri_consistency_for_uvscript():
     assert img.base_image == "python:3.12-slim-bookworm", "Base image should be python:3.12-slim-bookworm"
     # This value should work across python versions in CI because all values have been specified above and are hardcoded
     # Please don't change this value unless you are sure it's the right thing to do.
-    assert img.identifier == "jbBYppyNmLmWb13gy6ts8g", img._layers
+    assert img.identifier == "ymxz6JlYRNMt5gqSOuEcSw", img._layers
```
machichima, line 166: As we do not hash the python value, the hash will be different

src/flyte/_internal/runtime/task_serde.py
popojk: Should we add a logger here if fallback happend?

machichima: Yes, sorry I missed it. Added in https://github.com/flyteorg/flyte-sdk/pull/163/commits/5485fa2ee765a048c1e68c9f5196e08bc0828b6e

The output log will be like:

image

src/flyte/_internal/imagebuild/image_builder.py
```diff
@@ -273,10 +273,11 @@ def repr(self) -> typing.List[typing.List[Tuple[str, str]]]:
...
-                [
-                    ("Name", k),
-                    ("image", v),
-                ]
-            )
+            for py_version, image_uri in v.items():
+                tuples.append(
+                    [
+                        ("Name", f"{k} (py{py_version})"),
+                        ("image", image_uri),
```
pingsutw, line 280: It’s not backward compatible, right? If local uses the old SDK and the container uses the new SDK, it will fail since `v` is not a dict.

kumare3: Lets take this risk when we are in beta and go

pingsutw (approved):

machichima:
@wild-endeavor The test is added in the newest commit: https://github.com/flyteorg/flyte-sdk/pull/163/commits/30f5aad90188a340cdb88e1cb08fd1e9a1ef6c42

Thanks!

wild-endeavor (approved):
+1 let's merge after conflict is fixed

wild-endeavor (approved):

pingsutw (approved):

PR #166, opened 2025-09-11
add custom plaintext keyring for google colab

cosmicBboy:
This PR implements a low priority simple keyring service to support cases where keyring config doesn't exist (e.g. in google colab notebooks)

Example colab notebook: https://colab.research.google.com/drive/1gp4Opuh9u-GGLp55FxjXPrIH6cIFQtRT#scrollTo=blMgpBK4aPoG

### DIFF
--- a/src/flyte/_keyring/file.py
+++ b/src/flyte/_keyring/file.py
@@ -75,7 +75,7 @@ class SimplePlainTextKeyring(KeyringBackend):
         from flyte.config._reader import resolve_config_path
 
         config_path = resolve_config_path()
-        if str(config_path.parent) == ".flyte":
+        if config_path and str(config_path.parent) == ".flyte":
             # if the config is in a .flyte directory, use that as the path
             return config_path.parent / "keyring.cfg"
         # otherwise use the default path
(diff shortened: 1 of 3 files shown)


src/flyte/_keyring/file.py
```diff
@@ -0,0 +1,78 @@
+from base64 import decodebytes, encodebytes
+from configparser import ConfigParser, NoOptionError, NoSectionError
+from pathlib import Path
+from typing import Optional
+
+from keyring.backend import KeyringBackend
+from keyring.errors import PasswordDeleteError
+
+_FLYTE_KEYRING_PATH: Path = Path.home() / ".flyte" / "keyring.cfg"
```
kumare3, line 9: Can you just use the `config dir`?

cosmicbboy: which one? the one in the git repo? the one in the current work directory? the one in `HOME`?

keeping this at `HOME` path keeps things simple, esp. since this keyring backend is only used in specific circumstances (e.g. google colab). It's not immediately clear why this keyring path needs to be in the potentially various places where we look for `.flyte` directories.

kumare3: i was thinking, if we are setting `.flyte` why not make that the place where everything goes like keyring, local cache, raw data etc? if you want to change it, its easy to move right? But, your call really in this case.

PR #168, opened 2025-09-11
[FEAT] Enable Local cache

machichima:
## Description

Enable using local cache for local task execution.

## Changes

- Use SQLite for storing the cache, set the default cache path in `{config_dir}/local_cache`. `config_dir` is the path where we have the config installed
- Apply LocalTaskCache.get/set in local controller `submit()`, to read from cache before executing through `direct_dispatch` and store the result after execution
- Make `aiosqlite` an optional dependency that can be installed with `flyte[aiosqlite]`. If not installed, use sync `sqlite3` instead.
- Added unit test for checking if local cache can be get and set correctly

### DIFF
--- a/src/flyte/_cache/local_cache.py
+++ b/src/flyte/_cache/local_cache.py
@@ -13,7 +13,7 @@ from flyte._logging import logger
 from flyte._protos.workflow import run_definition_pb2
 from flyte.config import auto
 
-DEFAILT_CACHE_DIR = "~/.flyte"
+DEFAULT_CACHE_DIR = "~/.flyte"
 CACHE_LOCATION = "local-cache/cache.db"
 
 
@@ -33,7 +33,7 @@ class LocalTaskCache(object):
         if config.source:
             cache_dir = config.source.parent
         else:
-            cache_dir = Path(DEFAILT_CACHE_DIR).expanduser()
+            cache_dir = Path(DEFAULT_CACHE_DIR).expanduser()
 
         cache_path = cache_dir / CACHE_LOCATION
         # Ensure the directory exists


src/flyte/_cache/local_cache.py
```diff
@@ -0,0 +1,121 @@
...
+
+            LocalTaskCache._initialized = True
+
+    @staticmethod
+    async def clear():
+        """Clear all cache entries."""
+        if not LocalTaskCache._initialized:
+            await LocalTaskCache.initialize()
+
+        async with LocalTaskCache._session_factory() as session:  # type: ignore[misc]
```
machichima: the `# type: ignore` is needed as mypy will complaining that the `_session_factory` can be `None`, but it will be assigned after `initialize()`

src/flyte/_internal/controllers/_local_controller.py
```diff
@@ -81,31 +84,44 @@ async def submit(self, _task: TaskTemplate, *args, **kwargs) -> Any:
             raise flyte.errors.RuntimeSystemError("BadContext", "Task context not initialized")
 
         inputs = await convert.convert_from_native_to_inputs(_task.native_interface, *args, **kwargs)
-        serialized_inputs = inputs.proto_inputs.SerializeToString(deterministic=True)
+        inputs_hash = convert.generate_inputs_hash_from_proto(inputs.proto_inputs)
+        task_interface = transform_native_to_typed_interface(_task.interface)
 
         sub_action_id, sub_action_output_path = convert.generate_sub_action_id_and_output_path(
-            tctx, _task.name, serialized_inputs, 0
+            tctx, _task.name, inputs_hash, 0
```
machichima, line 91: the function requires `inputs_hash`: https://github.com/flyteorg/flyte-sdk/blob/a2a15d58b7101db2bce7b4ca340c9446d6e3d01c/src/flyte/_internal/runtime/convert.py#L434-L439

src/flyte/_run.py
```diff
@@ -436,7 +446,7 @@ async def _run_local(self, obj: TaskTemplate[P, R], *args: P.args, **kwargs: P.k
             code_bundle=None,
             output_path=self._metadata_path,
             run_base_dir=self._metadata_path,
-            version="na",
+            version=version if version else "na",
```
machichima: We need this version for calling `convert.generate_cache_key_hash `
https://github.com/flyteorg/flyte-sdk/blob/a2a15d58b7101db2bce7b4ca340c9446d6e3d01c/src/flyte/_internal/runtime/convert.py#L401-L408

tests/conftest.py
```diff
@@ -35,3 +38,18 @@ def dummy_serialization_context():
         output_path="s3://bucket/outputs/0/jfkljfa/0",
         root_dir=Path.cwd(),
     )
+
+
+@pytest_asyncio.fixture(autouse=True)
```
machichima, line 43: Some test will get into `LocalController.submit` (e.g. `test_run_hello_world`), therefor, we need to ensure that they are storing local cache in `tmp_path` so that tests are isolated

https://github.com/flyteorg/flyte-sdk/pull/168/files#diff-ef684937b6020a2b65e26c57c14b97285f0a436384c8e24f3d5b4d9d746f9f3fR77

machichima:
@pingsutw @wild-endeavor @kumare3 PTAL, thank you!
cc @popojk

src/flyte/_cache/local_cache.py
```diff
@@ -0,0 +1,121 @@
...
+from flyte._protos.workflow import run_definition_pb2
+
+CACHE_LOCATION = "~/.flyte/local-cache"
+
+
+class Base(DeclarativeBase):
+    pass
+
+
+class CachedOutput(Base):
```
pingsutw: Could we do `CachedOutput(DeclarativeBase)` instead? Any reason to create another base?

src/flyte/_cache/local_cache.py
pingsutw: Should we make `_session_factory` non-optional?

src/flyte/_internal/controllers/_local_controller.py
```diff
@@ -81,31 +84,44 @@ async def submit(self, _task: TaskTemplate, *args, **kwargs) -> Any:
...
+            )
+
+            if err:
+                exc = convert.convert_error_to_native(err)
+                if exc:
+                    raise exc
+                else:
+                    raise flyte.errors.RuntimeSystemError("BadError", "Unknown error")
+
+            # store into cache
```
pingsutw, line 141: Do we need to check if the cache is enabled

src/flyte/_internal/controllers/_local_controller.py
```diff
@@ -81,31 +84,44 @@ async def submit(self, _task: TaskTemplate, *args, **kwargs) -> Any:
...
-            run_base_dir=tctx.run_base_dir,
-        )
-        if err:
-            exc = convert.convert_error_to_native(err)
-            if exc:
-                raise exc
-            else:
-                raise flyte.errors.RuntimeSystemError("BadError", "Unknown error")
+        ignored_inputs = list(_task.cache.ignored_inputs) if isinstance(_task.cache, Cache) else []
+        out = await LocalTaskCache.get(_task.name, inputs_hash, inputs, task_interface, tctx.version, ignored_inputs)
```
pingsutw: Do we also need to check if the cache is enabled here

src/flyte/_run.py
pingsutw: I think we should get the version from `_task.cache.get_version()`

src/flyte/_internal/controllers/_local_controller.py
pingsutw:
```suggestion
        out = await LocalTaskCache.get(_task.name, inputs_hash, inputs, task_interface, _task.cache.get_version(), ignored_inputs)
```

src/flyte/_cache/local_cache.py
```diff
@@ -0,0 +1,121 @@
+from pathlib import Path
+
+from flyteidl.core import interface_pb2
+from sqlalchemy import Column, LargeBinary, String, delete
```
kumare3: :(, can we not use sqlalchemy. this will need a dependency. Sqlite can be used like a keyvalue store or just use shelve module

src/flyte/_cache/local_cache.py
```diff
@@ -0,0 +1,121 @@
...
+from sqlalchemy.ext.asyncio import AsyncEngine, AsyncSession, async_sessionmaker, create_async_engine
+from sqlalchemy.orm import DeclarativeBase
```
kumare3: lets not over complicate this - no orm please

src/flyte/_cache/local_cache.py
kumare3:
```python
import sqlite3

# Open or create database
conn = sqlite3.connect("kv.db")
cur = conn.cursor()

# Create a table if it doesn't exist
cur.execute("""
CREATE TABLE IF NOT EXISTS kv (
    key   TEXT PRIMARY KEY,
    value BLOB
)
""")

# Insert or update
def set_value(key, value):
    cur.execute("INSERT OR REPLACE INTO kv (key, value) VALUES (?, ?)", (key, value))
    conn.commit()

# Fetch value
def get_value(key):
    cur.execute("SELECT value FROM kv WHERE key = ?", (key,))
    row = cur.fetchone()
    return row[0] if row else None

# Delete key
def delete_value(key):
    cur.execute("DELETE FROM kv WHERE key = ?", (key,))
    conn.commit()


# Example usage
set_value("foo", "bar")
print(get_value("foo"))   # bar
delete_value("foo")
print(get_value("foo"))   # None
```

src/flyte/_internal/controllers/_local_controller.py
kumare3: We should not do this, please use the cache key logic like in remote controller. the Cache should be a simple
`get(cachekey)` and `put(cachkey, outputs_path)`
Make it as dumb as possible

pyproject.toml
```diff
@@ -26,6 +26,9 @@ dependencies = [
     "mashumaro",
     "dataclasses_json", # TODO: remove this dependency once we remove dataclasses_json from the code
     "aiolimiter>=1.2.1",
+    "aiosqlite>=0.21.0",
```
kumare3: no

pyproject.toml
```diff
@@ -26,6 +26,9 @@ dependencies = [
...
+    "greenlet>=3.2.4",
```
kumare3: OMG, please no

pyproject.toml
```diff
@@ -26,6 +26,9 @@ dependencies = [
...
+    "sqlalchemy>=2.0.43",
```
kumare3: no

kumare3:
@machichima / @pingsutw I have a related PR https://github.com/flyteorg/flyte-sdk/pull/207

src/flyte/_internal/controllers/_local_controller.py
machichima: Yes, I added in https://github.com/flyteorg/flyte-sdk/pull/168/commits/688e21eb80d0b59e66495bca999e337cb4225478 . Thank you!

src/flyte/_internal/controllers/_local_controller.py
machichima: Also fixed in https://github.com/flyteorg/flyte-sdk/pull/168/commits/688e21eb80d0b59e66495bca999e337cb4225478, thank you!

machichima: I change from sqlalchemy to sqlite and simply the `get`, `set` function in https://github.com/flyteorg/flyte-sdk/pull/168/commits/e9a6a09e0801eefc1033773d4633f484f73a9c5f. Thank you!

src/flyte/_cache/local_cache.py
```diff
@@ -0,0 +1,116 @@
...
+        output_bytes = value.proto_outputs.SerializeToString()
+
+        # NOTE: We will directly update the value in cache if it already exists
+        if existing:
+            await LocalTaskCache._conn.execute(
+                "UPDATE task_cache SET value = ? WHERE key = ?", (output_bytes, cache_key)
+            )
+        else:
+            await LocalTaskCache._conn.execute(
+                "INSERT INTO task_cache (key, value) VALUES (?, ?)", (cache_key, output_bytes)
```
pingsutw: Could we combine these two queries? Use `INSERT OR REPLACE INTO` instead?

src/flyte/_internal/controllers/_local_controller.py
```diff
@@ -81,31 +84,63 @@ async def submit(self, _task: TaskTemplate, *args, **kwargs) -> Any:
...
-            if exc:
-                raise exc
-            else:
-                raise flyte.errors.RuntimeSystemError("BadError", "Unknown error")
+
+        out = None
+        if cache_enabled:
+            out = await LocalTaskCache.get(cache_key)
+            if out is not None:
+                logger.info(
```
pingsutw: cache is enabled when the behavior is `auto` or `override`. I think we want to run the function here if the behavior is `override`.

pyproject.toml
pingsutw: cc @kumare3 @wild-endeavor, are we okay to add this, or should we use the sync version of SQLite?

src/flyte/_cache/local_cache.py
machichima: Sure! Updated in https://github.com/flyteorg/flyte-sdk/pull/168/commits/d953add88674c981a31afd40ead86b7443b55775

src/flyte/_internal/controllers/_local_controller.py
machichima: Fixed in https://github.com/flyteorg/flyte-sdk/pull/168/commits/49c11e5b150e750ff83eae4ce22b74cf95faff0b, thank you!

src/flyte/_cache/local_cache.py
```diff
@@ -0,0 +1,215 @@
...
+CACHE_LOCATION = "local-cache/cache.db"
+
+
+class LocalTaskCache(object):
+    """
+    This class implements a persistent store able to cache the result of local task executions.
+    """
+
+    _conn: "aiosqlite.Connection | None" = None
+    _conn_sync: sqlite3.Connection | None = None
```
machichima, line 26: I use `_conn` and  `_conn_sync` here. If we use `_conn: "aiosqlite.Connection | sqlite3.Connection | None" = None `, mypy willl complain about typing and we need to do type casting in all other functions before using `LocalTaskCache._conn`

tests/conftest.py
```diff
@@ -1,7 +1,10 @@
 from pathlib import Path
+from unittest.mock import patch
```
wild-endeavor, line 2: could we not use unittest?  we've only ever used pytest... is there a difference between mock and unittest.mock?

src/flyte/_cache/local_cache.py
```diff
@@ -0,0 +1,215 @@
...
+    HAS_AIOSQLITE = True
+except ImportError:
+    HAS_AIOSQLITE = False
+
+from flyte._internal.runtime import convert
+from flyte._logging import logger
+from flyte._protos.workflow import run_definition_pb2
+from flyte.config import auto
+
+DEFAILT_CACHE_DIR = "~/.flyte"
```
wild-endeavor: DEFAULT

src/flyte/_internal/controllers/remote/_controller.py
```diff
@@ -377,9 +377,10 @@ async def get_action_outputs(
         invoke_seq_num = self.generate_task_call_sequence(_func, current_action_id)
         inputs = await convert.convert_from_native_to_inputs(_interface, *args, **kwargs)
         serialized_inputs = inputs.proto_inputs.SerializeToString(deterministic=True)
+        inputs_hash = convert.generate_inputs_hash_from_proto(inputs.proto_inputs)
 
         sub_action_id, sub_action_output_path = convert.generate_sub_action_id_and_output_path(
-            tctx, func_name, serialized_inputs, invoke_seq_num
+            tctx, func_name, inputs_hash, invoke_seq_num
```
wild-endeavor, line 383: nice

wild-endeavor (approved):
how does local cache work on a task that doesn't produce an output?  it'll run again and again right?  could we add a unit test to codify the behavior?

(in prod there is no cache on tasks with no outputs, so running again is correct, just wondering if we should ensure that's the case with a test)

machichima:
> how does local cache work on a task that doesn't produce an output? it'll run again and again right? could we add a unit test to codify the behavior?

Yes, based on here: https://github.com/flyteorg/flyte-sdk/pull/168/files#diff-ef684937b6020a2b65e26c57c14b97285f0a436384c8e24f3d5b4d9d746f9f3fR142-R143

We will not store into cache if the output is `None`. Let me add a test case for this

tests/conftest.py
machichima: I think `unittest.mock` is also used in other places: https://github.com/flyteorg/flyte-sdk/blob/ca9d4b37662af6380d2b6955741e8ff4e54eba41/tests/cli/test_create.py#L1-L1
And I don't think pytest has mock method

src/flyte/_cache/local_cache.py
machichima: oops, sorry for the typo. Fixed in https://github.com/flyteorg/flyte-sdk/pull/168/commits/53c50f05ed8f2462d9fd8fb881bca9332dfc89a6

wild-endeavor (approved):

PR #170, opened 2025-09-12
Support Literal type (as enums) and Enums in dataclasses and pydantic models

kumare3:
```python
Intensity = Literal["low", "medium", "high"]

@env.task
async def literal_echo(i: Intensity) -> Intensity:
    return i


@env.task
async def literal_task(i: Intensity) -> str:
    return f"Intensity is {i}"
```

Works correctly

### DIFF
--- a/tests/user_api/test_interface.py
+++ b/tests/user_api/test_interface.py
@@ -141,7 +141,7 @@ def test_native_interface_int_literal():
     if sys.version_info >= (3, 11):
         assert interface.__repr__() == "(i: Any) -> (o0: Any, o1: str):"
     else:
-        assert interface.__repr__() == "(i: typing.Any) -> (o0: typing.Any, o1: str):"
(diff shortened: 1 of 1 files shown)


tests/user_api/test_interface.py
```diff
@@ -104,3 +105,23 @@ async def test_native_interface_with_union_type() -> None:
     assert repr == "(x: int | str) -> o0: int:"
     assert interface.inputs == {"x": (int | str, inspect.Parameter.empty)}
     assert interface.outputs == {"o0": int}
+
+
+Intensity = Literal["low", "medium", "high"]
```
cosmicbboy, line 111: can we add a test case for other types (int, float), and mixed types in the `Literal`? Like `Literal["low", 0, 3.14]`

kumare3: Like they should fail?
We do not support multi variate types. I have used enum as the archetype

PR #171, opened 2025-09-12
update logger rich styling to work on terminal and jupyter

cosmicBboy:
This PR updates `logging` and `click.secho` statements to use rich so that it all looks good on a terminal and jupyter notebook.

Terminal:
image

Jupyter notebook (before):
image

Jupyter notebook (after):
image

### DIFF
--- a/src/flyte/_initialize.py
+++ b/src/flyte/_initialize.py
@@ -110,6 +110,15 @@ async def _initialize_client(
     )
 
 
+def _initialize_logger(log_level: int | None = None):
+    from flyte._tools import ipython_check
+
+    # interactive_mode = not ipython_check()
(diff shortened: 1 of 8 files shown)


src/flyte/_initialize.py
```diff
@@ -110,6 +110,15 @@ async def _initialize_client(
     )
 
 
+def _initialize_logger(log_level: int | None = None):
+    from flyte._tools import ipython_check
+
+    # interactive_mode = not ipython_check()
```
pingsutw: Should we revert this line?

cosmicbboy: ah, good catch

kumare3 (approved):

PR #174, opened 2025-09-13
Add PoetryProject and update IDL

EngHabu:


tests/flyte/test_image.py
```diff
@@ -188,4 +188,3 @@ def test_ids_for_different_python_version():
     object.__setattr__(ex_11, "base_image", "python:3.10-slim-bookworm")
     object.__setattr__(ex_12, "base_image", "python:3.10-slim-bookworm")
     assert ex_10.identifier == ex_11.identifier
-    assert ex_11.identifier == ex_12.identifier
```
wild-endeavor: why did this change?  can you pull again?  this test shouldn't be relevant anymore after https://github.com/flyteorg/flyte-sdk/pull/211 just got merged

wild-endeavor (commented):
one nit and then good to merge

PR #178, opened 2025-09-15
Proto to flyteidl2

EngHabu:


pyproject.toml
```diff
@@ -39,6 +40,9 @@ flyte_keyring_file = "flyte._keyring.file"
 requires = ["setuptools", "setuptools_scm"]
 build-backend = "setuptools.build_meta"
 
+[[tool.uv.index]]
+url = "https://buf.build/gen/python"
```
wild-endeavor: what does this do?

examples/advanced/local_tasks.py
```diff
@@ -3,9 +3,13 @@
 
 import flyte
 
+idl2 = "git+https://github.com/flyteorg/flyte.git@enghabu/rename-idl2#subdirectory=gen/python"
```
wild-endeavor: Can we move this to a contributor guide?  this is good to have.  @pingsutw @kumare3 maybe we even add this as an argument to with_local_v2

examples/reuse/task_noop.py
```diff
@@ -0,0 +1,25 @@
+import logging
```
wild-endeavor: let's remove this file and add a separate example for name usage if there isn't one already?

pyproject.toml
EngHabu: There are two deps in flyteidl2 that come from a different registry.... I'm trying to get rid of them because people who install `flyte` will now how to add this `--extra-index-url` which isn't convenient...

examples/advanced/local_tasks.py
pingsutw: We can test it via `with_uv_project` instead.

machichima: So what I should do here is:
1. add the git hash usage to contributor guide
2. test if `with_uv_project` can also use this
3. A follow-up PR adding args in `with_local_v2`

Can I do those in follow-up PRs so that we can get this merge first.

machichima: I actually forgot why there's this change, I think I'm just trying things out. Should we keep this change?

examples/reuse/task_noop.py
machichima: Sorry I think this should be put into https://github.com/flyteorg/flyte-sdk/pull/190. let me remove this

examples/advanced/local_tasks.py
pingsutw: > Can I do those in follow-up PRs so that we can get this merge first.

Yes, let's just do this in the separate PR

pingsutw: Actually, we should just cut a new flyteidl2 release first and use `flyte.Image.from_debian_base()` here

PR #180, opened 2025-09-15
Fix deployment versioning with environment and image cache hash

pingsutw:
## Summary
- Fixed deployment versioning to include environment and image cache hashes in version calculation
- Added hashlib and cloudpickle imports for proper hash computation
- Updated version calculation logic to compute MD5 hash from deployment environments, code bundle version, and image cache when no explicit version is provided

## Test plan
- [ ] Test deployment with no explicit version set
- [ ] Verify version changes when environment or image dependencies change
- [ ] Ensure backward compatibility with explicit version setting

### DIFF
--- a/src/flyte/_deploy.py
+++ b/src/flyte/_deploy.py
@@ -1,7 +1,10 @@
 from __future__ import annotations
 
 import asyncio
+import hashlib
 import typing
+
+import cloudpickle
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Dict, List, Optional, Tuple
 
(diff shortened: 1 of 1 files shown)


src/flyte/_deploy.py
```diff
@@ -175,15 +178,18 @@ async def apply(deployment_plan: DeploymentPlan, copy_style: CopyFiles, dryrun:
...
         code_bundle = await build_code_bundle(from_dir=cfg.root_dir, dryrun=dryrun, copy_style=copy_style)
-        version = version or code_bundle.computed_version
-        # TODO we should update the version to include the image cache digest and code bundle digest. This is
-        # to ensure that changes in image dependencies, cause an update to the deployment version.
-        # TODO Also hash the environment and tasks to ensure that changes in the environment or tasks
+        if deployment_plan.version:
+            version = deployment_plan.version
+        else:
+            h = hashlib.md5()
+            h.update(cloudpickle.dumps(deployment_plan.envs))
```
kumare3, line 189: pickle may not be deterministic

pingsutw: Just tested it locally 10 times, it is deterministic

kumare3 (approved):

PR #186, opened 2025-09-16
Stress large runs

kumare3:
Runs a large number of runs sequentially to callibrate single threaded create run speed and observed latency.

### DIFF
--- a/examples/stress/runs_per_second.py
+++ b/examples/stress/runs_per_second.py
@@ -49,7 +49,7 @@ async def runs_per_second(max_rps: int = 50, n: int = 500):
     timestamps = []
 
     # Create report tabs
-    overview_tab = flyte.report.get_tab("Overview")
+    overview_tab = flyte.report.get_tab("main")
     performance_tab = flyte.report.get_tab("Performance Charts")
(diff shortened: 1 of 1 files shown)


examples/stress/runs_per_second.py
```diff
@@ -0,0 +1,100 @@
...
+
+        run_end = time.time()
+        run_times.append(run_end - run_start)
+
+        # Print progress every 50 runs
+        if (i + 1) % 50 == 0:
+            elapsed = run_end - start_time
+            current_rps = (i + 1) / elapsed
+            avg_run_time = sum(run_times[-50:]) / min(50, len(run_times))
+            print(f"Runs {i + 1:3d}: {current_rps:6.2f} RPS | Avg run time: {avg_run_time * 1000:6.2f}ms")
```
EngHabu: calls for a report to be produced with a beautiful graph :D

EngHabu (approved):

examples/stress/runs_per_second.py
kumare3: Aah let me do that

EngHabu (approved):

PR #190, opened 2025-09-17
[Feat] Enable setting image in cli

machichima:
## Description

In v1, user can set the image in cli using `pyflyte run --image image1=imageuri`, we should be able to support the same thing in v2.

## Changes

- add `--image` in flyte cli for setting image
- add `Image.from_name` to enable using image set in cli
- add test to ensure the image set in cli can be used by `Image.from_name`

### DIFF
--- a/src/flyte/_image.py
+++ b/src/flyte/_image.py
@@ -564,6 +564,17 @@ class Image:
         return img
 
     @classmethod
+    def from_name(cls, name: str) -> Image:
+        from ._initialize import _get_init_config
+
+        cfg = _get_init_config()
+        if name in cfg.images:
+            img = cls._new(name=name, base_image=cfg.images[name])
+        else:
+            img = cls._new()
+        return img
+
+    @classmethod
     def from_uv_script(
         cls,
         script: Path | str,
--- a/src/flyte/_image.py
+++ b/src/flyte/_image.py
@@ -568,7 +568,7 @@ class Image:
         from ._initialize import _get_init_config
 
         cfg = _get_init_config()
-        if name in cfg.images:
+        if cfg and name in cfg.images:
             img = cls._new(name=name, base_image=cfg.images[name])
         else:
             img = cls._new()
--- a/src/flyte/_initialize.py
+++ b/src/flyte/_initialize.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import functools
 import threading
 import typing
-from dataclasses import dataclass, replace
+from dataclasses import dataclass, field, replace
 from pathlib import Path
 from typing import TYPE_CHECKING, Callable, List, Literal, Optional, TypeVar
 
@@ -42,7 +42,7 @@ class _InitConfig(CommonInit):
     client: Optional[ClientSet] = None
     storage: Optional[Storage] = None
     image_builder: "ImageBuildEngine.ImageBuilderType" = "local"
-    images: typing.Dict[str, Image] | None = None
+    images: typing.Dict[str, str] = field(default_factory=dict)
 
     def replace(self, **kwargs) -> _InitConfig:
         return replace(self, **kwargs)
@@ -230,7 +230,7 @@ async def init(
             org=org or org_from_endpoint(endpoint),
             batch_size=batch_size,
             image_builder=image_builder,
-            images=images,
+            images=images or {},
             source_config_path=source_config_path,
         )
 
--- a/src/flyte/_initialize.py
+++ b/src/flyte/_initialize.py
@@ -7,6 +7,7 @@ from dataclasses import dataclass, replace
 from pathlib import Path
 from typing import TYPE_CHECKING, Callable, List, Literal, Optional, TypeVar
 
+from flyte._image import Image
 from flyte.errors import InitializationError
 from flyte.syncify import syncify
 
@@ -41,6 +42,7 @@ class _InitConfig(CommonInit):
     client: Optional[ClientSet] = None
     storage: Optional[Storage] = None
     image_builder: "ImageBuildEngine.ImageBuilderType" = "local"
+    images: typing.Dict[str, Image] | None = None
 
     def replace(self, **kwargs) -> _InitConfig:
         return replace(self, **kwargs)
@@ -141,6 +143,7 @@ async def init(
     storage: Storage | None = None,
     batch_size: int = 1000,
     image_builder: ImageBuildEngine.ImageBuilderType = "local",
+    images: typing.Dict[str, Image] | None = None,
     source_config_path: Optional[Path] = None,
 ) -> None:
     """
@@ -177,6 +180,7 @@ async def init(
     :param batch_size: Optional batch size for operations that use listings, defaults to 1000, so limit larger than
       batch_size will be split into multiple requests.
     :param image_builder: Optional image builder configuration, if not provided, the default image builder will be used.
+    :param images: Optional dict of images that can be used by referencing the image name.
     :param source_config_path: Optional path to the source configuration file (This is only used for documentation)
     :return: None
     """
@@ -226,6 +230,7 @@ async def init(
             org=org or org_from_endpoint(endpoint),
             batch_size=batch_size,
             image_builder=image_builder,
+            images=images,
             source_config_path=source_config_path,
         )
 
@@ -294,6 +299,7 @@ async def init_from_config(
         root_dir=root_dir,
         log_level=log_level,
         image_builder=cfg.image.builder,
+        images=cfg.image.images,
         storage=storage,
         source_config_path=cfg_path,
     )
--- a/src/flyte/cli/_run.py
+++ b/src/flyte/cli/_run.py
@@ -181,12 +181,10 @@ class RunTaskCommand(click.RichCommand):
         for value in values:
             if "=" in value:
                 image_name, image_uri = value.split("=", 1)
-                image = Image.from_base(image_uri)
-                cfg.images[image_name] = image
+                cfg.images[image_name] = image_uri
             else:
                 # If no name specified, use "default" as the name
-                image = Image.from_base(value)
-                cfg.images["default"] = image
+                cfg.images["default"] = value
 
     def get_params(self, ctx: click.Context) -> List[click.Parameter]:
         # Note this function may be called multiple times by click.
--- a/src/flyte/cli/_run.py
+++ b/src/flyte/cli/_run.py
@@ -12,7 +12,9 @@ import rich_click as click
 from rich.console import Console
 from typing_extensions import get_args
 
+from .. import Image
 from .._code_bundle._utils import CopyFiles
+from .._initialize import _get_init_config
 from .._task import TaskTemplate
 from ..remote import Run
 from . import _common as common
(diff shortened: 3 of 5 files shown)


machichima:
Hi @kumare3 

I would like to confirm some user experience regarding setting `--image` in `flyte run` cli.

The use case I think of is:

- User set the name of the image they want to use by `Image.from_name("some_image")` and use
it in `TaskEnvironment`
- When executing run, they can do `flyte run --image some_image=some_image_uri`. Then the
  "some_image_uri" will be pluged into `Image.from_name("some_image")` as the base image.
- If the image name is not set in the cli, i.e. `flyte run --image default_image_uri`,
then this image will be regard as a default image

Do you think this user experience is ok?


Moreover, regarding the default image, I think for now if the `image` is not set in the `Environment`, the image will be "auto", then we will use `from_debian_base`. Should we use the default image if it is set in the cli instead?


Thank you!

kumare3:
> Hi @kumare3
> 
> I would like to confirm some user experience regarding setting `--image` in `flyte run` cli.
> 
> The use case I think of is:
> 
> * User set the name of the image they want to use by `Image.from_name("some_image")` and use
>   it in `TaskEnvironment`
> * When executing run, they can do `flyte run --image some_image=some_image_uri`. Then the
>   "some_image_uri" will be pluged into `Image.from_name("some_image")` as the base image.
> * If the image name is not set in the cli, i.e. `flyte run --image default_image_uri`,
>   then this image will be regard as a default image
> 
> Do you think this user experience is ok?
> 
> Moreover, regarding the default image, I think for now if the `image` is not set in the `Environment`, the image will be "auto", then we will use `from_debian_base`. Should we use the default image if it is set in the cli instead?
> 
> Thank you!

I think `--image` should be a multi-value, i guess you can create multiple image named.
Also, if the image name is not found, it should fail cleanly, explaining the reason
We should also support --image-config `yaml with all images listed as name: image-uri`

kumare3:
How will the URI work? will we just look for the image name in the image-cache?

src/flyte/config/_config.py
```diff
@@ -159,8 +161,16 @@ def auto(cls, config_file: typing.Optional[typing.Union[str, ConfigFile]] = None
         config_file = get_config_file(config_file)
         kwargs: typing.Dict[str, typing.Any] = {}
         kwargs = set_if_exists(kwargs, "builder", _internal.Image.BUILDER.read(config_file))
+        # TODO: We can add support for setting images in the config file
         return ImageConfig(**kwargs)
 
+    @classmethod
+    def add_image(cls, name: str, image: Image) -> None:
```
kumare3: Why do we need this?

src/flyte/_image.py
```diff
@@ -563,6 +563,17 @@ def from_base(cls, image_uri: str) -> Image:
         img = cls._new(base_image=image_uri)
         return img
 
+    @classmethod
+    def from_name(cls, name: str) -> Image:
+        from ._initialize import _get_init_config
+
+        cfg = _get_init_config()
```
kumare3: This will fail on server and will fail on module load. We should not do this here. Only the name should be captured - maybe as the image?

kumare3: Or better yet create a separate object?

machichima:
> How will the URI work? will we just look for the image name in the image-cache?

It will be the full URI of the image, just like setting with `from_base`, but we just enable things like templating. User can just reference the image by name, then when running task through cli, they can set different uri for the image in different runs

src/flyte/config/_config.py
machichima: Oh sorry this is not used. I forgot to remove this

src/flyte/_image.py
machichima: Sure! I think I will create a new object `ImageConfig` in `src/flyte/_initialize.py`

src/flyte/_initialize.py
```diff
@@ -170,13 +173,13 @@ async def init(
     :param ca_cert_file_path: [optional] str Root Cert to be loaded and used to verify admin
     :param http_proxy_url: [optional] HTTP Proxy to be used for OAuth requests
     :param rpc_retries: [optional] int Number of times to retry the platform calls
-    :param audience: oauth2 audience for the token request. This is used to validate the token
```
machichima, line 173: This parameter does not exists

src/flyte/_initialize.py
```diff
@@ -238,7 +241,7 @@ async def init(
             else:
                 logger.info("No editable install found, using current working directory as root directory.")
                 root_dir = Path.cwd()
-        root_dir = root_dir or get_cwd_editable_install() or Path.cwd()
```
machichima, line 241: Logic here is redundant as `root_dir` will not be None here. This line will always be `root_dir = root_dir`

src/flyte/cli/_run.py
```diff
@@ -108,6 +109,17 @@ class RunArguments:
             )
         },
     )
+    image: List[str] = field(
+        default_factory=list,
+        metadata={
+            "click.option": click.Option(
+                ["--image"],
+                type=str,
+                multiple=True,
```
kumare3, line 117: you can have a callback here, that can parse_images. https://click.palletsprojects.com/en/stable/advanced/#id2

src/flyte/cli/_run.py
```diff
@@ -170,6 +183,19 @@ async def _run():
 
         asyncio.run(_run())
 
+    @staticmethod
+    def parse_images(values: List[str]) -> None:
```
kumare3: this should be in common, so that you can use it in run and deploy

src/flyte/cli/_run.py
machichima: `parse_images` can only be used after `_initialize_config ` in `invoke()`, so that we cannot use it in callback here

https://github.com/flyteorg/flyte-sdk/pull/190/files#diff-db6fcfb5640e750d38d2804766e92759199a666731872af457a4d7133fa59afdR144

src/flyte/_deploy.py
```diff
@@ -217,16 +217,34 @@ async def _build_images(deployment: DeploymentPlan) -> ImageCache:
...
+    from ._initialize import _get_init_config
     from ._internal.imagebuild.image_builder import ImageCache
 
     images = []
     image_identifier_map = {}
+    cfg = _get_init_config()
     for env_name, env in deployment.envs.items():
         if not isinstance(env.image, str):
+            # No base image but the name is set, try getting image uri from configm
+            if cfg and env.image.base_image is None and env.image.name is not None:
```
pingsutw: But if people use from_dockerfile, the base name will be None, and it will be overridden here.

pingsutw: we can just add one more check here

src/flyte/_deploy.py
```diff
@@ -217,16 +217,34 @@ async def _build_images(deployment: DeploymentPlan) -> ImageCache:
...
     """
+    from ._initialize import _get_init_config
     from ._internal.imagebuild.image_builder import ImageCache
 
     images = []
     image_identifier_map = {}
+    cfg = _get_init_config()
     for env_name, env in deployment.envs.items():
         if not isinstance(env.image, str):
+            # No base image but the name is set, try getting image uri from configm
```
pingsutw: nit
```suggestion
            # No base image but the name is set, try getting image uri from config.
```

src/flyte/_deploy.py
machichima: Fixed in [bcc8439](https://github.com/flyteorg/flyte-sdk/pull/190/commits/bcc8439946ef681c7f5f8552ddcaba088271dacb)

src/flyte/_deploy.py
machichima: Fixed!

pingsutw (approved):

machichima:
Testing locally (running `examples/image/image_from_ref_name.py main`):

- Setting image through cli

 ```sh
❯ flyte run --image custom-image=debian:stable examples/image/image_from_ref_name.py main
15:10:53.268352 WARNING  remote_builder.py:87 -  Image 356633062068.dkr.ecr.us-east-2.amazonaws.com/union/demo:custom-image-595430f39d9f776428756660507695d0 found. Skip building.
15:10:53.272405 WARNING  _deploy.py:255 -  Built Image for environment reuse_concurrency, image:
                         356633062068.dkr.ecr.us-east-2.amazonaws.com/union/demo:custom-image-595430f39d9f776428756660507695d0
╭────────────────────────────────────────────────── Run ───────────────────────────────────────────────────╮
│ Created Run: rhjf74r2fg9rt5fqfmv7  (Project: flytesnacks, Domain: development)                           │
│ ➡️  https://demo.hosted.unionai.cloud/v2/runs/project/flytesnacks/domain/development/rhjf74r2fg9rt5fqfmv7 │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────╯

```

image

- Setting image through cli, but remove additional `.with_... ` layers. This will not trigger image builder

```sh
❯ flyte run --image custom-image=debian:stable examples/image/image_from_ref_name.py main
╭────────────────────────────────────────────────── Run ───────────────────────────────────────────────────╮
│ Created Run: rwnhc87wnxrgwf27th4r  (Project: flytesnacks, Domain: development)                           │
│ ➡️  https://demo.hosted.unionai.cloud/v2/runs/project/flytesnacks/domain/development/rwnhc87wnxrgwf27th4r │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────╯

```

- Setting image in config

```yaml
# config.yaml

image:
  image_refs:
    custom-image: debian:stable

```

```sh
❯ flyte run examples/image/image_from_ref_name.py main
15:15:05.374767 WARNING  remote_builder.py:87 -  Image 356633062068.dkr.ecr.us-east-2.amazonaws.com/union/demo:custom-image-595430f39d9f776428756660507695d0 found. Skip building.
15:15:05.377087 WARNING  _deploy.py:255 -  Built Image for environment reuse_concurrency, image:
                         356633062068.dkr.ecr.us-east-2.amazonaws.com/union/demo:custom-image-595430f39d9f776428756660507695d0
╭────────────────────────────────────────────────── Run ───────────────────────────────────────────────────╮
│ Created Run: rsvkcs5r47jq4xbkxm75  (Project: flytesnacks, Domain: development)                           │
│ ➡️  https://demo.hosted.unionai.cloud/v2/runs/project/flytesnacks/domain/development/rsvkcs5r47jq4xbkxm75 │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

src/flyte/_deploy.py
```diff
@@ -423,5 +445,7 @@ async def build_images(envs: Environment) -> ImageCache:
     :param envs: Environment to build images for.
     :return: ImageCache containing the built images.
     """
+    cfg = get_init_config()
+    images = cfg.images if cfg else {}
     deployment = plan_deploy(envs)
-    return await _build_images(deployment[0])
+    return await _build_images(deployment[0], images)
```
machichima, line 451: Here is being called when building: 

https://github.com/flyteorg/flyte-sdk/blob/4b8fa2df2a3100fcf081b55edee888778f856d53/src/flyte/cli/_build.py#L39-L54

src/flyte/_image.py
```diff
@@ -399,6 +399,8 @@ class Image:
     name: Optional[str] = field(default=None)
     platform: Tuple[Architecture, ...] = field(default=("linux/amd64",))
     python_version: Tuple[int, int] = field(default_factory=_detect_python_version)
+    # Refer to the image_refs (name:image-uri) set in CLI or config
+    _ref_name: Optional[str] = field(default=None)
```
machichima, line 403: We cannot set `init=False` for this as we need to pass this in when doing `Image._new` in `clone`

https://github.com/flyteorg/flyte-sdk/pull/190/files#diff-d8bacb3ff6a674ba80976b1a5c46430412b8ef5b69ac3294ba13172ccb819b7eR664

PR #191, opened 2025-09-17
Artifacts

kumare3:
End-to-end artifacts on the v2 stack: publish from local, consume in `flyte run` with provenance, and produce from task outputs. All flows devbox-e2e-verified. Companion PRs: protos flyteorg/flyte#7733/#7745/#7748 (merged, released as flyteidl2 **2.0.35**), backend stack unionai/cloud#17372 → #17373 → #17334 → #17374.

## What can be an artifact
Artifacts are **offloaded assets only**: a `flyte.io` `File`, `Dir`, or `DataFrame`. Primitives, `bytes`, dataclasses, pydantic models, and arbitrary objects are rejected (`ensure_artifactable` is an allow-list). An artifact must be a top-level task output; nesting a wrapped value inside a model/container fails at conversion time.

## Consume / publish (flyte.remote.Artifact)
- `Artifact.create/get/listall/list_names` implemented against the new `flyteidl2.artifact.ArtifactService` (ConnectRPC client wired through `ClientSet.artifact_service`). `create()` publishes from the local machine: type-engine literal conversion (local files/dirs/dataframes offload via dataproxy first), metadata seeded from `flyte.artifacts.new(...)` wrappers. `created_by` renders the `EnrichedIdentity` attribution.
- CLI: `flyte get artifact` (distinct names with version counts via `ListArtifactNames`, `--search` substring filter), `flyte get artifact NAME` (all versions), `flyte get artifact NAME VERSION` (full details); `flyte create artifact NAME --from-file` (File publish).
- `flyte.run` consumption in **local and remote** modes: async `_unwrap_artifacts` materializes each artifact literal to its python value before input conversion.

## Metadata and cards
- `flyte.artifacts.Metadata` (+ `Metadata.create_model_metadata` for model artifacts: framework/architecture/task/modality/... as searchable key-values; the description travels in the dedicated `description` field, not the key-value map).
- `flyte.artifacts.Card` (`create_from(content, format, card_type)`): rendered HTML/markdown cards upload to run metadata storage and attach to the artifact (`spec.card`); the UI presigns and renders them inline.

## Provenance on run inputs and outputs
- Consumed artifacts are stamped into the bound input literal's metadata (`_ua`, v1-parity key) as `org/project/domain/name@version` — `flyte get io` renders `value (artifact: <tracker>)`. Client-side at submission, so no pod-image dependency.
- Produced outputs render as `value (produced artifact: name[@version])` in `flyte get io` / `ActionOutputs`.
- `Artifact.create()` inside a running task records the producing task action (`spec.source.task_action`) automatically; `external_ref=` overrides for imported assets.

## Producer side (`produces_artifacts`)
- `@env.task(produces_artifacts=True)` threads through `TaskTemplate`/`override()` into `TaskMetadata.produces_artifacts` for the backend (leaseworker extracts artifacts from stamped output literals — see the cloud PR).
- Output conversion now **preserves** `artifacts.new()` wrapper metadata (previously discarded): stamped as compact deterministic JSON under `_uap` on the output literal (unconditional; the flag only gates backend extraction). JSON fixtures pinned byte-identical with the Go reader in the cloud PR.

## Also in this PR
- **flyteidl2 pin bumped to the released 2.0.35** (artifact protos incl. `ListArtifactNames`, `produces_artifacts`; exact pins across root + all plugins, no local overrides).
- RunSpec/TaskAction `cluster`→`queue` rename (shipped in flyteidl2 ≥2.0.31) across the three SDK call sites + tests.
- Image builder: the final `--no-index --reinstall` step now forces **every** wheel in `dist/` (not just the `flyte` package), so sibling local wheels win over PyPI releases during development.


wild-endeavor:
should we add modelcard and user specified metadata? 

```python
artifacts.Metadata(
    name="my_artifact", version="1.0", description="An example artifact created in create_artifact task",
    card=ModelCard("single.html"),
    custom={"a": "one", "b": {"c": 3}}
)
```

kumare3:
> should we add modelcard and user specified metadata?
> 
> ```python
> artifacts.Metadata(
>     name="my_artifact", version="1.0", description="An example artifact created in create_artifact task",
>     card=ModelCard("single.html"),
>     custom={"a": "one", "b": {"c": 3}}
> )
> ```

Yes I want to, i was thinking we make ModelMetadata and DataMetadata?
Question is, how do we upload the modelcard, do we upload are part of artifact creation or should we do it at transform time, i think before right?

src/flyte/types/__init__.py
```diff
@@ -35,6 +35,8 @@
     "TypeEngine",
     "TypeTransformer",
     "TypeTransformerFailedError",
+    "artifact_annotation",
```
kumare3: why is this exported?

src/flyte/types/__init__.py
```diff
@@ -35,6 +35,8 @@
...
+    "produced_artifact_annotation",
```
kumare3: same

src/flyte/types/__init__.py
kumare3: Removed from the public surface in 8549a2c55f — internal display plumbing only; remote/_action.py imports from flyte.types._string_literals directly.

src/flyte/types/__init__.py
kumare3: Same — removed in 8549a2c55f.

kumare3:
Consolidated into #1362 (retargeted to main), which contains this branch's full history.

PR #192, opened 2025-09-17
Add an example for submitting rayjob to an existing cluster

pingsutw:


examples/plugins/ray_existing_cluster.py
```diff
@@ -0,0 +1,75 @@
...
+    return res
+
+
+@ray_env.task
+async def create_ray_cluster() -> str:
+    """
+    Create a Ray cluster and return the head node IP address.
+    """
+    print("creating ray cluster")
+    cluster_ip = os.getenv("MY_POD_IP")
```
pingsutw, line 59: kuberay will add this env to the head node

PR #202, opened 2025-09-18
Allow local builder's command to use mounted secrets, fix secret command generation from env vars

redartera:
## Description

This PR includes a bug fix in how secret env vars are added to the docker build command - the format should be `--secret id=<SOME_ID>,env=<ENV_VAR_NAME>` however currently `env=<ENV_VAR_VALUE>`.

Additionally, we add support for the `CommandsHandler` to pass secret mounts given the commands layer can accept those as arguments - it is also useful to run commands with some secret mounted.


pingsutw (approved):

pingsutw:
LGTM, thanks!

PR #203, opened 2025-09-18
Reduce startup overhead for the debug button

pingsutw:
## Summary
- Remove Jupyter extension from default code server extensions to reduce startup time
- Prevent overriding existing LOG_LEVEL environment variables in runner


src/flyte/_debug/constants.py
```diff
@@ -12,7 +12,6 @@
 }
 DEFAULT_CODE_SERVER_EXTENSIONS = [
     "https://raw.githubusercontent.com/flyteorg/flytetools/master/flytekitplugins/flyin/ms-python.python-2023.20.0.vsix",
-    "https://raw.githubusercontent.com/flyteorg/flytetools/master/flytekitplugins/flyin/ms-toolsai.jupyter-2023.9.100.vsix",
```
pingsutw, line 15: Downloading and installing this extension takes 10+ seconds, and we usually don’t use it, so I removed it here

PR #206, opened 2025-09-20
Fix pickle transformer

kumare3:
This pull request introduces support for handling Python pickled objects as both blobs and binary types, and adds comprehensive tests for these changes. The main updates focus on improving the type guessing and literal type construction for pickled data, ensuring better interoperability and robustness.

Enhancements to Pickle Type Handling:

* Updated `FlytePickleTransformer.guess_python_type` to return `FlytePickle` for both blob types with pickle format and for simple binary types, improving flexibility in type inference.
* Modified `FlytePickleTransformer.get_literal_type` to return a union type containing both blob and binary variants, allowing for broader compatibility when serializing Python objects.

Testing Improvements:

* Added tests to verify that `guess_python_type` correctly handles binary simple types, that `get_literal_type` returns a union type with the expected variants, and that unsupported types raise appropriate errors.

Example Usage:

* Added a new example in `examples/basics/arbitrary_data.py` demonstrating how to define and pass arbitrary Python objects between Flyte tasks, showcasing the improved pickling support.

### DIFF
--- a/src/flyte/types/_pickle.py
+++ b/src/flyte/types/_pickle.py
@@ -129,9 +129,12 @@ class FlytePickleTransformer(TypeTransformer[FlytePickle]):
                         blob=types_pb2.BlobType(
                             format=self.PYTHON_PICKLE_FORMAT,
                             dimensionality=types_pb2.BlobType.BlobDimensionality.SINGLE,
-                        )
+                        ),
+                        structure=types_pb2.TypeStructure(tag=self.name),
                     ),
-                    types_pb2.LiteralType(simple=types_pb2.SimpleType.BINARY),
+                    types_pb2.LiteralType(
+                        simple=types_pb2.SimpleType.BINARY, structure=types_pb2.TypeStructure(tag=self.name)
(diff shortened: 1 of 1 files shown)


pingsutw (dismissed):

src/flyte/types/_pickle.py
```diff
@@ -116,14 +116,24 @@ def guess_python_type(self, literal_type: types_pb2.LiteralType) -> typing.Type[
...
-            )
+            union_type=types_pb2.UnionType(
+                variants=[
+                    types_pb2.LiteralType(
+                        blob=types_pb2.BlobType(
+                            format=self.PYTHON_PICKLE_FORMAT,
+                            dimensionality=types_pb2.BlobType.BlobDimensionality.SINGLE,
+                        )
+                    ),
+                    types_pb2.LiteralType(simple=types_pb2.SimpleType.BINARY),
```
kumare3: For binary should we add a pickle tag?

pingsutw (approved):

PR #207, opened 2025-09-21
Fixed local execution for paths/reports/raw-data etc

kumare3:


src/flyte/_run.py
```diff
@@ -437,8 +448,8 @@ async def _run_local(self, obj: TaskTemplate[P, R], *args: P.args, **kwargs: P.k
                 checkpoint_path=internal_ctx().raw_data.path,
             ),
             code_bundle=None,
-            output_path=self._metadata_path,
-            run_base_dir=self._metadata_path,
+            output_path=str(output_path),
+            run_base_dir=str(metadata_path),
             version="na",
             raw_data_path=internal_ctx().raw_data,
```
machichima:
```suggestion
            raw_data_path= raw_data_path,
```
I think we should use `raw_data_path` here?

examples/basics/report_example.py
```diff
@@ -25,6 +25,8 @@ async def main():
 
 
 if __name__ == "__main__":
-    flyte.init_from_config()
-    run = flyte.run(main)
+    import flyte.git
+
+    flyte.init_from_config(flyte.git.config_from_root())
+    run = flyte.with_runcontext("local").run(main)
```
machichima: Should we create a new `report_example_local.py` instead of modifying the original one?

kumare3: the goal is that you should be able to run anything locally and remote. I think we might want to run all our examples locally, maybe support an env var.

src/flyte/_run.py
kumare3: fixed, thank you and good catch

examples/basics/report_example.py
```diff
@@ -25,6 +25,8 @@ async def main():
 
 
 if __name__ == "__main__":
-    flyte.init_from_config()
+    import flyte.git
+
+    flyte.init_from_config(flyte.git.config_from_root())
```
wild-endeavor, line 30: i thought this wasn't needed anymore.

kumare3: wdym, this is needed?

PR #208, opened 2025-09-22
Add uv workspace example in deploy_patterns

popojk:
## Description
We need to document the common issues encountered when running tasks in a uv workspace.

## Changes
Add an example for uv workspace.

### DIFF
--- a/examples/deploy_patterns/uv_workspace/README.md
+++ b/examples/deploy_patterns/uv_workspace/README.md
@@ -24,13 +24,13 @@ uv_workspace/
 ```
 
 ##  Testing This Example
-Execute the shell script:
-```bash
-./run.sh
-```
-Or execute the flyte command directly:
+Execute the flyte command directly:
 ```bash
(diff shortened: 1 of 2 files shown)


examples/deploy_patterns/uv_workspace/README.md
```diff
@@ -0,0 +1,79 @@
+
+# Flyte SDK: UV Workspace Deployment Pattern
```
wild-endeavor, line 2: rename file?  spelling

examples/deploy_patterns/uv_workspace/run.sh
```diff
@@ -0,0 +1,3 @@
+CURR_PATH=$(pwd)
+echo $CURR_PATH
+PYTHONPATH=${CURR_PATH} python src/workflows/albatross.py
```
pingsutw: Does it work if we just run `python src/workflows/albatross.py`? If so, could we remove this file?

PR #209, opened 2025-09-22
[WIP] Connector Interface

pingsutw:


plugins/connectors/src/flyteplugins/connectors/bigquery/connector.py
```diff
@@ -0,0 +1,118 @@
...
+        location = custom["Location"]
+
+        if custom.get("google_application_credentials") is not None:
+            google_application_credentials = service_account.Credentials.from_service_account_info(
+                custom["google_application_credentials"]
+            )
+        else:
+            google_application_credentials = None
+
+        client = bigquery.Client(
```
kumare3: i think we should use alru_cache to cache creation of client, where the cache key is driven by the "secret key and secret value"

plugins/connectors/src/flyteplugins/connectors/bigquery/task.py
```diff
@@ -68,6 +72,8 @@ def custom_config(self, sctx: SerializationContext) -> Optional[Dict[str, Any]]:
         }
         if self.plugin_config.QueryJobConfig is not None:
             config.update(self.plugin_config.QueryJobConfig.to_api_repr()["query"])
+        if self.google_application_credentials is not None:
+            config["google_application_credentials"] = f"{{ .secrets.{self.google_application_credentials} }}"
```
kumare3: we dont need templating, we could simply add the "google_app_credentials_secret_key=key" to the Metadata, and then in connector look at all secrets to find the value that matches the key

pyproject.toml
```diff
@@ -137,7 +137,7 @@ dev-dependencies = [
 ]
 
 [tool.uv.sources]
-flyteidl2 = { git = "https://github.com/flyteorg/flyte.git", subdirectory = "gen/python", rev = "v2" }
+flyteidl2 = { git = "https://github.com/flyteorg/flyte.git", subdirectory = "gen/python", rev = "16e01e918fe9d4ec73955cbd707435996ddac83f" }
```
pingsutw: Need to merge this first. https://github.com/flyteorg/flyte/pull/6659

plugins/connectors/src/flyteplugins/connectors/bigquery/connector.py
```diff
@@ -0,0 +1,140 @@
...
+@dataclass
+class BigQueryMetadata(ResourceMeta):
+    job_id: str
+    project: str
+    location: str
+    user_agent: str
+
+
+@lru_cache
+def _get_bigquery_client(
```
wild-endeavor: does this need to be async?

src/flyte/connectors/utils.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def is_terminal_phase(phase: TaskExecution.Phase) -> bool:
+    """
+    Return true if the phase is terminal.
+    """
+    return phase in [TaskExecution.SUCCEEDED, TaskExecution.ABORTED, TaskExecution.FAILED]
+
+
+def convert_to_flyte_phase(state: str) -> TaskExecution.Phase:
```
wild-endeavor, line 11: is this used for other connector task types besides bigquery?

src/flyte/connectors/_connector.py
```diff
@@ -0,0 +1,269 @@
...
+    def __init__(self, secrets: Dict[str, str]):
+        # Key is the id of the secret, value is the secret name.
+        self._secrets = secrets
+
+    @property
+    def secrets(self) -> List[Secret]:
+        return [Secret(key=k, as_env_var=v) for k, v in self._secrets.items()]
+
+
+class AsyncConnectorExecutorMixin(TaskTemplate):
```
wild-endeavor: i think we should not call this a mixin... or if we do, we should remove the TaskTemplate inheritance.  In bq task, we're already inheriting from TaskTemplate.  mixins in general should not inherit from concrete classes.  if we want to, I think we can add a type to self if you really want a mixin.  Or if we keep it here, then remove it from bq task and remove the word Mixin.

src/flyte/connectors/utils.py
pingsutw: yes

pingsutw: all the connectors will use it

src/flyte/connectors/_connector.py
pingsutw: removed it

plugins/connectors/src/flyteplugins/connectors/bigquery/connector.py
pingsutw: updated to async

wild-endeavor (approved):

PR #228, opened 2025-09-30
**Breaking** flyte.io.File sync/async api and Faster data

kumare3:
The File API is changed.
open now always returns an async file handle in async mode. open_sync returns a sync file handle.

open, fh.read, returns a memoryview, which has to be type casted to `bytes`. 

@wild-endeavor shall we modify it somehow to make it `bytes`? Memory view is faster

### DIFF
--- a/src/flyte/storage/_storage.py
+++ b/src/flyte/storage/_storage.py
@@ -7,6 +7,7 @@ from typing import AsyncGenerator, Optional
 from uuid import UUID
 
 import fsspec
+import obstore
 from fsspec.asyn import AsyncFileSystem
 from fsspec.utils import get_protocol
 from obstore.exceptions import GenericError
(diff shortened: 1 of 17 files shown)


src/flyte/storage/_storage.py
```diff
@@ -204,34 +204,48 @@ async def put(from_path: str, to_path: Optional[str] = None, recursive: bool = F
...
+        attributes = kwargs.pop("attributes", {})
+        file_handle = obstore.open_writer_async(store, file_path, attributes=attributes)
+    else:  # read mode
+        buffer_size = kwargs.pop("buffer_size", 10 * 2**20)
+        file_handle = await obstore.open_reader_async(store, file_path, buffer_size=buffer_size)
+
+    return file_handle
+
+
+async def open(path: str, mode: str = "rb", **kwargs):
```
wild-endeavor: can we add the type hint here?  is it just IO?

wild-endeavor (commented):
i'll add back the hashing writer stuff if someone complains

src/flyte/storage/_storage.py
kumare3: Filehandle type

wild-endeavor (approved):

PR #229, opened 2025-09-30
Use same client instance in controller initialization

pingsutw:
## Summary
- Refactor controller initialization to use the same client instance from init config
- Add proper thread startup synchronization with separate `_wait` method
- Ensure controller is properly initialized before running coroutines

## Changes
- `src/flyte/_bin/runtime.py`: Import and use client from init config
- `src/flyte/_internal/controllers/remote/_core.py`: Extract thread waiting logic into separate method and ensure proper initialization order

## Test plan
- [ ] Verify controller initialization works correctly
- [ ] Test that coroutines wait for controller thread to be ready
- [ ] Confirm client instance is properly shared

### DIFF
--- a/src/flyte/_bin/runtime.py
+++ b/src/flyte/_bin/runtime.py
@@ -12,6 +12,7 @@ from typing import Any, List
 
 import click
 
+from flyte._initialize import _get_init_config
 from flyte.models import PathRewrite
 
 # Todo: work with pvditt to make these the names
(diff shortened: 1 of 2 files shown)


src/flyte/_bin/runtime.py
```diff
@@ -142,8 +142,9 @@ def main(
     bundle = None
     if tgz or pkl:
         bundle = CodeBundle(tgz=tgz, pkl=pkl, destination=dest, computed_version=version)
-    init(org=org, project=project, domain=domain, image_builder="remote", **controller_kwargs)
```
pingsutw, line 145: We created a gRPC channel twice (`init`, `create_controller`). I think we should be able to reuse it in the controller.

PR #231, opened 2025-09-30
[WIP] TensorFlow plugin

kevinliao852:
# Summary

This PR introduces the initial implementation of the TensorFlow plugin for Flyte SDK.
It sets up the basic structure needed to run TensorFlow tasks within Flyte.

### DIFF
--- a/examples/plugins/tensorflow_example.py
+++ b/examples/plugins/tensorflow_example.py
@@ -1,0 +1,74 @@
+from pathlib import Path
+
+import tensorflow as tf
+from flyteplugins.tensorflow.task import Tensorflow
+from tensorflow.keras import datasets, layers, models
+
+import flyte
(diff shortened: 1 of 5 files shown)


kevinliao852:
> I do not think we should add this @kevinliao852 and @pingsutw I do not think tensorflow is in favor anymore and I am not sure if folks use it. I would much rather start porting types and connectors

Thanks for the feedback. That makes sense.

I was aiming to get a basic TensorFlow plugin in place since there wasn't one yet, but I understand if the team prefers to focus efforts on porting types and connectors first.

Happy to close this for now or just leave it open if you think it might be useful later. Let me know what works best for the project!

pingsutw:
Just keep it here. we can revive it once someone wants to use it.

PR #232, opened 2025-10-01
Deploy trigger

kumare3:


src/flyte/_internal/runtime/trigger_serde.py
```diff
@@ -45,11 +62,50 @@ def to_task_trigger(t: Trigger) -> trigger_definition_pb2.TaskTrigger:
     )
 
     kickoff_arg_name = None
+    default_inputs = {}
     if t.inputs:
         for k, v in t.inputs.items():
             if v is TriggerTime:
                 kickoff_arg_name = k
                 break
```
machichima: Should we use `continue` here instead to ensure all inputs are being parsed?

src/flyte/_internal/runtime/trigger_serde.py
```diff
@@ -45,11 +62,50 @@ def to_task_trigger(t: Trigger) -> trigger_definition_pb2.TaskTrigger:
...
+            literal_coros.append(flyte.types.TypeEngine.to_literal(v, type(v), task_inputs.variables[k].type))
+            keys.append(k)
+
+    final_literals: list[literals_pb2.Literal] = await asyncio.gather(*literal_coros)
+
+    for p in task_default_inputs or []:
+        if p.name not in keys:
+            keys.append(p.name)
+            final_literals.append(p.parameter.default)
+
```
machichima: Do we need to check if all args are set?

src/flyte/remote/_trigger.py
```diff
@@ -278,7 +308,9 @@ def _rich_automation(self, automation: common_pb2.TriggerAutomationSpec):
         if automation.type == common_pb2.TriggerAutomationSpec.TYPE_NONE:
             yield "none", None
         elif automation.type == common_pb2.TriggerAutomationSpec.TYPE_SCHEDULE:
-            if automation.WhichOneof("schedule_type") == "fixed_rate":
+            if automation.schedule.cron_expression is not None:
+                yield "cron", automation.schedule.cron_expression
```
wild-endeavor, line 293: do these work?  you don't have to do the silly hasfield thing right?

wild-endeavor (dismissed):
quick question +1

src/flyte/_internal/runtime/trigger_serde.py
wild-endeavor: oops totally missed this.  yes please  thanks @machichima!

kumare3: yes, i changed the code, oops

src/flyte/_internal/runtime/trigger_serde.py
kumare3: all args may not be set. This is because, only some might have defaults

src/flyte/remote/_trigger.py
kumare3: seems to work

machichima (approved):

machichima (approved):

PR #239, opened 2025-10-03
[wip] Accelerator type support

kumare3:


src/flyte/_resources.py
```diff
@@ -147,7 +233,31 @@ def TPU(device: TPUType, partition: V5PParts | V6EParts | None = None):
     elif partition is not None and device == "V5E":
         if partition not in get_args(V5EParts):
             raise ValueError(f"Invalid partition for V5E: {partition}. Must be one of {get_args(V5EParts)}")
-    return Device(1, device, partition)
+    return Device(1, "TPU", device, partition)
```
jeevb, line 236: Should we use kwargs here for consistency?

PR #253, opened 2025-10-09
[Breaking] Obstore parallel reader

wild-endeavor:
Just copying from @jeevb.  Thanks!

The obstore store object by default uses a single connection - this PR introduces a parallel reader (writer to come) to the storage layer, to read files in parallel.  Files under the default chunk size of 16MBs will continue to just use one connection.

Also using this opportunity to update the API around how the offloaded types (files/dir) behave.  The new behavior better mirrors how `aws s3 cp` works.

### File objects
If you have a task that takes in some `f: flyte.io.File` with a `path` set to `s3://bucket/file.txt`, the behavior for `download` now is
* `f.download()`
    without an argument for the local destination - will download to a randomly created local folder with a filename of `file.txt`
* `f.download("some/filename")`
    with a file name specified - will download to that specified file name.
* `f.download("new/path/")` or `f.download("existing/folder")`
    with an existing local directory or with a non-existent string that ends with `/` - will download to `file.txt` under that folder name, creating it if necessary.

#### file.open()
This was introduced in another PR, but will mention it again here as some examples were updated.  If you have a file object, and you do `async with f.open() as fh`, you'll now need to do `await fh.read()` instead of just `fh.read`.

Also, the output of `fh.read` (when using s3/gcs/abfs) will be an obstore Bytes object, not a python bytes object, which means if you are calling `.decode` on it directly it will fail.  If you're trying to read it as a string, use `str(b, "utf-8")` rather than calling `to_bytes` on it to save the memory copy.

### Dir objects
If you have a task that takes in some `d: flyte.io.Dir` with a `path` set to `s3://bucket/my_remote_folder`, the behavior for download is now
* `d.download()`
    The _contents_ of `my_remote_folder` will be downloaded into a new local directory.
* `d.download("new_parent/new_folder")`
    The _contents_ of `my_remote_folder` will be downloaded as the contents of `new_parent/new_folder`
* `d.download("existing_dir")`
    The _contents_ of `my_remote_folder` will be downloaded as the contents of `existing_dir`

In no case will the name of the remote folder `my_remote_folder` show up.  Previously this was not the case and the remote folder name would show up locally.  The new behavior matches what `aws s3 cp --recursive` does and also mostly matches *nix `cp -R`.

Also,
* The `remote_path` input to `Dir.from_local` has been changed to `remote_destination` to match the `File` object.

### DIFF
--- a/src/flyte/io/_dir.py
+++ b/src/flyte/io/_dir.py
@@ -606,10 +606,9 @@ class Dir(BaseModel, Generic[T], SerializableType):
         protocol = get_protocol(resolved_remote_path)
 
         # Shortcut for local, don't copy and just return
-        if "file" in protocol:
-            if remote_destination is None:
-                output_path = str(Path(local_path).absolute())
-                return cls(path=output_path, name=dirname, hash=dir_cache_key)
+        if "file" in protocol and remote_destination is None:
+            output_path = str(Path(local_path).absolute())
+            return cls(path=output_path, name=dirname, hash=dir_cache_key)
 
         # todo: in the future, mirror File and set the file to_path here
         output_path = await storage.put(from_path=local_path_str, to_path=remote_destination, recursive=True)
@@ -678,10 +677,9 @@ class Dir(BaseModel, Generic[T], SerializableType):
         protocol = get_protocol(resolved_remote_path)
 
         # Shortcut for local, don't copy and just return
-        if "file" in protocol:
-            if remote_destination is None:
-                output_path = str(Path(local_path).absolute())
-                return cls(path=output_path, name=dirname, hash=dir_cache_key)
+        if "file" in protocol and remote_destination is None:
+            output_path = str(Path(local_path).absolute())
+            return cls(path=output_path, name=dirname, hash=dir_cache_key)
 
         fs = storage.get_underlying_filesystem(path=resolved_remote_path)
         fs.put(local_path_str, resolved_remote_path, recursive=True)
--- a/src/flyte/io/_file.py
+++ b/src/flyte/io/_file.py
@@ -533,7 +533,7 @@ class File(BaseModel, Generic[T], SerializableType):
             local_path_str = str(local_path)
             has_trailing_sep = local_path_str.endswith(os.sep)
             local_path = str(Path(local_path).absolute())
-            if has_trailing_sep and not local_path.endswith(os.sep):
+            if has_trailing_sep:
                 local_path = local_path + os.sep
 
         fs = storage.get_underlying_filesystem(path=self.path)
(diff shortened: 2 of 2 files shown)


examples/basics/offloaded_type_basic_usage.py
```diff
@@ -0,0 +1,118 @@
...
+
+env = flyte.TaskEnvironment(
+    "file_dir_downloading_semantics",
+)
+
+
+@env.task
+async def create_a_file() -> flyte.io.File:
+    chunk_size = 1024  # 1 KiB
+    with open("example.txt", "wb") as f:  # noqa: ASYNC230
```
pingsutw, line 17: nit: Should we use async open in the example?

src/flyte/io/_dir.py
```diff
@@ -643,14 +674,18 @@ def upload_with_cache_key_sync() -> Dir:
...
         dirname = os.path.basename(os.path.normpath(local_path_str))
 
-        if not remote_path:
-            from flyte._context import internal_ctx
+        resolved_remote_path = remote_destination or internal_ctx().raw_data.get_random_remote_path(dirname)
+        protocol = get_protocol(resolved_remote_path)
+
+        # Shortcut for local, don't copy and just return
+        if "file" in protocol:
+            if remote_destination is None:
```
pingsutw: nit
```suggestion
        if "file" in protocol and remote_destination is None:
```

src/flyte/io/_dir.py
```diff
@@ -581,15 +602,24 @@ async def upload_with_cache_key() -> Dir:
         """
         local_path_str = str(local_path)
         dirname = os.path.basename(os.path.normpath(local_path_str))
+        resolved_remote_path = remote_destination or internal_ctx().raw_data.get_random_remote_path(dirname)
+        protocol = get_protocol(resolved_remote_path)
+
+        # Shortcut for local, don't copy and just return
+        if "file" in protocol:
+            if remote_destination is None:
```
pingsutw:
```suggestion
        if "file" in protocol and remote_destination is None:
```

src/flyte/io/_file.py
```diff
@@ -527,23 +527,40 @@ async def download_to_path(f: File) -> str:
...
         """
         if local_path is None:
-            local_path = storage.get_random_local_path(file_path_or_file_name=local_path)
+            local_path = storage.get_random_local_path(file_path_or_file_name=self.path)
         else:
+            # Preserve trailing separator if present (Path.absolute() strips it)
+            local_path_str = str(local_path)
+            has_trailing_sep = local_path_str.endswith(os.sep)
             local_path = str(Path(local_path).absolute())
+            if has_trailing_sep and not local_path.endswith(os.sep):
```
pingsutw: > local_path.endswith(os.sep)

Isn't it always return false?

pingsutw: since local path is from Path(local_path).absolute()

src/flyte/io/_file.py
```diff
@@ -579,19 +596,36 @@ def download_to_path_sync(f: File) -> str:
...
```
pingsutw: ditto

examples/basics/offloaded_type_basic_usage.py
wild-endeavor: hmm?  what is that?  aiofiles?  meh

src/flyte/io/_file.py
wild-endeavor: oops, claude, removing.

pingsutw (approved):

PR #254, opened 2025-10-09
add `context`

samhita-alla:


src/flyte/_internal/controllers/remote/_controller.py
```diff
@@ -17,6 +17,7 @@
 import flyte.storage as storage
 from flyte._code_bundle import build_pkl_bundle
 from flyte._context import internal_ctx
+from flyte._input_context import _input_context_var
```
kumare3: shall we call this custom_context?

src/flyte/_input_context.py
```diff
@@ -0,0 +1,120 @@
...
+    task_ctx = ctx()
+    if task_ctx and task_ctx.input_context:
+        return task_ctx.input_context.copy()
+
+    # Otherwise, check the context variable (for context manager usage)
+    return _input_context_var.get().copy()
+
+
+@asynccontextmanager
+async def input_context(**context: str) -> AsyncIterator[None]:
```
kumare3: I like it, but this has a problem. If we want a context key like `wandb-ctx` `-` will not be supported. That might be ok, i just wanted to flag

src/flyte/_input_context.py
```diff
@@ -0,0 +1,120 @@
...
+
+import contextvars
+from contextlib import asynccontextmanager, contextmanager
+from typing import AsyncIterator, Dict, Iterator
+
+from flyte._context import ctx
+
+# Context variable to store the input context
+# This stores both global context (from with_runcontext) and local context (from context managers)
+_input_context_var: contextvars.ContextVar[Dict[str, str]] = contextvars.ContextVar("input_context", default={})
```
kumare3: Why do we need this? 
Input context should only be on the task right? Global context is dangerous and will not work for actors. 
Let's just use taskcontext.

To pass it down, you can modify the taskcontext data. This is what i was going to write

src/flyte/_input_context.py
kumare3: We do not need async. Lets just have sync context manager thats it. check out group. We should be doing no IO and you can access task context in sync

src/flyte/_run.py
```diff
@@ -462,15 +466,26 @@ async def _run_local(self, obj: TaskTemplate[P, R], *args: P.args, **kwargs: P.k
...
-                fut = controller.submit_sync(obj, *args, **kwargs)
-                awaitable = asyncio.wrap_future(fut)
-                outputs = await awaitable
-            else:
-                outputs = await controller.submit(obj, *args, **kwargs)
+
+        # Set input context from with_runcontext if provided
+        context_token = None
+        if self._input_context:
+            context_token = _input_context_var.set(self._input_context.copy())
```
kumare3: lets just set it on task context., we do not need to change any code, besides adding task context

src/flyte/_input_context.py
samhita-alla: 1. as for `wandb_context`, i think we can extend `custom_context` to support it, right?
```
@contextmanager
def wandb_context(
      project: str | None = None,
      entity: str | None = None,
      run_id: str | None = None,
      tags: list[str] | None = None,
  ):
      wandb_config = {}
      if project is not None:
          wandb_config["wandb_project"] = project
     ...

      with custom_context(**wandb_config):
          yield
```

2. removed async and just kept the sync context manager. and yes, no IO anymore and global context var.

kumare3 (approved):

pingsutw (approved):

PR #256, opened 2025-10-09
Add require_project_and_domain decorator

pingsutw:
Add a `require_project_and_domain` decorator, and apply it only to functions that need it

### DIFF
--- a/plugins/spark/src/flyteplugins/spark/task.py
+++ b/plugins/spark/src/flyteplugins/spark/task.py
@@ -8,7 +8,7 @@ import flyte
 from flyte import PodTemplate
 from flyte.extend import AsyncFunctionTaskTemplate, TaskPluginRegistry
 from flyte.models import SerializationContext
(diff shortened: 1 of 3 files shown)


examples/plugins/dask_example.py
```diff
@@ -11,8 +11,8 @@
 image = flyte.Image.from_debian_base(python_version=(3, 12)).with_pip_packages("flyteplugins-dask")
 
 dask_config = Dask(
-    scheduler=Scheduler(resources=Resources(cpu="1", memory="1Gi")),
```
pingsutw, line 14: Temporarily remove it since the resource will be serialized to IDL2 proto, but the Dask plugin proto is still from IDL1.
https://github.com/flyteorg/flyte-sdk/blob/784e45fe6684319ad77a0140884afdd361b5a76d/plugins/dask/src/flyteplugins/dask/task.py#L96

PR #260, opened 2025-10-10
introduce timezone for CRON

fiedlerNr9:
- Introduce timezones for cron triggers

``` python
custom_trigger = flyte.Trigger(
    "custom_cron",
    flyte.Cron("0 0 * * *", timezone="Europe/London"),  # Runs once every day at midnight London time
    inputs={"start_time": flyte.TriggerTime, "x": 1},
)
```

### DIFF
--- a/src/flyte/_trigger.py
+++ b/src/flyte/_trigger.py
@@ -696,8 +696,9 @@ class Trigger:
        values for inputs to these defaults.
     :param env_vars: (Dict[str, str]) Optional environment variables for the trigger, default is None. If provided, will
         replace the environment variables set in the config of the task.
-    :param interruptible: (bool) Whether the trigger is interruptible, default is None. If provided,
-     it overrides whatever is set in the config of the task.
+    :param interruptible: (bool) Whether the trigger run is interruptible,
+      default is None (maintains the configured behavior). If provided, it overrides whatever is set in the config
+      of the task.
     :param overwrite_cache: (bool) Whether to overwrite the cache, default is False.
     :param queue: (str) Optional queue to run the trigger in, default is None.
     :param labels: (Mapping[str, str]) Optional labels to attach to the trigger, default is None.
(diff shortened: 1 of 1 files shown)


src/flyte/_trigger.py
```diff
@@ -33,9 +35,14 @@ class Cron:
     """
 
     expression: str
+    timezone: Literal["America/Los_Angeles", "Europe/London", "Australia/Sydney", "America/New_York"] = DEFAULT_TIMEZONE
+
+    @property
+    def timezone_expression(self) -> str:
+        return f"{self.timezone} {self.expression}"
```
iaroslav-ciupin: it must be 
```suggestion
        return f"CRON_TZ={self.timezone} {self.expression}"
```
for example:
`CRON_TZ=America/Chicago * * * * *`

src/flyte/_trigger.py
```diff
@@ -33,9 +35,14 @@ class Cron:
     """
 
     expression: str
+    timezone: Literal["America/Los_Angeles", "Europe/London", "Australia/Sydney", "America/New_York"] = DEFAULT_TIMEZONE
```
kumare3: Why these 4?

src/flyte/_trigger.py
```diff
@@ -2,10 +2,12 @@
 
 from dataclasses import dataclass
 from datetime import datetime
-from typing import Any, Dict, Mapping, Union
+from typing import Any, Dict, Literal, Mapping, Union
 
 import rich.repr
 
+DEFAULT_TIMEZONE = "America/Los_Angeles"
```
kumare3: Should this be utc - no time zone to preserve current behavior

src/flyte/_trigger.py
fiedlerNr9: This would be extended to all the timezones

src/flyte/_trigger.py
fiedlerNr9: Yeah easy to default to UTC

src/flyte/_trigger.py
```diff
@@ -2,10 +2,611 @@
 
 from dataclasses import dataclass
 from datetime import datetime
-from typing import Any, Dict, Mapping, Union
+from typing import Any, Dict, Literal, Mapping, Union
 
 import rich.repr
 
+Timezone = Literal[
```
iaroslav-ciupin, line 9: maybe naive questions: 
* why do we need to list all timezones here? 
* is this list fixed or can it change in future?
* if a user provides an invalid timezone(e.g. `cron = Cron("0 * * * *", timezone="blabla")`), will it fail at runtime?

kumare3: list of timezones is to get linter errors.

kumare3 (approved):

pingsutw (approved):

PR #261, opened 2025-10-10
Pin idl2

wild-endeavor:
This is to help with upcoming changes.


pyproject.toml
```diff
@@ -26,7 +26,7 @@ dependencies = [
     "mashumaro",
     "dataclasses_json", # TODO: remove this dependency once we remove dataclasses_json from the code
     "aiolimiter>=1.2.1",
-    "flyteidl2>=2.0.0a8",
+    "flyteidl2==2.0.0a8",
```
pingsutw, line 29: but we will unpin it in the next next release, right

wild-endeavor: no i don't think so.  at least not all the way.

PR #265, opened 2025-10-13
Fix mypy tasks

cosmicBboy:
This PR adds the `F` function type to the `TaskTemplate` generic type def, and `@overload`s to the `@env.task` method, task template `__call__` and `aio` methods to make the mypy linter happy. Without it, it tries to type check operations between `Coroutine[Any, Any, R]` with types in user code, resulting in mypy errors.

Example:

```python
"""Basic task type checking test."""

from flyte import TaskEnvironment

env = TaskEnvironment(name="test_env")


@env.task
async def simple_task(x: int) -> int:
    """Simple async task with proper types."""
    return x * 2


@env.task
def sync_task(x: str) -> str:
    """Simple sync task with proper types."""
    return x.upper()


@env.task
async def multi_input_task(x: int, y: str, z: float = 1.0) -> tuple[int, str, float]:
    """Task with multiple inputs and tuple return."""
    a = await simple_task(x)
    b = await sync_task.aio(y)
    o1 = a + 1
    o2 = b + "foo"
    o3 = z + 1.0
    return o1, o2, o3
```

mypy error:
```
tests/mypy/modules/basic_task.py:24: error: Item "function" of "AsyncFunctionTaskTemplate[[str], str] | Callable[[str], str]" has no attribute "aio"  [union-attr]
tests/mypy/modules/basic_task.py:25: error: Unsupported operand types for + ("Coroutine[Any, Any, int]" and "int")  [operator]
tests/mypy/modules/basic_task.py:25: note: Maybe you forgot to use "await"?
tests/mypy/modules/basic_task.py:25: note: Left operand is of type "Coroutine[Any, Any, int] | int"
tests/mypy/modules/basic_task.py:26: error: Unsupported left operand type for + ("Coroutine[Any, Any, str]")  [operator]
tests/mypy/modules/basic_task.py:26: note: Maybe you forgot to use "await"?
tests/mypy/modules/basic_task.py:26: note: Left operand is of type "Coroutine[Any, Any, str] | str | Any"
Found 3 errors in 1 file (checked 1 source file)
```

### DIFF
--- a/src/flyte/_map.py
+++ b/src/flyte/_map.py
@@ -1,13 +1,13 @@
 import asyncio
 import functools
 import logging
-from typing import overload, Any, AsyncGenerator, AsyncIterator, Generic, Iterable, Iterator, List, Union, cast
+from typing import Any, AsyncGenerator, AsyncIterator, Generic, Iterable, Iterator, List, Union, cast, overload
 
 from flyte.syncify import syncify
 
 from ._group import group
 from ._logging import logger
-from ._task import F, P, R, AsyncFunctionTaskTemplate
+from ._task import AsyncFunctionTaskTemplate, F, P, R
 
 
 class MapAsyncIterator(Generic[P, R]):
@@ -292,6 +292,7 @@ class _Mapper(Generic[P, R]):
                 return_exceptions=return_exceptions,
             ):
                 yield cast(Union[R, Exception], x)
+
 
 @syncify
 async def _map(


src/flyte/_task.py
```diff
@@ -49,10 +50,11 @@
 AsyncFunctionType: TypeAlias = Callable[P, Coroutine[Any, Any, R]]
 SyncFunctionType: TypeAlias = Callable[P, R]
 FunctionTypes: TypeAlias = Union[AsyncFunctionType, SyncFunctionType]
+F = TypeVar("F", bound=FunctionTypes)
```
kumare3, line 54: 💖

src/flyte/_task.py
```diff
@@ -226,6 +228,12 @@ def sql(self, sctx: SerializationContext) -> Optional[str]:
     def native_interface(self) -> NativeInterface:
         return self.interface
 
+    @overload
+    async def aio(self: TaskTemplate[P, R, SyncFunctionType], *args: P.args, **kwargs: P.kwargs) -> R: ...
```
kumare3, line 232: ❤️ note blank line

tests/mypy/test_lint_tasks.py
```diff
@@ -0,0 +1,30 @@
+"""Test running mypy on task modules."""
+
+import pathlib
```
kumare3: We don't need to do this we could just turn on mypy for examples for which today is turned off

kumare3 (dismissed):

PR #276, opened 2025-10-17
Fixes logging of error path

kumare3:


samhita-alla (approved):

src/flyte/_internal/runtime/io.py
```diff
@@ -88,7 +88,7 @@ async def upload_error(err: execution_pb2.ExecutionError, output_prefix: str):
         )
     )
     error_uri = error_path(output_prefix)
-    await storage.put_stream(data_iterable=error_document.SerializeToString(), to_path=error_uri)
+    return await storage.put_stream(data_iterable=error_document.SerializeToString(), to_path=error_uri)
```
machichima, line 91: Should we also add the return type to the function?

kumare3: Great catch, added

pingsutw (approved):

PR #279, opened 2025-10-18
Events API

kumare3:


examples/advanced/events.py
```diff
@@ -0,0 +1,54 @@
...
+
+@env.task
+async def my_task(x: int) -> int:
+    event1 = await flyte.new_event.aio(
+        "my_event",
+        scope="run",
+        prompt="Is it ok to continue?",
+        data_type=bool,
+    )
+    event2 = await flyte.new_event.aio(
```
kumare3: We should allow users to make the prompt fancy, like html etc (reports)
And then we should also have an even type that is a webhook invocation with a callback to signal

SVilgelm (dismissed):

SVilgelm (approved):

PR #289, opened 2025-10-23
spark transformer

kevinliao852:
## Overview

This PR introduces a new spark-transformer module for FlyteSDK, enabling integration between PySpark DataFrames and Flyte’s StructuredDataset abstraction.


examples/plugins/spark_transformer_exmaple.py
```diff
@@ -0,0 +1,114 @@
...
+from pathlib import Path
+from typing import cast
+
+import pandas
+from flyteplugins.spark.task import Spark
+from pandas.io.formats.style import pd
+from pyspark.sql import SparkSession
+
+import flyte
+from flyte._context import internal_ctx
```
kumare3: do not use internal_ctx even in examples

plugins/spark/src/flyteplugins/spark/sd_transformers.py
```diff
@@ -0,0 +1,88 @@
...
+from pyspark.sql.dataframe import DataFrame as PSDataFrame
+from typing_extensions import Any
+
+# pd = lazy_module("pandas")
+# pyspark = lazy_module("pyspark")
+# ps_dataframe = lazy_module("pyspark.sql.dataframe")
+# PSDataFrame = DataFrame
+
+
+class SparkDataFrameRenderer:
```
kumare3: we dont need this

plugins/spark/src/flyteplugins/spark/sd_transformers.py
```diff
@@ -0,0 +1,88 @@
...
+    def __init__(self):
+        super().__init__(PSDataFrame, None, PARQUET)
+
+    async def encode(
+        self,
+        dataframe: DataFrame,
+        structured_dataset_type: types_pb2.StructuredDatasetType,
+    ) -> literals_pb2.StructuredDataset:
+        path = typing.cast(str, dataframe.uri)
+        ctx = internal_ctx()
```
kumare3: you dont need this,
this is available in flyte.ctx() please only stick to that

kevinliao852: Got it, thanks for the clarification. I’ll switch to using flyte.ctx()

plugins/spark/src/flyteplugins/spark/df_transformer.py
```diff
@@ -0,0 +1,64 @@
...
+    def __init__(self):
+        super().__init__(pyspark.sql.DataFrame, None, PARQUET)
+
+    async def decode(
+        self,
+        flyte_value: literals_pb2.StructuredDataset,
+        current_task_metadata: literals_pb2.StructuredDatasetMetadata,
+    ) -> pyspark.sql.DataFrame:
+        spark = pyspark.sql.SparkSession.builder.getOrCreate()
+        path = os.path.join(flyte_value.uri, f"{0:05}")
```
pingsutw: hmm, why do we need to do this?

kevinliao852: I initially thought the code wouldn’t automatically create a file path, so I added this line to explicitly specify 00000.
However, I realized that the package itself handles path creation and file naming automatically.
So we don’t need to manually specify 00000 — the generated file path will already look like this:

`/tmp/flyte/raw_data/55e42ca5-142a-4499-9a8d-4d72fb079a76/423c84e5a0bf305ddf6b14e88312f515/part-00000-4ee63313-f5e7-4c89-9da4-43946627ab74-c000.snappy.parquet`

So yes, we don’t need this logic. I will remove it.

pingsutw (approved):

PR #293, opened 2025-10-23
fix sync map

samhita-alla:
Since the type hints for `__call__` are already defined in the `_Mapper` class, mypy will provide proper type checking. We don't need to add overloads again.

### DIFF
--- a/src/flyte/_map.py
+++ b/src/flyte/_map.py
@@ -189,7 +189,7 @@ class _Mapper(Generic[P, R]):
         group_name: str | None = None,
         concurrency: int = 0,
         return_exceptions: bool = True,
-    ) -> Iterator[R]: ...
+    ) -> Iterator[Union[R, Exception]]: ...
 
     def __call__(
         self,


src/flyte/_map.py
```diff
@@ -309,37 +309,4 @@ async def _map(
         yield result
 
 
-@overload
-def map(
```
cosmicbboy, line 313: we should add these overloads to the `__call__` method.

Namely, if the `return_exceptions` argument isn't provided then `__call__` should return `Iterator[R]` instead of `Iterator[Union[R, Exception]]`.

This matters for many cases when the user expects an iterator of the return type with no exceptions.

samhita-alla: they are available: https://github.com/flyteorg/flyte-sdk/blob/101b273c424426fbfc2cc2ada29ad5fa59eec89d/src/flyte/_map.py#L175-L192

samhita-alla: fixed the type hint.

PR #298, opened 2025-10-24
Improve the error message when the image is not found in the image cache

pingsutw:
This pull request improves error handling and user guidance when working with environment image caches in Flyte, making it easier to diagnose and resolve missing environment issues. The main focus is on providing clearer error messages and removing unnecessary code.

Error handling and user guidance improvements:

* Replaced a warning and fallback logic with a raised `RuntimeUserError` in `_get_urun_container`, providing a detailed message and actionable steps for users when an environment is missing from the image cache.

Code cleanup:

* Removed an unused variable assignment (`env = deployment.envs[env_name]`) from the `_build_images` function in `src/flyte/_deploy.py`.

Before:
Screenshot 2025-10-24 at 12 21 08 PM

After:
Screenshot 2025-10-24 at 12 27 10 PM

### DIFF
--- a/src/flyte/_internal/runtime/task_serde.py
+++ b/src/flyte/_internal/runtime/task_serde.py
@@ -225,19 +225,21 @@ def _get_urun_container(
             "MissingEnvironment",
             f"Environment '{env_name}' not found in image cache.\n\n"
             "💡 To fix this:\n"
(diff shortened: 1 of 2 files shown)


wild-endeavor (dismissed):

davidmirror-ops:
@pingsutw nice, so to better understand: the image layers and their cache are visible/accessible between `TaskEnvironment` objects but to enable it we need to declare dependencies between those environments right?

PR #302, opened 2025-10-29
Update creation of image pull secrets

wild-endeavor:
Update image pull secret creation flow:

```
 For image pull secrets, you have several options:

  1 Interactive mode (prompts for registry, username, password):

  $ flyte create secret my_secret --type image_pull

  2 With explicit credentials:


  $ flyte create secret my_secret --type image_pull --registry ghcr.io --username myuser

  3 Lastly, you can create a secret from your existing Docker installation (i.e., you've run docker login in the past) and you just want to pull from those
    credentials. Since you may have logged in to multiple registries, you can specify which registries to include. If no registries are specified, all
    registries are added.


  $ flyte create secret my_secret --type image_pull --from-docker-config --registries ghcr.io,docker.io
```

Previously, the pulling of the secrets from docker was hidden behind the old `union create imagepullsecret` command, which would grab all your local logins and dump them to a temp file.

* Also there was an issue (not sure recently triggered by a backend change) with list secrets - it was infinite looping because the token didn't quite work.  Switching to the per cluster token solves the issue.

### DIFF
--- a/examples/image/private_base_image.py
+++ b/examples/image/private_base_image.py
@@ -4,30 +4,46 @@ from flyte import Image
 """
 This example demonstrates how to build a new image using a private base image,
 
-
 Keep in mind we only support one secret for now for a given image. If you have a private registry that you need to pull
 from, and a different private registry you want to push to, and hence need more than one registry secret, this is not
 supported.
 
 Note that you do not need a secret to push to (or pull from) the default remote builder registry.
+
+Please see `flyte create secret --help` for all the options.
 """
 
+BASE_IMG_URL = "ghcr.io/username/private:123123"
+
+"""
+In this first example, set the registry_secret on the image, so that when building the image, the remote builder
+can pull the base image from the private registry, and build from there. However since the registry that the remote
+builder pushes to is already availble for tasks, you don't need to specify a secret in the task environment itself.
+"""
 image = (
-    Image.from_base("ghcr.io/wild-endeavor/newprivate:0977dcd202aef2c1262ded713fe12abd")
-    .clone(name="from-private-image5", registry_secret="yt_ghcr_tst5")
+    Image.from_base(BASE_IMG_URL)
+    .clone(name="from-private-image5", registry_secret="my_registry_secret")
     .with_apt_packages("vim")
     .with_local_v2()
 )
 
-"""
-Use the same registry secret in the task environment to ensure the container
-can pull the private image during task execution.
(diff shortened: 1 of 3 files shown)


examples/image/private_base_image.py
```diff
@@ -3,48 +3,50 @@
...
+"""
+In this first example, set the registry_secret on the image, so that when building the image, the remote builder
+can pull the base image from the private registry, and build from there. However since the registry that the remote
+builder pushes to is already availble for tasks, you don't need to specify a secret in the task environment itself.
 """
 image = (
-    Image.from_base("pingsutw/private:d1742efed83fc4b18c7751e53e771bbe")
-    .clone(registry="docker.io/pingsutw", name="private", registry_secret="pingsutw")
+    Image.from_base(BASE_IMG_URL)
+    .clone(name="from-private-image5", registry_secret="my_registry_secret")
```
pingsutw, line 25: We also need to specify the registry here, right? To push the image to the private registry

pingsutw (approved):

src/flyte/cli/_create.py
```diff
@@ -84,9 +136,54 @@ def secret(
...
     cfg.init(project, domain)
-    if from_file:
+
+    # Handle image pull secret creation
+    if type == "image_pull":
+        if project != "" or domain != "":
+            raise click.ClickException("Project and domain must not be set when creating an image pull secret.")
+
+        if from_docker_config:
+            # Mode 3: From Docker config
```
pingsutw, line 146:
```suggestion
            # Mode 1: From Docker config
```

src/flyte/cli/_create.py
```diff
@@ -84,9 +136,54 @@ def secret(
...
+
+            if not username:
+                username = click.prompt("Username")
+            if not password:
+                password = click.prompt("Password", hide_input=True)
+
+            value = create_dockerconfigjson_from_credentials(registry, username, password)
+
+        else:
+            # Mode 1: Interactive prompts
```
pingsutw, line 170:
```suggestion
            # Mode 3: Interactive prompts
```

examples/image/private_base_image.py
wild-endeavor: yeah but for this example i just wanted to push it to the default remote builder registry.

pingsutw: got it.

PR #306, opened 2025-10-30
Type plugins

pingsutw:
Please see the included Readme.  This adds a type plugin system, including an example.

### DIFF
--- a/src/flyte/_initialize.py
+++ b/src/flyte/_initialize.py
@@ -193,7 +193,8 @@ async def init(
     from flyte.types import _load_custom_type_transformers
 
     _initialize_logger(log_level=log_level)
-    _load_custom_type_transformers()
+    if load_plugin_type_transformers:
+        _load_custom_type_transformers()
 
     global _init_config  # noqa: PLW0603
 
--- a/examples/type_transformers/README.md
+++ b/examples/type_transformers/README.md
@@ -43,7 +43,8 @@ Register your type plugin with group name `flyte.plugins.types` and Flyte will b
 my_transformer = "my_transformer.transformer:register_positive_int_transformer"
 ```
 
-If the entry_point object is a callable function, it will be called at load time with no arguments.
+If the entry_point object is a callable function, it will be called at load time with no arguments. To avoid loading and initializing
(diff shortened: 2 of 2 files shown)


src/flyte/types/__init__.py
```diff
@@ -34,3 +36,20 @@
...
+    for ep in plugins:
+        try:
+            transformer: TypeTransformer = ep.load()
+            if not issubclass(transformer, TypeTransformer):
+                logger.error(f"Plugin {ep.name} did not return a type transformer, got {type(transformer).__name__}")
+                continue
+
+            type_transformer = transformer()
+            logger.debug(f"Registering type transformer '{type_transformer.name}'")
+            TypeEngine.register(type_transformer)
```
pingsutw: I think maybe we don't need to do it here. Users should register it in their Python package.

src/flyte/_initialize.py
```diff
@@ -187,8 +187,10 @@ async def init(
     :return: None
     """
     from flyte._utils import get_cwd_editable_install, org_from_endpoint, sanitize_endpoint
+    from flyte.types import _load_custom_type_transformers
 
     _initialize_logger(log_level=log_level)
+    _load_custom_type_transformers()
```
kumare3: lets adda a flag, `load_all_type_transformers=True`

src/flyte/types/__init__.py
```diff
@@ -34,3 +38,13 @@
     "guess_interface",
     "literal_string_repr",
 ]
+
+
+def _load_custom_type_transformers():
+    plugins = entry_points(group="flyte_type_transformers")
```
wild-endeavor: for the cli plugins i used
```
entry_points(group="flyte.plugins.cli.commands")
```
and
```
entry_points(group="flyte.plugins.cli.hooks")
```

i think i'll change this to `flyte.plugins.types`?

src/flyte/_initialize.py
wild-endeavor: added `load_plugin_type_transformers` if that's okay.  `load_all...` makes it sound like if it's False, maybe we don't even load the default ones (like int/str).

PR #307, opened 2025-10-31
Change setup to sync in dask plugin

BarryWu0812:
This PR is to change setup method to async in DownloadCodeBundleWorkerPlugin(WorkerPlugin) at dask to align official setup method in WorkerPlugin, and solve the original error `asyncio.run() cannot be called from a running event loop` when running the dask example.

https://github.com/dask/distributed/blob/1ee4756b97cd1cd5c074c40b2057d31613fe1ab9/distributed/diagnostics/plugin.py#L285

### DIFF
--- a/plugins/dask/src/flyteplugins/dask/task.py
+++ b/plugins/dask/src/flyteplugins/dask/task.py
@@ -78,13 +78,12 @@ class DownloadCodeBundleWorkerPlugin(WorkerPlugin):
     def __init__(self, code_bundle: CodeBundle):
         self.code_bundle = code_bundle
 
-    def setup(self, worker):
+    async def setup(self, worker):
         """
         Runs on each worker as it is initialized.
         """
         sys.path.insert(0, ".")
-        download_code_bundle(self.code_bundle)
-
+        await download_code_bundle(self.code_bundle)
 
 @dataclass(kw_only=True)
 class DaskTask(AsyncFunctionTaskTemplate):


pingsutw (approved):

plugins/dask/src/flyteplugins/dask/task.py
```diff
@@ -83,7 +83,7 @@ def setup(self, worker):
         Runs on each worker as it is initialized.
         """
         sys.path.insert(0, ".")
-        asyncio.run(download_code_bundle(self.code_bundle))
```
kumare3, line 86: uhho this is bad

plugins/dask/src/flyteplugins/dask/task.py
```diff
@@ -83,7 +83,7 @@ def setup(self, worker):
...
+        download_code_bundle(self.code_bundle)
```
kumare3: this looks wrong, how can download_code_bundle be both sync and async

plugins/dask/src/flyteplugins/dask/task.py
kumare3, line 70: async here

plugins/dask/src/flyteplugins/dask/task.py
kumare3: is it syncify?

BarryWu0812: Hi, thank you for pointing out.
After discussing with @pingsutw and referring to  https://github.com/dask/distributed/blob/1ee4756b97cd1cd5c074c40b2057d31613fe1ab9/distributed/diagnostics/plugin.py#L285 , I will change it to async function.

pingsutw (approved):

PR #308, opened 2025-10-31
[ImageBuilder] Add project_install_mode arg to Poetry layer in Image spec

popojk:
## Issue
Referencing to UV project, we should add a `project_install_mode` which is either "dependencies_only" or "install_project". The arg will be defaulted to "dependencies_only" so that poetry will be installed with `--no-root` arg, which tells poetry only install dependencies, otherwise it will also install the project.

## Changes
Local docker builder and remote docker builder

## Tests
Use this script to run task. The poetry `project_install_mode` config in Image spec is set to "dependencies_only". We expect that poetry will install with `--no-root` mode.
```
env = flyte.TaskEnvironment(
    name="poetry_example",
    resources=flyte.Resources(memory="250Mi"),
    image=(
        flyte.Image.from_debian_base().with_poetry_project(
            pyproject_file=pathlib.Path("pyproject.toml"),
            poetry_lock=pathlib.Path("poetry.lock"),
            project_install_mode="dependencies_only",
        )
    ),
)


@env.task
def calculate(x: int) -> int:
    ...task code

@env.task
def fetch_data(url: str) -> str:
    ...task code

@env.task
def main(x_list: list[int]) -> dict:
    """Main workflow that processes a list of numbers."""
    print("hi")
    import numpy as np

    x_len = len(x_list)
    if x_len < 5:
        raise ValueError(f"x_list needs at least 5 items, found: {x_len}")

    # Run calculations in parallel using flyte.map
    results = list(flyte.map(calculate, x_list))

    # Calculate statistics using numpy
    y_mean = float(np.mean(results))
    y_std = float(np.std(results))

    # Fetch some example data
    data_info = fetch_data("https://httpbin.org/get")

    return {
        "mean": y_mean,
        "std": y_std,
        "results": results,
        "data_info": data_info,
    }


if __name__ == "__main__":
    # Initialize remote connection
    flyte.init_from_config()

    # Run the workflow remotely
    run = flyte.run(main, x_list=list(range(10)))

    # Wait for completion and stream logs
    run.wait(run)
```

Local builder execute poetry install with `--no-root` mode:
截圖 2025-10-31 下午4 42 16

Remote builder poetry install everything without project itself:
截圖 2025-10-31 下午5 15 52

### DIFF
--- a/src/flyte/_internal/imagebuild/docker_builder.py
+++ b/src/flyte/_internal/imagebuild/docker_builder.py
@@ -310,11 +310,11 @@ class PoetryProjectHandler:
         layer: PoetryProject, context_path: Path, dockerfile: str, docker_ignore_patterns: list[str] = []
     ) -> str:
         secret_mounts = _get_secret_mounts_layer(layer.secret_mounts)
+        extra_args = layer.extra_args or ""
         if layer.project_install_mode == "dependencies_only":
             # Only Copy pyproject.yaml and poetry.lock.
             pyproject_dst = copy_files_to_context(layer.pyproject, context_path)
             poetry_lock_dst = copy_files_to_context(layer.poetry_lock, context_path)
-            extra_args = layer.extra_args or ""
             if "--no-root" not in extra_args:
                 extra_args += " --no-root"
             delta = POETRY_LOCK_WITHOUT_PROJECT_INSTALL_TEMPLATE.substitute(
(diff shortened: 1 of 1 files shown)


src/flyte/_internal/imagebuild/docker_builder.py
```diff
@@ -84,7 +84,7 @@
 RUN --mount=type=cache,sharing=locked,mode=0777,target=/tmp/poetry_cache,id=poetry \
    --mount=type=bind,target=/root/.flyte/$PYPROJECT_PATH,src=$PYPROJECT_PATH,rw \
    $SECRET_MOUNT \
-   poetry install $POETRY_INSTALL_ARGS
+   poetry install $POETRY_INSTALL_ARGS -C /root/.flyte/$PYPROJECT_PATH
```
popojk, line 87: The project dir in image should be under /root/.flyte

popojk:
@pingsutw All tests are passed. PTAL thanks!

PR #313, opened 2025-11-02
[wip] Webhook through apps and some other improvements

kumare3:


examples/apps/run_webhook.py
```diff
@@ -0,0 +1,115 @@
...
+        credentials: Bearer token for authentication
+
+    Returns:
+        Dictionary containing the launched run information:
+        - url: URL to view the run in the Flyte UI
+        - id: Unique identifier for the launched run
+    """
+    tk = await remote.TaskDetails.fetch(project=project, domain=domain, name=name, version=version)
+    r = await flyte.run.aio(tk, **inputs)
+    return {"url": r.url, "id": r.id}
```
EngHabu, line 97: Does this wait for the task to finish? If not, can we add a `/wait` as an option?

examples/apps/run_webhook.py
```diff
@@ -0,0 +1,115 @@
...
+        inputs: Dictionary of input parameters for the task
+        credentials: Bearer token for authentication
+
+    Returns:
+        Dictionary containing the launched run information:
+        - url: URL to view the run in the Flyte UI
+        - id: Unique identifier for the launched run
+    """
+    tk = await remote.TaskDetails.fetch(project=project, domain=domain, name=name, version=version)
+    r = await flyte.run.aio(tk, **inputs)
```
EngHabu, line 96: So if the post has:
```
{
   "a": true,
   "b": 3
}
```

It expects two inputs to the task? What if I just want to receive them as a single dict input?
```
def my_task(req: dict)
  ...
```

EngHabu (commented):
cc @cosmicBboy if you can answer my questions too

examples/apps/run_webhook.py
kumare3: we can, but that is odd. how can you wait for the task to finish in an http call. it might take long time? I feel it should be a poll option. Ideally you want a "callback url"

EngHabu: They are useful for different scenarios and different constraints...
1. If we dictate our own protocol, callback url makes sense... and also adding other options that we can parse our makes sense..
2. If you are trying to implement an existing protocol and/or if the task runtime is short (and mostly coming back through reusuable containers, waiting might be simpler to code against

examples/apps/run_webhook.py
cosmicbboy: I guess in that case the input to the webhook would be:
```
{"req": {"a": true, "b": 3}}
```

examples/apps/run_webhook.py
cosmicbboy: I think we can write examples for these two cases:
- For fast-running tasks (that complete within let's say 60 seconds), the webhook can wait for the task to complete so the webhook can actually return the result.
- For slow-running tasks, the webhook can simply return the URL and maybe any other metadata needed for the caller to fetch the `Run` object and wait for it to complete on the client side.

PR #314, opened 2025-11-04
Add description to task deploy

popojk:
## Issue
We need to send 4 types of description to the backend when deploying a task:
1. Task environment description
2. Trigger description
3. Task short description
4. Task long description

The first 3 types of description should be truncated at 255 characters before sending to the backend, and the last one should be truncated at 2048.

## Changes

1. Add 4 types of description to the idl when deploying tasks
2. Add unit tests


pyproject.toml
```diff
@@ -10,6 +10,7 @@ dependencies = [
     "click>=8.2.1",
     "flyteidl>=1.15.4b0,<2.0.0",
     "cloudpickle>=3.1.1",
+    "docstring_parser>=0.16",
```
popojk, line 13: 0.16 is the previous version that released in 2024, I choose this version instead of 0.17 that released in 2025 to ensure stability.

src/flyte/_environment.py
```diff
@@ -67,6 +67,8 @@ def __post_init__(self):
             raise TypeError(f"Expected resources to be of type Resources, got {type(self.resources)}")
         if self.env_vars is not None and not isinstance(self.env_vars, dict):
             raise TypeError(f"Expected env_vars to be of type Dict[str, str], got {type(self.env_vars)}")
+        if self.description is not None and len(self.description) > 255:
+            self.description = self.description[:255]
         # Automatically register this environment instance in load order
```
popojk, line 79: We truncated at 255 characters as that is the limit of backend db.

src/flyte/_trigger.py
```diff
@@ -722,6 +722,8 @@ def __post_init__(self):
             raise ValueError("Trigger name cannot be empty")
         if self.automation is None:
             raise ValueError("Automation cannot be None")
+        if self.description and len(self.description) > 255:
+            object.__setattr__(self, "description", self.description[:255])
```
popojk: We truncated at 255 characters as that is the limit of backend db.

src/flyte/_internal/runtime/task_serde.py
```diff
@@ -321,6 +323,27 @@ def _get_k8s_pod(primary_container: tasks_pb2.Container, pod_template: PodTempla
...
+    Short descriptions are truncated to 255 chars, long descriptions to 2048 chars.
+
+    :param task_template: TaskTemplate containing the interface docstring.
+    :return: DescriptionEntity with truncated short and long descriptions.
+    """
+    docstring = task_template.interface.docstring
+    short_desc = None
+    long_desc = None
+    if docstring and docstring.short_description:
+        short_desc = docstring.short_description[:255]
```
popojk: We truncated at 255 characters as that is the limit of backend db.

src/flyte/_internal/runtime/task_serde.py
```diff
@@ -50,7 +50,7 @@ def translate_task_to_wire(
     if task.parent_env and task.parent_env():
         _env = task.parent_env()
         if _env:
-            env = environment_pb2.Environment(name=_env.name[:_MAX_ENV_NAME_LENGTH])
+            env = environment_pb2.Environment(name=_env.name[:_MAX_ENV_NAME_LENGTH], description=_env.description)
```
kumare3: I think we should do it only when deploying and not when running on remote

kumare3: nm, i see it is 255 chars only. But the challenge is this istored in queue service and every byte is precious. But, lets start with this and we can optimize

src/flyte/models.py
```diff
@@ -382,7 +382,11 @@ def from_callable(cls, func: Callable) -> NativeInterface:
 
         # Get return type
         outputs = extract_return_annotation(hints.get("return", sig.return_annotation))
-        return cls(inputs=param_info, outputs=outputs)
+
+        # Parse docstring if available
+        docstring = Docstring(callable_=func) if func.__doc__ else None
```
kumare3, line 387: do we already have this?

kumare3:
Can we get some unit tests for env, trigger and the inputs/outputs. Especially with long descriptions, missing descriptions etc

src/flyte/models.py
popojk: We need to initialize `DocString` object here and registered to NativeInterface so that we can get `DocString` from `TaskTemplate`.

tests/flyte/internal/runtime/test_convert.py
```diff
@@ -42,17 +42,18 @@
 
 
 @pytest.fixture(params=test_cases)
-async def generate_inputs(request) -> Tuple[Inputs, str]:
+def generate_inputs(request) -> Tuple[Inputs, str]:
+    import asyncio
```
popojk: I turn this function into a sync function because I bumped into this error when running unit tests.

cc @wild-endeavor 
截圖 2025-11-25 上午11 35 43

kumare3: Should be async- will check the error closer

src/flyte/_internal/runtime/types_serde.py
```diff
@@ -35,9 +35,9 @@ def transform_native_to_typed_interface(
     input_descriptions: Dict[str, str] = {}
     output_descriptions: Dict[str, str] = {}
     if interface.docstring:
-        # Fill in descriptions from docstring in the future
-        input_descriptions = {}
-        output_descriptions = {}
+        # Extract descriptions from the parsed docstring
```
kumare3: How can we make sure this only runs for deploy

tests/flyte/internal/runtime/test_convert.py
kumare3: ohh you have to use`@pytest.mark.asyncio`

src/flyte/_internal/runtime/types_serde.py
```diff
@@ -10,17 +10,15 @@
```
popojk, line 10: We handle input/output description in src/flyte/_deploy.py instead of in run time as docstring should only be handled at deploy time

src/flyte/_internal/runtime/trigger_serde.py
```diff
@@ -148,6 +148,7 @@ async def to_task_trigger(
             active=t.auto_activate,
             run_spec=run_spec,
             inputs=common_pb2.Inputs(literals=literals),
+            description=t.description,
```
popojk, line 151: `to_task_trigger` function is only called at deploy time, so I think it is ok to insert description into idl here

src/flyte/_deploy.py
```diff
@@ -208,6 +220,62 @@ async def _deploy_task(
...
+    if docstring and docstring.long_description:
+        long_desc = docstring.long_description[:2048]
+    return task_definition_pb2.DocumentationEntity(
+        short_description=short_desc,
+        long_description=long_desc,
+    )
+
+
+def _update_interface_inputs_and_outputs_docstring(
+    typed_interface: "interface_pb2.TypedInterface", native_interface: "NativeInterface"
```
wild-endeavor: don't need "" - if it complains then add the future annotations import.

src/flyte/_deploy.py
```diff
@@ -208,6 +220,62 @@ async def _deploy_task(
...
+) -> "interface_pb2.TypedInterface":
+    """
+    Update the descriptions of inputs and outputs in the TypedInterface from the NativeInterface docstring.
+    This is done during deployment to avoid runtime overhead of parsing docstrings during task execution.
+
+    :param typed_interface: The protobuf TypedInterface to update.
+    :param native_interface: The NativeInterface containing the docstring.
+    :return: Updated TypedInterface with descriptions from docstring.
+    """
+    if not native_interface.docstring:
```
wild-endeavor, line 265: i think we should mutate or return a new one, not both.  if you do both then it becomes confusing at the call site.

popojk: Yea we should just mutate or return new one, fixed

src/flyte/_internal/runtime/types_serde.py
popojk: @kumare3 You are right we should handle the docstring  at deploy time instead of run time. Fixed

src/flyte/_deploy.py
popojk: Of course, fixed

examples/context/custom_context.py
```diff
@@ -2,16 +2,19 @@
```
popojk, line 2: The changes are just lint fix

src/flyte/_deploy.py
wild-endeavor: not fixed right? the function signature hasn't changed.

src/flyte/_trigger.py
pingsutw: Should we raise an error or a warning if it exceeds the limit? @wild-endeavor

tests/flyte/type_engine/test_serde.py
```diff
@@ -56,79 +56,3 @@ def t1() -> File[svg]: ...
     return_type = extract_return_annotation(typing.get_type_hints(t1).get("return", None))
     o0 = return_type["o0"]
     assert issubclass(o0, File)
-
```
popojk, line 59: The docstring will be handled in deploy time, so we remove this test.

src/flyte/_trigger.py
wild-endeavor: i'd vote neither... we should log a debug statement and then replace the last 8 characters of the description with
```
...(tr.)
```

wild-endeavor (approved):

pingsutw (approved):

PR #320, opened 2025-11-06
Add a lightweight JSON formatter

wild-endeavor:
At the request of users who prefer structured log output, this adds an optional json log formatter.  The formatting is just a simple function for now - let's start with this and if it becomes necessary we can pull in an external library.  To use json logging set `LOG_FORMAT=json` when you run it.

Also the `log_format_from_env` function was never used, now we use it.  Also propagate via environment variable to the Run downstream actions.

## To use
From the CLI, either:
```bash
flyte --log-format json ...
```
or
```bash
LOG_FORMAT=json flyte ...
```

Programmatically,
```python
flyte.with_runcontext(log_format="json")
```

### DIFF
--- a/src/flyte/_logging.py
+++ b/src/flyte/_logging.py
@@ -227,13 +227,13 @@ class ContextFilter(logging.Filter):
         if c:
             action = c.action
             # Add as attributes for structured logging (JSON)
-            record.run_name = action.run_name  # type: ignore[attr-defined]
-            record.action_name = action.name  # type: ignore[attr-defined]
+            record.run_name = action.run_name
+            record.action_name = action.name
             # Also modify message for console/Rich output
             record.msg = f"[{action.run_name}][{action.name}] {record.msg}"
         else:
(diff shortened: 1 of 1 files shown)


src/flyte/_logging.py
```diff
@@ -166,13 +220,20 @@ class ContextFilter(logging.Filter):
...
         if c:
             action = c.action
+            # Add as attributes for structured logging (JSON)
+            record.run_name = action.run_name  # type: ignore[attr-defined]
+            record.action_name = action.name  # type: ignore[attr-defined]
+            # Also modify message for console/Rich output
             record.msg = f"[{action.run_name}][{action.name}] {record.msg}"
+        else:
+            record.run_name = None  # type: ignore[attr-defined]
+            record.action_name = None  # type: ignore[attr-defined]
```
pingsutw: Do we really need these `type ignore`? can we remove them?

pingsutw (dismissed):

PR #321, opened 2025-11-06
Build flyte connector image

pingsutw:
- Update `maint_tools/build_default_image.py` to build the connector image
- Update CI to push the connector image
- Add a ConnectorService class to connector/_server.py

Screenshot 2025-11-06 at 2 34 54 PM


maint_tools/build_default_image.py
```diff
@@ -1,30 +1,87 @@
 import argparse
 import asyncio
+from pathlib import Path
 
+import flyte
 from flyte import Image
+from flyte._image import _detect_python_version
 from flyte._internal.imagebuild.image_builder import ImageBuildEngine
```
pingsutw: Can I make `ImageBuildEngine` public? I want users to be able to use the image builder to build their connector images

pingsutw: i think I will add it to flyte extend

src/flyte/connectors/_connector.py
```diff
@@ -183,11 +183,11 @@ def get_connector(task_type_name: str, task_type_version: int = 0) -> AsyncConne
         return ConnectorRegistry._REGISTRY[key]
 
     @staticmethod
-    def list_connectors() -> List[Connector]:
+    def list_connectors() -> List[ConnectorProto]:
```
wild-endeavor: I think it's okay for this pr, but can you spend ten minutes thinking about whether we should expose this to the user?  This type I mean.
* If we expose the proto to the user, that means we're forever bound to it.
* If we expose the proto to the user, we should add it to the top level `connectors/__init__.py` file, in the `__all__`.  I don't think we should be training our users to import from flyteidl2 - it's ugly.  And if we're not comfortable adding it there we shouldn't expose the proto.

wild-endeavor: in fact as part of removing the beta designation I think we should go through everywhere and make sure every proto that we're exposing is
* a deliberate decision
* exposed through an `__all__` somewhere.

src/flyte/extend.py
```diff
@@ -1,14 +1,17 @@
...
 from ._internal.runtime.resources_serde import get_proto_resources
 from ._resources import PRIMARY_CONTAINER_DEFAULT_NAME, pod_spec_from_resources
-from ._task import AsyncFunctionTaskTemplate
+from ._task import AsyncFunctionTaskTemplate, TaskTemplate
 from ._task_plugins import TaskPluginRegistry
 
 __all__ = [
     "PRIMARY_CONTAINER_DEFAULT_NAME",
     "AsyncFunctionTaskTemplate",
+    "ImageBuildEngine",
```
wild-endeavor, line 12: should this be here?  or in the main init file?  `build_images` and `Image` are already exposed there.  I think we should keep these together.  Also does accessing ImageBuildEngine do something that build_images can't do?

plugins/connectors/pyproject.toml
```diff
@@ -7,7 +7,9 @@ authors = [{ name = "Kevin Su", email = "pingsutw@users.noreply.github.com" }]
 requires-python = ">=3.10"
 dependencies = [
     "flyte",
-    "flyteidl"
+    "flyteidl",
```
wild-endeavor: should we remove flyteidl?  it's already part of `flyte`.  And if we remove it from flyte one day then connectors shouldn't need it either right?

src/flyte/connectors/_connector.py
pingsutw: list_connectors is only used internally. wil not expose it to user

wild-endeavor (dismissed):

maint_tools/build_default_image.py
wild-endeavor: yeah

src/flyte/connectors/_connector.py
wild-endeavor: 😈

PR #322, opened 2025-11-07
Scope image build runs to the same target project-domain

EngHabu:

### DIFF
--- a/src/flyte/_internal/imagebuild/remote_builder.py
+++ b/src/flyte/_internal/imagebuild/remote_builder.py
@@ -118,9 +118,12 @@ class RemoteImageBuilder(ImageBuilder):
         else:
             # Use the default system registry in the backend.
             target_image = image_name
+
+        from flyte._initialize import get_init_config
+        cfg = get_init_config()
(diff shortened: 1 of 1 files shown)


src/flyte/_internal/imagebuild/remote_builder.py
```diff
@@ -118,9 +118,12 @@ async def build_image(self, image: Image, dry_run: bool = False) -> str:
...
             # Use the default system registry in the backend.
             target_image = image_name
+
+        from flyte._initialize import get_init_config
+        cfg = get_init_config()
         run = cast(
             Run,
-            await flyte.with_runcontext(project=IMAGE_TASK_PROJECT, domain=IMAGE_TASK_DOMAIN).run.aio(
+            await flyte.with_runcontext(project=cfg.project, domain=cfg.domain).run.aio(
                 entity, spec=spec, context=context, target_image=target_image
```
kumare3, line 127: The weird part is in with run context you. Can override project domain for the run (the task run too)

But for now nobody does that so +1

PR #323, opened 2025-11-07
Don't use local file keyring when not in ipython

wild-endeavor:
* All keyring backends are active by default - this should not be the case for our local file keyring - that one is meant only for notebooks.
* `get_init_config()` can raise an exception if flyte is not initialized, leading to keyring not storing the token and leading to web pages popping open again and again.

### DIFF
--- a/src/flyte/_keyring/file.py
+++ b/src/flyte/_keyring/file.py
@@ -98,7 +98,7 @@ class SimplePlainTextKeyring(KeyringBackend):
             try:
                 config = get_init_config()
                 config_path = config.source_config_path
-                if config_path and str(config_path.parent) == ".flyte":
+                if config_path and str(config_path.parent.name) == ".flyte":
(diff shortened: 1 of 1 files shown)


cosmicBboy (dismissed):

src/flyte/_keyring/file.py
```diff
@@ -72,13 +90,25 @@ def _ensure_file_path(self):
...
-        # otherwise use the default path
+        from flyte._initialize import get_init_config, is_initialized
+        from flyte._logging import logger
+
+        # Only try to use source_config_path if flyte.init() has been called
+        if is_initialized():
+            try:
+                config = get_init_config()
+                config_path = config.source_config_path
+                if config_path and str(config_path.parent) == ".flyte":
```
kumare3: Do you want to check if .flyte is in that path?

pingsutw (dismissed):

PR #329, opened 2025-11-10
Update trigger_serde

pingsutw:


src/flyte/_internal/runtime/trigger_serde.py
```diff
@@ -12,7 +12,10 @@
 def _to_schedule(m: Union[Cron, FixedRate], kickoff_arg_name: str | None = None) -> common_pb2.Schedule:
     if isinstance(m, Cron):
         return common_pb2.Schedule(
-            cron_expression=m.timezone_expression,
```
iaroslav-ciupin, line 15: do you still need timezone_expression helper method? can you remove it?

pingsutw: I'm not sure if we can remove that. I'll keep it for now

PR #334, opened 2025-11-10
feat(maint_tools): Add bigquery extra to flyteplugins connector

pingsutw:


maint_tools/build_default_image.py
```diff
@@ -53,7 +53,7 @@ async def build_flyte_connector_image(
         )
     else:
         default_image = Image.from_debian_base(registry=registry, name=name).with_pip_packages(
-            "flyteplugins-connectors", pre=True, extra_args="--all-extras"
```
pingsutw, line 56: all-extras only works with pyproject.toml locally

maint_tools/build_default_image.py
```diff
@@ -53,7 +53,7 @@ async def build_flyte_connector_image(
...
+            "flyteplugins-connectors[bigquery]", pre=True
```
kumare3, line 56: Why are all connectors in one package? Why are we not creating one package per connector. This will Become a maintenance nightmare

kumare3 (approved):

PR #335, opened 2025-11-11
[Feature] Add task git repo host uri when deploying task

popojk:
## Changes
We will store flyte task git repo host url in the backend when sending a deploy request request to the run service. The url will be respond in the GetTaskDetails api. 

How the url is defined:
1. If task code file is under git repo and the git tree is clean, the url will be suffixed with line number.
2. If task code file is under git repo and the git tree is not clean, the url will only points to the code file in remote repo.
3. If task code file is not under git repo, the url will not be sent to the backend.
4. The SDK will choose origin as default remote. If origin remote does not exist, the SDK will sort remotes alphabetically and choose the first one.

## Tests
Git operation is related to local environment, which is not proper for unit tests. Thus, in this PR we test it manually

### Github remote project test:
deploy a project with github remotes, we expect that the github link url should be the one linked to remote "origin"
截圖 2025-12-04 上午10 40 31

The github link url is stored in the backend and respond by GetTaskDetails api
截圖 2025-12-04 上午10 43 25

The link can navigate us to the remote repo with right file and right line of code
截圖 2025-12-04 上午10 44 26

### Gitlab remote project test:
deploy a project with gitlab remotes
截圖 2025-12-04 上午10 47 47

The gitlab link will navigate us to to the remote repo with right file and right line of code
截圖 2025-12-04 上午11 02 48

### Deploy a task not in git repo:
The task will still be deployed, but no git link url will be stored and a warning will shown in the terminal
截圖 2025-12-04 上午11 07 41

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,7 @@ dependencies = [
 flyte = "flyte.cli.main:main"
 a0 = "flyte._bin.runtime:main"
 fserve = "flyte._bin.serve:main"
+c0 = "flyte._bin.connect:main"
 
 [project.entry-points."keyring.backends"]
 flyte_keyring_file = "flyte._keyring.file"
--- a/src/flyte/_bin/connect.py
+++ b/src/flyte/_bin/connect.py
@@ -1,0 +1,61 @@
+from typing import List
+
+import click
+
+
+@click.group()
+def _connect():
+    """Debug commands for Flyte."""
+
+
+@_connect.command()
+@click.option(
+    "--port",
+    default="8000",
+    is_flag=False,
+    type=int,
+    help="Grpc port for the connector service",
+)
+@click.option(
+    "--prometheus_port",
+    default="9090",
+    is_flag=False,
+    type=int,
+    help="Prometheus port for the connector service",
+)
+@click.option(
+    "--worker",
+    default="10",
+    is_flag=False,
+    type=int,
+    help="Number of workers for the grpc server",
+)
+@click.option(
+    "--timeout",
+    default=None,
+    is_flag=False,
+    type=int,
+    help="It will wait for the specified number of seconds before shutting down grpc server. It should only be used "
+    "for testing.",
+)
+@click.option(
+    "--modules",
+    required=False,
+    multiple=True,
+    type=str,
+    help="List of additional files or module that defines the connector",
+)
+@click.pass_context
+def main(
+    _: click.Context, port: int, prometheus_port: int, worker: int, timeout: int | None, modules: List[str] | None
+):
+    """
+    Start a grpc server for the connector service.
+    """
+    from flyte.connectors import ConnectorService
+
+    ConnectorService.run(port, prometheus_port, worker, timeout, modules)
+
+
+if __name__ == "__main__":
+    _connect()
--- a/src/flyte/connectors/utils.py
+++ b/src/flyte/connectors/utils.py
@@ -130,7 +130,7 @@ def _load_connectors(modules: List[str] | None):
             logger.info(f"Loading connector: {ep.name}")
             ep.load()
         except Exception as e:
-            logger.warning(f"Failed to load type transformer {ep.name} with error: {e}")
+            logger.warning(f"Failed to load connector '{ep.name}' with error: {e}")
 
     if modules:
         for m in modules:


src/flyte/_initialize.py
```diff
@@ -9,6 +9,7 @@
 
 from flyte.errors import InitializationError
 from flyte.syncify import syncify
+from flyte.git import GitConfig
```
kumare3: Please do not import at module level

src/flyte/_internal/runtime/task_serde.py
```diff
@@ -47,6 +48,17 @@ def translate_task_to_wire(
     """
     tt = get_proto_task(task, serialization_context)
     env: environment_pb2.Environment | None = None
+
+    # Todo: Make document_entity with description here
+
+    if hasattr(task.func, '__code__') and task.func.__code__:
```
kumare3: I do not think we should add it to every task generation. It should only ever be done for deployment. Runtime has to be fast. And docs are not needed

src/flyte/git/_config.py
```diff
@@ -1,7 +1,153 @@
 import pathlib
 import subprocess
+from pathlib import Path
 
 import flyte.config
+from flyte._logging import logger
+
+
+class GitConfig:
```
kumare3: Please move this to flyte.git

kumare3 (commented):
Please first document the cli and the flyte.deploy examples. This should also be optional and not fail not found, but clearly explain.

All got related stuff is cleanly separated into flyte.git
And it is not part of runtime. Runtime Has to be as fast as possible. Also cleanly separate paths 

Why do we need this in initialize? Only deployment should ever need this

src/flyte/git/_config.py
```diff
@@ -1,13 +1,258 @@
...
+
+    @staticmethod
+    def build_url(remote_url: str, file_path: str, commit_sha: str, line_number: int, is_tree_clean: bool) -> str:
+        url = f"{remote_url}/-/blob/{commit_sha}/{file_path}"
+        if is_tree_clean:
+            url += f"#L{line_number}"
+        return url
+
+
+GIT_URL_BUILDER_REGISTRY: Dict[str, GitUrlBuilder] = {
```
popojk, line 38: Use a registry pattern here so that we can add more builders to support more git remote hosts in the future.

src/flyte/_deploy.py
```diff
@@ -226,24 +227,38 @@ async def _deploy_task(
...
         short_desc = parse_description(docstring.short_description, 255)
     if docstring and docstring.long_description:
         long_desc = parse_description(docstring.long_description, 2048)
+    if hasattr(task_template, "func") and hasattr(task_template.func, "__code__") and task_template.func.__code__:
+        line_number = (
+            task_template.func.__code__.co_firstlineno + 1
+        )  # The function definition line number is located at the line after @env.task decorator
+        file_path = task_template.func.__code__.co_filename
+        git_config = GitConfig()
+        if git_config.is_valid:
```
popojk: The deploy will not fail even though git_config is not valid.

src/flyte/_initialize.py
popojk: This change is removed

src/flyte/_internal/runtime/task_serde.py
popojk: This change is removed

src/flyte/git/_config.py
```diff
@@ -1,13 +1,258 @@
...
+    GithubUrlBuilder.host_name: GithubUrlBuilder,
+    GitlabUrlBuilder.host_name: GitlabUrlBuilder,
+}
+
+
+class GitConfig:
```
wild-endeavor: I think this is better named as GitStatus or something right?  GitConfig makes me think of my `~/.gitconfig` file.

src/flyte/git/_config.py
```diff
@@ -1,13 +1,258 @@
...
+
+        :return: None
+        """
+        self.is_valid = False
+        self.is_tree_clean = False
+        self.remote_url = ""
+        self.repo_dir = Path.cwd()
+        self.commit_sha = ""
+
+        try:
```
wild-endeavor, line 70: i think cleaner to move all this to a separate class constructor method.  Leave init as just barebones, taking in the values in the init function.

src/flyte/git/_config.py
popojk: Renamed to GitStatus, thanks!

src/flyte/git/_config.py
popojk: I just wrapped the construction logics into `from_current_repo` function, thanks!

popojk:
> looks good for the most part... some of the functions in _config lend very well to unit testing. could you add some where it makes sense to do so?

@wild-endeavor I've added some unit tests to the new feature, PTAL thanks!

kumare3:
Thank you looks great to me

PR #348, opened 2025-11-20
fix(images): Support custom image URIs as strings in TaskEnvironment

pingsutw:
```python
import flyte

# TaskEnvironments provide a simple way of grouping configuration used by tasks (more later).
env = flyte.TaskEnvironment(name="hello_world", resources=flyte.Resources(memory="250Mi"), image="376129846803.dkr.ecr.us-east-2.amazonaws.com/union/dogfood:flyte-07640f2b7ea376512801786964ec8072")


# use TaskEnvironments to define tasks, which are regular Python functions.
@env.task
def fn(x: int) -> int:  # type annotations are recommended.
    slope, intercept = 2, 5
    return slope * x + intercept


# tasks can also call other tasks, which will be manifested in different containers.
@env.task
def main(x_list: list[int]) -> float:
    x_len = len(x_list)
    if x_len < 10:
        raise ValueError(f"x_list doesn't have a larger enough sample size, found: {x_len}")

    y_list = list(flyte.map(fn, x_list))  # flyte.map is like Python map, but runs in parallel.
    y_mean = sum(y_list) / len(y_list)
    return y_mean


if __name__ == "__main__":
    flyte.init_from_config()  # establish remote connection from within your script.
    run = flyte.run(main, x_list=list(range(10)))  # run remotely inline and pass data.

    # print various attributes of the run.
    print(run.name)
    print(run.url)

    run.wait()  # stream the logs from the root action to the terminal.
```

### DIFF
--- a/examples/apps/call_apps_in_tasks/app.py
+++ b/examples/apps/call_apps_in_tasks/app.py
@@ -36,4 +36,5 @@ async def add_one_task(x: int) -> int:
     print(f"Calling app at {app_env.endpoint}")
     async with httpx.AsyncClient() as client:
         response = await client.get(app_env.endpoint, params={"x": x})
+        response.raise_for_status()
         return response.json()["result"]
--- a/examples/context/custom_context.py
+++ b/examples/context/custom_context.py
@@ -1,0 +1,19 @@
+import flyte
+
+env = flyte.TaskEnvironment("custom-context-example")
+
+@env.task
+async def leaf_task() -> str:
+    # Reads run-level context
+    print("leaf sees:", flyte.ctx().custom_context)
+    return flyte.ctx().custom_context.get("trace_id")
+
+@env.task
+async def root() -> str:
+    return await leaf_task()
+
+if __name__ == "__main__":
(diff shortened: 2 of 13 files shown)


src/flyte/_internal/runtime/task_serde.py
```diff
@@ -239,14 +239,14 @@ def _get_urun_container(
...
-        raise flyte.errors.RuntimeSystemError("BadConfig", "Image is not a valid image")
-
     env_name = task_template.parent_env_name
     if env_name is None:
         raise flyte.errors.RuntimeSystemError("BadConfig", f"Task {task_template.name} has no parent environment name")
 
-    img_uri = lookup_image_in_cache(serialize_context, env_name, img)
+    if isinstance(img, flyte.Image):
+        img_uri = lookup_image_in_cache(serialize_context, env_name, img)
+    else:
```
kumare3: lets not deal with strings at all in this layer. lets look for image has no layers then just use the base image
if this is not true we should raise an error

kumare3 (approved):

PR #349, opened 2025-11-21
Call Apps from tasks.

kumare3:


examples/basics/devbox_one.py
```diff
@@ -35,6 +35,7 @@ async def say_hello_nested(data: str = "default string", n: int = 3) -> str:
 
 if __name__ == "__main__":
     flyte.init_from_config()
+    # What?
```
cosmicbboy: delete?

kumare3: Ya

src/flyte/app/_app_environment.py
```diff
@@ -147,9 +147,11 @@ def get_port(self) -> Port:
 
     @property
     def endpoint(self) -> str:
+        import flyte
+
```
cosmicbboy: what is this import for?

kumare3: gone

examples/basics/devbox_one.py
kumare3: removed

PR #357, opened 2025-11-25
Add support tool-uv-index in the remote builder

pingsutw:


src/flyte/_internal/imagebuild/remote_builder.py
```diff
@@ -269,10 +269,6 @@ def _get_layers_proto(image: Image, context_path: Path) -> "image_definition_pb2
             )
             layers.append(pip_layer)
         elif isinstance(layer, UVProject):
-            for line in layer.pyproject.read_text().splitlines():
```
pingsutw, line 272: This constraint is only for serverless

PR #366, opened 2025-12-02
fix run and syncify typing

cosmicBboy:
This PR fixes the `syncify` and `run` type signatures such that the `run` output is correctly revealed to be of type `Run` instead of `Any`:
image

### DIFF
--- a/src/flyte/remote/_app.py
+++ b/src/flyte/remote/_app.py
@@ -141,7 +141,7 @@ class App(ToJSONMixin):
         """
         if self.is_active():
             return self
-        return await self.update(
+        return await self.update.aio(
             app_definition_pb2.Spec.DESIRED_STATE_STARTED,
             "User requested to activate app from flyte-sdk",
             "activated" if wait else None,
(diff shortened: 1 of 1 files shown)


src/flyte/_run.py
```diff
@@ -541,19 +542,21 @@ def __init__(self, outputs: Tuple[Any, ...] | Any):
...
                 quiet: bool = False,
                 wait_for: Literal["terminal", "running"] = "terminal",
-            ):
+            ) -> None:
                 pass
 
-            def outputs(self) -> R:
-                return cast(R, self._outputs)
+            @syncify
+            async def outputs(self) -> ActionOutputs:  # type: ignore[override]
```
kumare3, line 554: We will have to write outputs.aio I guess

src/flyte/cli/_get.py
```diff
@@ -227,10 +227,11 @@ async def _run_log_view(_obj):
         except KeyboardInterrupt:
             task.cancel()
 
+    obj: Union[remote.Action, remote.Run]
```
kumare3, line 230: This should never return action

cosmicbboy: hmm, if the user supplies `action_name`, it looks like on line 231 `obj` will be an Action, no?

src/flyte/_run.py
cosmicbboy: for async, they need to write `outputs.aio()`, but for sync it should be just `outputs()` right?

src/flyte/cli/_get.py
kumare3: aah right, sorry in that case we are reading an action! got it

PR #369, opened 2025-12-02
update recsys example

cosmicBboy:


examples/ml/recsys/recsys_serving.py
```diff
@@ -397,17 +397,34 @@ async def get_item_info(item_id: str):
...
+    #
+    # Then supplying the run name to the serving script
+    #
+    # `uv run recsys_serving.py --training-run <training-run-id>`
+    #
+    # Call the endpoint:
+    #
+    # `curl -X POST https://<app-url>/embed-text "Content-Type: application/json" '{"text": "Testing testing 123"}'`
+
+    app = flyte.with_servecontext(env_vars={"ARTIFACTS_DIR": artifacts_dir.path}).serve(env)
```
kumare3, line 429: I don't think this works. The env is evaluated locally

PR #377, opened 2025-12-04
App input from task run: delayed input values

cosmicBboy:
This PR implements a `_DelayedValue` input type, which will materialize values at app startup time instead of serving time.


src/flyte/app/_input.py
```diff
@@ -1,19 +1,162 @@
...
+    type: _SerializedInputType
+
+    @model_validator(mode="before")
+    @classmethod
+    def check_type(cls, data: typing.Any) -> typing.Any:
+        if "type" in data:
+            data["type"] = INPUT_TYPE_MAP.get(data["type"], data["type"])
+        return data
+
+    async def get(self) -> str:
```
kumare3, line 42: Can we put some logs, in debug mode and also a warning log, after it is materialized for the name of the key that was materialized from remote.

PR #379, opened 2025-12-05
Implement VLLMAppEnvironment and SafeTensorStreaming

cosmicBboy:
This PR introduces a `VLLMAppEnvironment` app environment to support serving vllm apps, and a `SafeTensorStreaming` class to load weights directly from remote storage into memory.

### DIFF
--- a/src/flyte/_bin/serve.py
+++ b/src/flyte/_bin/serve.py
@@ -35,7 +35,7 @@ async def sync_inputs(serialized_inputs: str, dest: str) -> Tuple[dict, dict]:
         The environment variables dictionary maps environment variable names to their values.
     """
     import flyte.storage as storage
-    from flyte.app._input import AppEndpoint, RunOutput, SerializableInputCollection
+    from flyte.app._input import SerializableInputCollection
 
     print(f"Log level: {logger.getEffectiveLevel()} is set from env {os.environ.get('LOG_LEVEL')}", flush=True)
     logger.info("Reading inputs...")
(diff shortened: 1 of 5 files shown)


examples/genai/vllm/cache_model.py
```diff
@@ -0,0 +1,153 @@
...
+
+        # Upload the model directory to remote storage
+        model_dir = await flyte.io.Dir.from_local(local_model_path)
+        print(f"Model cached at: {model_dir.path}")
+
+        return model_dir
+
+
+@cache_env.task
+async def main(
```
kumare3, line 127: why we do need this task? Why not just make it one task, faster more efficient?

pyproject.toml
```diff
@@ -33,6 +33,7 @@ dependencies = [
 flyte = "flyte.cli.main:main"
 a0 = "flyte._bin.runtime:main"
 fserve = "flyte._bin.serve:main"
+vllm-fserve = "flyte.app.extras._vllm._model_loader.shim:main"
```
kumare3: I think we should make this a plugin. Lets not add vllm to the core

plugins/vllm/src/flyteplugins/vllm/_app_environment.py
```diff
@@ -0,0 +1,112 @@
+import shlex
```
kumare3, line 1: Lets not put vllm directly into core. Lets keep it in a plugin. you can also simply put it into the union plugin, but otherwise just put it in plugin

PR #380, opened 2025-12-05
Add H200 partitions

davidmirror-ops:
Adding the supported MIG profiles for the NVIDIA H200 device.

Tested like this

```
acc_env = flyte.TaskEnvironment(
    name="mig",
    resources=flyte.Resources(gpu=flyte.GPU(device="H200", quantity=1, partition="2g.35gb")),
    image=custom_image,
 
)

```

### DIFF
--- a/src/flyte/_resources.py
+++ b/src/flyte/_resources.py
@@ -23,6 +23,11 @@ A100_80GBParts = Literal["1g.10gb", "2g.20gb", "3g.40gb", "4g.40gb", "7g.80gb"]
 Partitions for NVIDIA A100 80GB GPU.
 """
 
+H200Parts = Literal["1g.18gb", "1g.35gb", "2g.35gb", "3g.71gb", "4g.71gb", "7g.141gb"] 
+"""
+Partitions for NVIDIA H200 GPU (141GB HBM3e).
+"""
+
 TPUType = Literal["V5P", "V6E"]
 V5EParts = Literal["1x1", "2x2", "2x4", "4x4", "4x8", "8x8", "8x16", "16x16"]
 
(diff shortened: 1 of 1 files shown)


src/flyte/_resources.py
```diff
@@ -249,7 +254,7 @@ def __post_init__(self):
             raise ValueError("GPU quantity must be at least 1")
 
 
-def GPU(device: GPUType, quantity: GPUQuantity, partition: A100Parts | A100_80GBParts | None = None) -> Device:
+def GPU(device: GPUType, quantity: GPUQuantity, partition: A100Parts | A100_80GBParts | H200Parts | None = None) -> Device:
```
jeevb, line 257: Maybe worth creating a type alias for these MIG types.

src/flyte/_resources.py
```diff
@@ -23,6 +23,11 @@
 Partitions for NVIDIA A100 80GB GPU.
 """
 
+H200Parts = Literal["1g.18gb", "1g.35gb", "2g.35gb", "3g.71gb", "4g.71gb", "7g.141gb"]
```
jeevb, line 26: Didn't cross-check these. Looks reasonable though.

davidmirror-ops: Yep, come from here
https://docs.nvidia.com/datacenter/tesla/mig-user-guide/supported-mig-profiles.html#h200-mig-profiles

src/flyte/_resources.py
kumare3: Why not just keep it partitions?

jeevb: Thinking we can group this union types into a single alias. As we add more partition types, function def will look cleaner. Def not critical.

PR #384, opened 2025-12-08
Update `flyte gen` to detect plugins

kumare3:


src/flyte/cli/_gen.py
```diff
@@ -88,15 +88,47 @@ def walk_commands(ctx: click.Context) -> Generator[Tuple[str, click.Command, cli
...
+        return False, None
+
+    module = cmd.callback.__module__
+    if "flyte." not in module:
+        # External plugin
+        parts = module.split(".")
+        if len(parts) == 1:
+            return True, parts[0]
+        return True, f"{parts[0]}.{parts[1]}"
+    elif module.startswith("flyte.") and not module.startswith("flyte.cli"):
```
wild-endeavor, line 108: can skip the and not right?

PR #392, opened 2025-12-09
removing unwanted section from grouping example files

debasisdwivedy:
Hi,

The `grouping.py` example file have redundant and unnecessary section, which has been removed for clarity.

Regards,


kumare3:
Its not unwanted, but shows you 2 ways of doing this?

debasisdwivedy:
Got it !!!

PR #399, opened 2025-12-10
Add SGLangAppEnvironment

cosmicBboy:
This PR:
- Add an sglang plugin with an `SGLangAppEnvironment` object to easily deploy sglang apps
- Fixes a bug in the Obstore parallel reader to take into account source offset for streaming tensors

### DIFF
--- a/examples/basics/types/optional_int_collection.py
+++ b/examples/basics/types/optional_int_collection.py
@@ -10,17 +10,17 @@ def main(numbers: Optional[list[int]] = None) -> str:
     """Process an optional collection of integers"""
     if numbers is None:
         return "No numbers provided (None)"
-    
+
     if not numbers:
         return "Empty list provided"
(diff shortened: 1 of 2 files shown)


plugins/sglang/src/flyteplugins/sglang/_app_environment.py
```diff
@@ -0,0 +1,114 @@
...
+    """
+
+    port: int = 8000
+    type: str = "SGLang"
+    extra_args: Union[str, List[str]] = ""
+    model_path: str | RunOutput = ""
+    model_hf_path: str = ""
+    model_id: str = ""
+    stream_model: bool = True
+    _model_mount_path: str = field(default="/root/flyte", init=False)
```
kumare3, line 50: why this? why not use inputs in the same way?

PR #406, opened 2025-12-10
adding blake2b instead of md5 for collision proof hashing

debasisdwivedy:
Hi,

Below is the modification from `md5` hashing to `blake`. Could we give it a try.

MD5 :-> Collision, unsafe, fast
blake2b -> no collision, safe, faster
sha256 -> no collision, safe, slower
blake3 -> no collision, safe, fastest; but will require installation of `blake3 (pip install blake3)` package as it is not yet part of the standard `hashlib`. This is very good for file content hashing.

Regards

### DIFF
--- a/src/flyte/_code_bundle/_packaging.py
+++ b/src/flyte/_code_bundle/_packaging.py
@@ -168,7 +168,7 @@ def compute_digest(source: Union[os.PathLike, List[os.PathLike]], filter: Option
     :param callable filter:
     :return Text:
     """
-    hasher = hashlib.md5()
+    hasher = hashlib.blake2b()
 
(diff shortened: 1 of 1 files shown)


src/flyte/_code_bundle/_packaging.py
```diff
@@ -168,7 +168,7 @@ def compute_digest(source: Union[os.PathLike, List[os.PathLike]], filter: Option
     :param callable filter:
     :return Text:
     """
-    hasher = hashlib.md5()
+    hasher = hashlib.blake2b()
```
kumare3, line 171: sadly we cannot do this, as the blob stores only support md5 only for signed urls

debasisdwivedy: Ahhh. The sad reality to have backward compatibility.

PR #414, opened 2025-12-11
Notifications for triggers and runs (local and remote)

kumare3:


examples/triggers/notifications.py
```diff
@@ -0,0 +1,56 @@
...
+trig2 = flyte.Trigger(
+    name="hourly",
+    auto_activate=True,
+    automation=flyte.Cron("* * * * *"),
+    notifications=(
+        flyte.notify.Slack(
+            on_phase=(
+                "FAILED",
+                "TIMED OUT",
+            ),
```
squiishyy: Is there some way to pass in list of enums instead of strings?

kumare3: enums we avoid as you have to import so we use literals. these are literals

kumare3: I can make them enums, no problem though

kumare3: I am thinking, if we do this, then all phases will be in a common model for all of user facing apis and we can then use them everywhere

cosmicbboy: Can we re-use the phase literals as defined here? https://github.com/flyteorg/flyte-sdk/blob/main/src/flyte/remote/_run.py#L21-L23 (i.e. lower-cased)

kumare3: Ohh ya - but I think I will
Move them to a common place

examples/triggers/notifications.py
```diff
@@ -0,0 +1,74 @@
...
+
+env = flyte.TaskEnvironment(
+    name="example_task",
+)
+
+trig1 = flyte.Trigger(
+    name="hourly",
+    auto_activate=True,
+    automation=flyte.Cron("* * * * *"),
+    notifications=flyte.notify.Slack(
```
iaroslav-ciupin, line 14: 1. Is this saying that notifications will ONLY be sent for failed **triggered** runs? 
What if I want to configure notifications even for runs that are manually created(via Web or CLI). Lets say, its an important workflow on prod domain and everybody in the team must be aware about failures even If I'm the author of the run and I saw the failure.

2. Are the notifications only attached to individual tasks? 
do we want to allow configuring notifications for entire domain(prod), project+domain(<important_project> on prod) or maybe project+domain+task_regex?
What if I want to configure your notifications once, without repeating same configs on multiple task definitions? 

As a follow up for 1 and 2, should notifications be a separate SDK entity in which you can configure multiple matching criteria independently? project, domain, task_regex, triggered vs manual, maybe others in future?

examples/triggers/notifications.py
```diff
@@ -0,0 +1,74 @@
...
+            on_phase=(
+                "FAILED",
+                "TIMED OUT",
+            ),
+            webhook_url="https://webhook.site/",
+            message="Hello world! from {task.name}",
+        ),
+        flyte.notify.Webhook(
+            on_phase="SUCCEEDED",
+            headers={"Content-Type": "application/json"},
```
iaroslav-ciupin, line 53: 1. should webhooks URLs be accessible without any auth credentials? if not, what should be the DevUX for customer to provide credentials for his webhook securely to Union? and I guess Union notifications service(workers) should be able to read those secure credentials during the notification delivery.
2. same question about Slack API. Not familiar with it, but do we need some token to hit the API? who and how will provide this token to us?

examples/triggers/notifications.py
```diff
@@ -0,0 +1,74 @@
...
+            url="https://webhook.site/",
+            body={
+                "xyz": "{run.name}",
+            },
+        ),
+        flyte.notify.Email(
+            on_phase="SUCCEEDED",
+            subject="Hello world! from {task.name}",
+            body="Hello world! from {task.name}",
```
iaroslav-ciupin, line 62: just want to double check the DevUX here: if I want the same email template to be sent for all my failed tasks with some variables like `<task_name>`, `<domain>`, `<project>`, etc. , I can simply store this template in a Python variable, and then re-use it in all Email notification configs? 
not sure what email provider are we going to use or if its critical right now, but, for example, AWS SES supports both sending raw emails with entire HTML body and sending emails with template id, that refers to previously uploaded template.

examples/triggers/notifications.py
jpvotta: I would like it to be possible to notify at the organization, domain, and project-domain levels as this is a frequent request we got in v1 over the past few years. People want to be able to notify on any failure in the prod domain, for instance. I don't know that this makes sense to include in the SDK, though - I feel like this makes sense in the UI for sure, as well as the API/CLI (which can be used by a terraform provider).

examples/triggers/notifications.py
jpvotta: I think we will stick to unauthenticated webhooks (or more specifically, webhooks which embed the auth into the webhook itself)

examples/triggers/notifications.py
kumare3: I moved them to a common place now, will update PR

examples/triggers/notifications.py
kumare3: +1 to Votta. With scoping for now to be triggered runs only

examples/triggers/notifications.py
kumare3: +1

examples/triggers/notifications.py
kumare3: Just render on client side. As this may change all the time

examples/triggers/notifications.py
wild-endeavor: given that we really only do email for now even for slack, should we just do email?

examples/triggers/notifications.py
```diff
@@ -0,0 +1,71 @@
...
+
+trig1 = flyte.Trigger(
+    name="hourly",
+    auto_activate=True,
+    automation=flyte.Cron("* * * * *"),
+    notifications=flyte.notify.Slack(
+        on_phase=ActionPhase.FAILED,
+        webhook_url="https://webhook.site/",
+        message="Hello world! from {task.name}",
+        blocks=(
```
wild-endeavor, line 18: can we drop this for now, until customers ask?  just wary of supporting too many features out of the gate, and then we have to mark things deprecated if they ever change.

examples/triggers/notifications.py
```diff
@@ -0,0 +1,71 @@
...
+env = flyte.TaskEnvironment(
+    name="example_task",
+)
+
+trig1 = flyte.Trigger(
+    name="hourly",
+    auto_activate=True,
+    automation=flyte.Cron("* * * * *"),
+    notifications=flyte.notify.Slack(
+        on_phase=ActionPhase.FAILED,
```
wild-endeavor, line 15: should we just call this `on`?

examples/triggers/notifications.py
```diff
@@ -0,0 +1,71 @@
...
+        webhook_url="https://webhook.site/",
```
wild-endeavor, line 16: can we change the name maybe so that it doesn't cause any confusion with the Webhook notification?  `slack_endpoint` or `service_url` or something?

src/flyte/notify/__init__.py
```diff
@@ -0,0 +1,77 @@
...
+Available Notification Types:
+    - Email: Send email notifications
+    - Slack: Send Slack messages (with optional Block Kit)
+    - Teams: Send Microsoft Teams messages (with optional Adaptive Cards)
+    - Webhook: Send custom HTTP requests (most flexible)
+
+Supported Phases:
+    - "SUCCEEDED": Task completed successfully
+    - "FAILED": Task failed
+    - "TIMED_OUT": Task timed out
```
wild-endeavor: will lease failovers count here?

kumare3: not right now only terminal. I think we should also drop support for Queued and Running, cc @iaroslav-ciupin / @EngHabu

iaroslav-ciupin: I don't know what is a lease failover.
I think the original requirements was to ONLY send notifications for terminal root actions(when entire run completes).

PR #416, opened 2025-12-11
stress-test: parallel image builds

pingsutw:


examples/stress/image_builds.py
```diff
@@ -0,0 +1,72 @@
...
+
+    # Run all builds in parallel
+    builds = await asyncio.gather(*build_tasks)
+
+    # Print results
+    print("\n✅ All 10 image builds triggered successfully!")
+    for i, build in enumerate(builds, 1):
+        print(f"\nBuild {i}/10:")
+        print(f"  Name: {build.name}")
+        print(f"  URL: {build.url}")
```
yuhuan130: This test is amazing, it is just there's something I am aware of when running this test script. flyte.build.aio(image) returns an Image URI, which might cause an error at Line 64 and Line 65. Not sure if this'll help (? 🙌

pingsutw: Good catch, thanks!

kumare3:
cool!

PR #418, opened 2025-12-12
Add flyte store verb with hf-model noun to store huggingface models

cosmicBboy:
This PR:
- Adds a `flyte store` CLI with a `hf-model` subcommand to store huggingface models into flyte object store
- Adds a `flyte.store.hf_model` function that does the same thing as the CLI
- Adds `clone_with` method to AppEnvironment, VLLMAppEnvironment, and SGLangAppEnvironment
- Updates the vllm and sglang examples

### DIFF
--- a/src/flyte/cli/_prefetch.py
+++ b/src/flyte/cli/_prefetch.py
@@ -28,6 +28,13 @@ def prefetch():
 @prefetch.command(name="hf-model", cls=CommandBase)
 @click.argument("repo", type=str)
 @click.option(
+    "--s3-path",
+    type=str,
+    required=False,
+    default=None,
+    help="S3 path to store the model. If not provided, the model will be stored using the default Flyte storage layer.",
+)
+@click.option(
     "--artifact-name",
     type=str,
     required=False,
@@ -138,6 +145,7 @@ def prefetch():
 def hf_model(
     cfg,
     repo: str,
+    s3_path: str | None,
     artifact_name: str | None,
     architecture: str | None,
     task: str,
@@ -229,6 +237,7 @@ def hf_model(
     console.print("[bold green]Starting model prefetch task...")
     run = prefetch_hf_model(
         repo=repo,
+        s3_path=s3_path,
         artifact_name=artifact_name,
         architecture=architecture,
         task=task,
--- a/src/flyte/prefetch/_hf_model.py
+++ b/src/flyte/prefetch/_hf_model.py
@@ -18,16 +18,11 @@ from pydantic import BaseModel, Field
 
 from flyte._logging import logger
 from flyte._task_environment import TaskEnvironment
-from flyte._utils.lazy_module import lazy_module
 from flyte.io import Dir
 
 if TYPE_CHECKING:
     from flyte._resources import Resources
     from flyte.remote import Run
-
-
-huggingface_hub = lazy_module("huggingface_hub")
-vllm = lazy_module("vllm")
 
 
 DEFAULT_SHARD_PATTERN = "model-rank-{rank}-part-{part}.safetensors"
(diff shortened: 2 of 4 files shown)


src/flyte/cli/_prefetch.py
```diff
@@ -0,0 +1,264 @@
...
+    "--shard-config",
+    type=click.Path(exists=True, path_type=Path),
+    help=(
+        "Path to a YAML file containing sharding configuration. "
+        "The file should have 'engine' (currently only 'vllm') and 'args' keys."
+    ),
+)
+@click.pass_obj
+def hf_model(
+    cfg,
```
kumare3, line 146: can you also pass raw-data-storage-path like s3://. This is optional. If passed in we will store here.

src/flyte/prefetch/_hf_model.py
```diff
@@ -0,0 +1,564 @@
...
+    cache: CacheRequest = "disable" if force > 0 else "auto"
+    env = TaskEnvironment(
+        name="prefetch-hf-model",
+        image=image,
+        resources=resources,
+        secrets=[Secret(key=hf_token_key, as_env_var="HF_TOKEN")],
+        cache=cache,
+    )
+    task = env.task(report=True)(store_hf_model_task)  # type: ignore [assignment]
+    run = flyte.with_runcontext(interactive_mode=True).run(task, info.model_dump_json())
```
kumare3: Why interactive. Little hard to debug if the it breaks. Can we make it a file local method?

src/flyte/prefetch/_hf_model.py
```diff
@@ -0,0 +1,564 @@
...
+            )
+            .with_pip_packages(*VLLM_SHARDING_IMAGE_PACKAGES)
+        )
+    else:
+        image = flyte.Image.from_debian_base(name="prefetch-hf-model-image").with_pip_packages(
+            *HF_DOWNLOAD_IMAGE_PACKAGES
+        )
+
+    # Create a task from the module-level function with the configured environment
+    cache: CacheRequest = "disable" if force > 0 else "auto"
```
kumare3: you can disable cache in with_runcontext

src/flyte/prefetch/_hf_model.py
kumare3: you can also override the resources before running if you want

cosmicbboy: File local method errors out with `src` module not found. looks like code bundling specifically excludes files inside the `src.flyte` package.

We should tackle this as a separate PR.

cosmicbboy: where? `with_runcontext` has no resource option

kumare3: No the task of you want before you give it to flyte.run

src/flyte/app/_app_environment.py
```diff
@@ -1,17 +1,23 @@
...
 
 import rich.repr
 
-from flyte import Environment
+from flyte import Environment, Image, Resources, SecretRequest
 from flyte.app._input import Input
 from flyte.app._types import Domain, Link, Port, Scaling
 from flyte.models import SerializationContext
 
+if TYPE_CHECKING:
```
kumare3, line 17: delete?

src/flyte/cli/_prefetch.py
```diff
@@ -0,0 +1,273 @@
...
+
+    These commands help you download and prefetch artifacts like HuggingFace models
+    to your Flyte storage for faster access during task execution.
+    """
+
+
+@prefetch.command(name="hf-model", cls=CommandBase)
+@click.argument("repo", type=str)
+@click.option(
+    "--s3-path",
```
kumare3, line 31: can we call this objectstore-path?

kumare3 (approved):

PR #421, opened 2025-12-12
wandb plugin

samhita-alla:


src/flyte/link/link.py
```diff
@@ -0,0 +1,12 @@
+from typing import Dict, Protocol
+
+
+class Link(Protocol):
+    link_type: str
+    port: str
+
+    def get_config(self) -> Dict[str, str]:
```
kumare3: can we change this?

src/flyte/_internal/runtime/task_serde.py
```diff
@@ -131,6 +131,9 @@ def get_proto_task(task: TaskTemplate, serialize_context: SerializationContext)
         container = _get_urun_container(serialize_context, task)
         pod = None
 
+    if task.link:
+        extra_config.update(task.link.get_config())
```
kumare3: change it to pass the parameters

plugins/wandb/example.py
```diff
@@ -0,0 +1,264 @@
...
+    print(f"Traced Child task - Tags: {run.tags}")
+
+    # Log some metrics
+    run.log({"traced_child_metric": x * 3, "input": x})
+
+    return run.id
+
+
+@env.task
+@wandb_init(new_run=False)  # Use existing run from parent
```
kumare3: can we make it so that it will create a new run if not but use existing. I think it should be new_run="auto" so that it can decide correctly

kumare3: and this should be default

plugins/wandb/example.py
```diff
@@ -0,0 +1,264 @@
...
+@flyte.trace
+async def traced_no_wandb_init() -> str | int:
+    run = flyte.ctx().wandb_run
+    print(f"Traced no wandb init task - Run ID: {run}")  # Should be None
+
+    return run.id if run else -1
+
+
+@env.task
+@wandb_init
```
kumare3: does order matter, for async functions its weird

plugins/wandb/example.py
```diff
@@ -0,0 +1,264 @@
...
+    print(f"Grandchild task - Name: {run.name}")
+    print(f"Grandchild task - Tags: {run.tags}")
+
+    # Log some metrics
+    run.log({"child_metric": x * 4, "input": x})
+
+    return run.id
+
+
+@wandb_init
```
kumare3: order is different here

plugins/wandb/src/flyteplugins/wandb/link.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def get_link(
+        self,
+        run_name: str,
+        project: str,
+        domain: str,
+        context: Dict[str, str],
+        parent_action_name: str,
+        action_name: str,
+        pod_name: str,
```
kumare3, line 92: should we put **kwargs, in the example, i think we should do the same in the base, so that folks dont break us easily

plugins/wandb/src/flyteplugins/wandb/link.py
```diff
@@ -0,0 +1,63 @@
...
```
kumare3, line 26: sam here

plugins/wandb/src/flyteplugins/wandb/decorator.py
```diff
@@ -0,0 +1,352 @@
...
+            ctx.custom_context.pop("_wandb_init_action", None)
+
+        # Restore init kwargs
+        if saved_init_kwargs is not None:
+            ctx.data["_wandb_init_kwargs"] = saved_init_kwargs
+        else:
+            ctx.data.pop("_wandb_init_kwargs", None)
+
+
+def wandb_init(
```
kumare3, line 180: dont you think these decorator are complex and may cause bugs?

plugins/wandb/src/flyteplugins/wandb/link.py
samhita-alla: these are the values that custom link authors can use out of the box and we provide these values. i don't think we should support kwargs right?

samhita-alla: @pingsutw

plugins/wandb/src/flyteplugins/wandb/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    if not self.custom_context:
+        return None
+
+    # Return the sweep_id
+    return self.custom_context.get("_wandb_sweep_id")
+
+
+# Monkey-patch the properties onto TaskContext
+TaskContext.wandb_run = property(_wandb_run_property)
```
kumare3: Lets not do this please. Lets not monkey patch. Lets just add simple flyte.ctx().get_custom_context().get("..")
Or I like your helper method
get_wandb_config

plugins/wandb/src/flyteplugins/wandb/__init__.py
```diff
@@ -0,0 +1,60 @@
...
+from .context import (
+    get_wandb_context,
+    get_wandb_sweep_context,
+    wandb_config,
+    wandb_sweep_config,
+)
+from .decorator import wandb_init, wandb_sweep
+from .link import Wandb, WandbSweep
+
+logger = logging.getLogger(__name__)
```
kumare3: Why do you need this here?

plugins/wandb/src/flyteplugins/wandb/__init__.py
```diff
@@ -0,0 +1,60 @@
+import logging
```
kumare3: At the top of the module can we add a documentaion, that explains usages.
Examples

src/flyte/_trace.py
```diff
@@ -42,19 +43,35 @@ async def wrapper_async(*args: Any, **kwargs: Any) -> Any:
             else:
                 logger.debug(f"No existing trace info found for {func}, proceeding to execute.")
             start_time = time.time()
-            try:
+
+            if ctx.data.task_context is None:
```
kumare3: can you add unit test for this in tests/user_api

plugins/wandb/example.py
```diff
@@ -0,0 +1,279 @@
...
+    .with_apt_packages("git")
+    .with_pip_packages("git+https://github.com/flyteorg/flyte-sdk.git@144738932528f0fbaefcffc56e953824aca6d701"),
+    secrets=[flyte.Secret(key="wandb_api_key", as_env_var="WANDB_API_KEY")],
+)
+
+
+# Traces can access parent task's wandb run (no @wandb_init decorator allowed on traces)
+@flyte.trace
+async def traced_child_task(x: int) -> str:
+    run = flyte.ctx().wandb_run
```
kumare3: lets use get_wandb_config in the example

plugins/wandb/example.py
```diff
@@ -0,0 +1,279 @@
...
+    print(f"Traced Child task - Name: {run.name}")
+    print(f"Traced Child task - Tags: {run.tags}")
+
+    # Log some metrics
+    run.log({"traced_child_metric": x * 3, "input": x})
+
+    return run.id
+
+
+@wandb_init(new_run=False)  # same as new_run="auto"
```
kumare3: Lets write 3 different examples
1. Simple Link(...)
2. wandb_init
3. wandb_sweep

plugins/wandb/src/flyteplugins/wandb/decorator.py
```diff
@@ -0,0 +1,369 @@
...
+        extra_kwargs = config_dict.pop("kwargs", None) or {}
+        return {
+            **extra_kwargs,
+            **{k: v for k, v in config_dict.items() if v is not None},
+        }
+    return {}
+
+
+@contextmanager
+def _wandb_run(new_run: bool = "auto", func: bool = False, **decorator_kwargs):
```
kumare3:
```suggestion
def _wandb_run(new_run: Literal["auto", "always", "disabled"] = "auto", func: bool = False, **decorator_kwargs):
```

samhita-alla: replaced `new_run` with `run_mode` now. 
```python
run_mode: Literal["auto", "new", "shared"]
```

samhita-alla: @kumare3 let me know if this looks good to you.

plugins/wandb/src/flyteplugins/wandb/decorator.py
```diff
@@ -0,0 +1,370 @@
...
+        return {
+            **extra_kwargs,
+            **{k: v for k, v in config_dict.items() if v is not None},
+        }
+    return {}
+
+
+@contextmanager
+def _wandb_run(
+    run_mode: Literal["auto", "new", "shared"] = "auto",
```
kumare3: you should declare one Literal at the top of the file and then use it like a type. Like
`RunMode = Literal["...","..."]`

kumare3 (approved):

kumare3 (approved):

PR #422, opened 2025-12-12
Build: Docker layer granularity optimization plan

yuhuan130:
## Summary

Explore improving Docker build cache efficiency by separating layers based on how frequently they change.

## Problem

Some Docker layers currently mix dependencies with very different change frequencies.  
This causes unnecessary cache invalidation and slower rebuilds during development.

## Proposed direction

Order layers from most stable → most frequently changing:

1. System packages (apt)
2. Base Python dependencies
3. Project / feature dependencies
4. Local source code

This should allow common code-only changes to rebuild only the final layer.

## Example (illustrative)
```python
# Less cache-friendly
Image.from_debian_base().with_pip_packages(
    "pandas", "numpy", "scikit-learn", "matplotlib", "jupyter", "requests"
)
# More cache-friendly
(
    Image.from_debian_base()
        .with_pip_packages("pandas", "numpy", "requests")
        .with_pip_packages("scikit-learn")
        .with_pip_packages("matplotlib", "jupyter")
)

Before (Cache-Unfriendly):
┌────────────────────────────────────────┐
│ UVScript: Script + All Dependencies    │  ← Any change rebuilds everything
│ (pandas + matplotlib + jupyter + code) │
└────────────────────────────────────────┘

After (Cache-Friendly with Auto-Separation):
┌────────────────────────────────────────┐
│ UVScript: Script only (Optional)       │  ← Highest churn
├────────────────────────────────────────┤
│ Layer 4 (dev): jupyterlab, ruff, …     │  ← Freq: High | Cost: Low
├────────────────────────────────────────┤
│ Layer 3 (viz): matplotlib, plotly, …   │  ← Freq: Med  | Cost: Low
├────────────────────────────────────────┤
│ Layer 2 (utils): httpx, boto3, …       │  ← Freq: Med  | Cost: Low
├────────────────────────────────────────┤
│ Layer 1 (core_data): numpy, pandas…    │  ← Freq: Low  | Cost: Med
├────────────────────────────────────────┤
│ Layer 0 (heavy_ml): torch, tensorflow… │  ← Freq: Very Low | Cost: High
├────────────────────────────────────────┤
│ Base layers (Python + System)          │  ← Rarely change
└────────────────────────────────────────┘
```
## Docker Layering Tradeoff
We intentionally structured dependency layers by **build cost and change frequency**.

- Heavy frameworks (e.g. TensorFlow, PyTorch, Flyte) are isolated in low-churn layers to avoid expensive rebuilds.
- Faster-changing tooling and visualization libraries are kept higher to preserve iteration speed.
- This trades a slightly larger image context for **significantly faster rebuilds during ML development and training**.

Net effect: **lower developer wait time, predictable builds, and better scalability as pipelines grow**.

## Real-time examples
tensorflow + optimization
Screenshot 2025-12-14 at 16 28 28
tensorflow + mypy + optimization
Screenshot 2025-12-14 at 16 28 44
tensorflow
Screenshot 2025-12-14 at 16 28 57
tensorflow + mypy
Screenshot 2025-12-14 at 16 31 48

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -194,3 +194,5 @@ examples/remote_example.py
 config.yaml
 .claude/
 tests/flyte/internal/bin/flyte-inputs.json
+uv.lock
+uv.lock


src/flyte/_image.py
```diff
@@ -851,18 +844,53 @@ def my_task(x: int) -> int:
...
-        ll = PipPackages(
-            packages=new_packages,
-            index_url=index_url,
-            extra_index_urls=new_extra_index_urls,
-            pre=pre,
-            extra_args=extra_args,
-            secret_mounts=_ensure_tuple(secret_mounts) if secret_mounts else None,
-        )
-        new_image = self.clone(addl_layer=ll)
+        PACKAGE_IMPORTANCE = {
```
kumare3: Lets keep these in the file right at the top of the file as a constant not in the function please

kumare3: Also use immutable map

src/flyte/_image.py
kumare3, line 836: Can we add flag optimize_layers=True, which users can turn off if needed

yuhuan130: Absolutely

PR #425, opened 2025-12-14
Build: Docker layer granularity optimization plan - 1

yuhuan130:
## What This Does

Optimizes Docker build cache by separating heavy ML dependencies (torch, tensorflow) into their own layers and moving Flyte Python wheels to the bottom, so adding lightweight packages doesn't trigger full rebuilds.

### Heavy Benchmark (torch, tensorflow, transformers)
| Metric | Without Optimization | With Optimization | Result |
|--------|---------------------|-------------------|--------|
| Build Time | 310. 3s (5.2 min) | 8.7s (0.14 min) | **35.9× faster** |

### Standard Benchmark (torch, numpy, pandas)
| Metric | Without Optimization | With Optimization | Result |
|--------|---------------------|-------------------|--------|
| Build Time | 249.7s (4.2 min) | 7.7s | **32.3× faster** |

## How It Works

```
BEFORE:  Any change rebuilds everything
┌─────────────────────────────────────┐
│ ALL packages together               │
└─────────────────────────────────────┘

AFTER:  Optimized layer ordering  
┌─────────────────────────────────────┐
│ Heavy:   torch, tensorflow (cached) │ ← Rarely changes
├─────────────────────────────────────┤
│ Other                               │ ← Changes more often
├─────────────────────────────────────┤
│ Flyte PythonWheels (bottom layer)   │ ← Compiled once, fully cached
└─────────────────────────────────────┘
```

## Key Changes

- **`heavy_deps.py`**: New centralized configuration file for heavy dependencies (tensorflow, torch, scikit-learn, etc.)
- **`image_builder.py`**: Enhanced `_optimize_image_layers()` method with:
  - Heavy dependency extraction to top layer
  - Flyte wheel layer placement at bottom for maximum caching
  - UV script dependency parsing support
- **New flag**: `optimize_layers=True` (enabled by default, can disable)
- **3 benchmark examples** showing 32-36× speedup
- **Smarter layer ordering**:  Heavy deps cached separately, Flyte wheels at bottom

## Files Changed

- `src/flyte/_internal/imagebuild/heavy_deps.py` *(new)*
- `src/flyte/_internal/imagebuild/image_builder.py`
- `src/flyte/_image. py`
- `tests/flyte/test_image. py`
- `examples/image_layer_optimize/` (3 benchmark files)

**Total:** 7 files (1 new, 6 modified)

## Screenshots
Screenshot 2025-12-30 at 17 57 18
Screenshot 2025-12-30 at 18 13 42
Screenshot 2026-01-06 at 16 42 45
Screenshot 2026-01-09 at 00 15 09

*Replaces #422 with cleaner git history*


src/flyte/_image.py
```diff
@@ -849,20 +877,51 @@ def my_task(x: int) -> int:
...
+            category = "unknown"
+            if optimize_layers:
+                for cat, pkg_list in PACKAGE_IMPORTANCE.items():
+                    if pkg_name in pkg_list:
+                        category = cat
+                        break
+            categorized[category].append(pkg)
+
+        # Helper function to create a layer
+        def create_pip_layer(pkgs):
```
pingsutw:
```python
Image.from_debian_base(name="test").with_pip_packages("tensorflow").with_pip_packages("pytorch")
```

In this case, TensorFlow and Pytorch will be installed in different layers, right? Let's install them in the same layer.

Also, I think it might be better to add `_optimize` method in ImageBuildEngine, and we can do

```python
image = ImageBuildEngine._optimize(image)
result = await img_builder.build_image(image, dry_run=dry_run)
```
https://github.com/flyteorg/flyte-sdk/blob/55d1b63cb60f86fe507b33bed2b709b5dd94db5a/src/flyte/_internal/imagebuild/image_builder.py#L209-L210

It will be cleaner, wdyt?

src/flyte/_image.py
```diff
@@ -812,6 +839,7 @@ def with_pip_packages(
         pre: bool = False,
         extra_args: Optional[str] = None,
         secret_mounts: Optional[SecretRequest] = None,
+        optimize_layers: bool = True,
```
pingsutw: Could we move it to `clone()`

```python
Image.from_debian_base().clone(optimize_layers=True)
```

src/flyte/_image.py
```diff
@@ -29,6 +30,16 @@
 
 T = TypeVar("T")
 
+PACKAGE_IMPORTANCE = MappingProxyType(
+    {
+        # Layer 0: ~1GB+ | Rebuild cost: High | Freq: Very Low
+        "heavy": ("tensorflow", "torch", "torchaudio", "torchvision", "scikit-learn"),
+        # -----------------[ MIDDLE ]----------------- #
+        # Layer 1: ~200MB | Rebuild cost: Med  | Freq: Low
+        "core": ("numpy", "pandas", "pydantic", "requests", "httpx", "boto3", "fastapi", "uvicorn"),
```
pingsutw: I just tried to build these two images; the build times are almost the same. Do we really need this core layer?

```python
image = (
    Image.from_debian_base(install_flyte=False)
    .with_apt_packages("vim", "wget")
    .with_pip_packages("pandas", "numpy")
)
```

```python
image = (
    Image.from_debian_base(install_flyte=False)
    .with_apt_packages("vim", "wget")
    .with_pip_packages("pandas", "numpy", "ty")
)
```

examples/benchmark.py
```diff
@@ -0,0 +1,75 @@
...
+import time
+
+import flyte
+from flyte import Image
+
+env = flyte.TaskEnvironment(name="benchmark")
+
+
+@env.task
+async def main():
```
pingsutw: 👍

src/flyte/_image.py
yuhuan130: I think the similar build times make sense here, without the additional optimize_layers flag marked false. Both images reuse the cached apt + pandas/numpy layers, so the only real installing is ty, which is small. Technically it'll save around 45 seconds - 1 minute.

yuhuan130: Hmm interesting, I tried running it locally once with and without optimize, and I found out python:3.12-slim-bookworm base actually pre-installs the wheels for pandas and numpy etc... so it'll be fast for these two packages either way.

src/flyte/_image.py
yuhuan130: Great idea. I've updated your suggestion on my latest commit!!

src/flyte/_internal/imagebuild/image_builder.py
```diff
@@ -135,6 +146,100 @@ class ImageBuildEngine:
...
+        logger.info(
+            f"Optimizing {len(pip_layers)} pip layer(s): {len(heavy_packages)} heavy, \
+                  {len(other_packages)} other packages"
+        )
+
+        # Build optimized layer list
+        optimized_layers: list[Layer] = []
+
+        # Use settings from first pip layer as template for consolidated layers
+        template_layer = pip_layers[0]
```
pingsutw: Should we use the settings from the layer where we install the heavy packages instead?

src/flyte/_image.py
```diff
@@ -626,6 +626,11 @@ def from_uv_script(
         Args:
             secret_mounts:
         """
+        from ._utils import parse_uv_script_file
+
+        metadata = parse_uv_script_file(Path(script))
```
pingsutw: Could we move it to _optimize_image_layers?

yuhuan130: I think this is a good idea.

src/flyte/_internal/imagebuild/image_builder.py
yuhuan130: ah~~, i see. Yea that'll be better indeed

src/flyte/_internal/imagebuild/image_builder.py
```diff
@@ -135,6 +138,149 @@ class ImageBuildEngine:
...
+        Optimize pip layers by extracting heavy dependencies to the top for better caching.
+        Original layers remain in their original positions but with heavy packages removed.
+        PythonWheels layers with package_name 'flyte' are moved to the very end.
+        """
+        from flyte._utils import parse_uv_script_file
+
+        # Step 1: Collect heavy packages and build new layer list
+        all_heavy_packages: list[str] = []
+        template_layer: PipPackages | None = None
+        optimized_layers: list[Layer] = []
```
pingsutw: nit: could we call it `original_layers` or `other_layers`?

src/flyte/_internal/imagebuild/image_builder.py
```diff
@@ -135,6 +138,149 @@ class ImageBuildEngine:
...
+                            light_pkgs.append(pkg)
+
+                    # Collect heavy packages
+                    if uv_heavy_pkgs:
+                        all_heavy_packages.extend(uv_heavy_pkgs)
+                        if template_layer is None:
+                            # Create template from UV layer config
+                            template_layer = PipPackages(
+                                packages=(),
+                                index_url=layer.index_url,
```
pingsutw: I think we should have `heavy_layers: List`. each layers has different settings

src/flyte/_internal/imagebuild/image_builder.py
```diff
@@ -135,6 +138,149 @@ class ImageBuildEngine:
 
     ImageBuilderType = typing.Literal["local", "remote"]
 
+    @staticmethod
+    def _optimize_image_layers(image: Image) -> Image:
```
pingsutw, line 154: Could we also add some unit tests for it, thanks

PR #431, opened 2025-12-16
Improve Rust controller

machichima:
- Add `make dev-rs-dist` to run `make build-wheels`, `make dist`, build image, and install `flyte_controller_base` locally
- Stop `_submit_loop` in rust controller correctly when running sync task
- Remove color code in remote logging (see following image)

image

### DIFF
--- a/rs_controller/src/lib.rs
+++ b/rs_controller/src/lib.rs
@@ -152,8 +152,14 @@ impl BaseController {
 fn flyte_controller_base(_py: Python<'_>, m: &Bound<'_, PyModule>) -> PyResult<()> {
     static INIT: std::sync::Once = std::sync::Once::new();
     INIT.call_once(|| {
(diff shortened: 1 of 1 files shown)


rs_controller/Makefile
```diff
@@ -48,7 +48,7 @@ build-wheels-amd64: dist-dirs
 clean_dist:
 	rm -rf $(DIST_DIRS)/*whl
 
-build-wheels: clean_dist build-wheels-arm64 build-wheels-amd64
```
wild-endeavor, line 51: removing the clean intentionally?

machichima: Oh yes sorry, I was trying to make build faster, I think remove clean can give us incremental build rather than starting clean

machichima: Oh yes sorry, I was trying to make build faster, I think remove clean can give us incremental build rather than starting clean

PR #435, opened 2025-12-16
First attempt to have a micro-batching example

davidmirror-ops:


examples/higher_order_patterns/micro_batching.py
```diff
@@ -0,0 +1,243 @@
...
+    results = await microbatch_workflow()
+    print(f"Workflow completed. Total results: {len(results)}")
+
+
+
+# Alternative: Two-level batching for extreme scale
+# Use this if you need even more control or have >1M items
+
+@orchestrator_env.task
+async def two_level_microbatch_workflow(
```
kumare3: lets drop this one. I want to abolish this pattern. you should not need 2 levels, you can always stride in one level

examples/higher_order_patterns/micro_batching.py
```diff
@@ -0,0 +1,243 @@
...
+            print(f"Exception for item {batch_start + i}: {result}")
+            processed_results.append(-1)
+        else:
+            processed_results.append(result)
+    
+    return processed_results
+
+
+@orchestrator_env.task
+async def microbatch_workflow(
```
kumare3: this is good, i have some examples like https://github.com/flyteorg/flyte-sdk/blob/main/examples/data_processing/pyiceberg_example.py

PR #439, opened 2025-12-18
Fix PKCE to read the entire callback request properly

EngHabu:

### DIFF
--- a/src/flyte/remote/_client/auth/_authenticators/pkce.py
+++ b/src/flyte/remote/_client/auth/_authenticators/pkce.py
@@ -414,7 +414,7 @@ class OAuthCallbackHandler:
         # Using readline() instead of read() because read() waits for EOF, which won't come
         # until the client closes the connection - but the client is waiting for our response first.
         request_line = await reader.readline()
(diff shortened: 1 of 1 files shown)


src/flyte/remote/_client/auth/_authenticators/pkce.py
```diff
@@ -410,10 +410,11 @@ async def handle(self, reader, writer):
...
         """
-        data = await reader.read()
-        message = data.decode()
-        headers = message.split("\r\n")
-        path = headers[0].split(" ")[1]
+        # Read only the first line of the HTTP request (e.g., "GET /callback?code=...&state=... HTTP/1.1")
+        # Using readline() instead of read() because read() waits for EOF, which won't come
+        # until the client closes the connection - but the client is waiting for our response first.
+        request_line = await reader.readline()
+        path = request_line.decode().split(" ")[1]
```
kumare3, line 418: you can just do split() and it should split on all white spaces. Also do you want [1] or [0]. If you want [1] this probably means you want `lstrip()` and `lstrip` or better just run `strip()` and that will remove all whitespaces

EngHabu: I added a comment to show an example.. the line has 3 parts 'GET' then '/callback?.....' then 'HTTP.....' we want [1].. `strip` only looks at the beginning and end of the line to remove whitespaces.

PR #449, opened 2025-12-19
Implement app_env.server, on_startup, on_shutdown decorators and pickled apps

cosmicBboy:
This PR introduces an `AppEnvironment.startup` decorator. This is a Python-native way to define app startup logic and supports pickling app environments so that users can do `flyte.with_servecontext(interactive_mode=True).serve(app_env)`.

When a user defines `app_env.startup`, it will take the place of `args`, which typically creates a subprocess in the app container which runs that command. Instead, the `AppEnvironment` is loaded into the app container and the `AppEnvironment._startup_fn` is run.

```python
import fastapi
import uvicorn

import flyte
import flyte.app

env = flyte.app.AppEnvironment(
    name="pickled-app",
    image=flyte.Image.from_debian_base().with_pip_packages("fastapi", "uvicorn"),
    resources=flyte.Resources(cpu=1, memory="512Mi"),
    requires_auth=False,
    port=8080,
)


@env.startup
def fastapi_app():
    # NOTE: since FastAPI cannot be pickled (because starlette.datastructures.State cannot be pickled due to
    # circular references), we need to use a factory pattern to create the app instance. In the startup function.
    app = fastapi.FastAPI()

    @app.get("/")
    async def root() -> str:
        return "Hello, World!"

    uvicorn.run(app, port=8080)


if __name__ == "__main__":
    import logging

    flyte.init_from_config(log_level=logging.DEBUG)
    app = flyte.with_servecontext(interactive_mode=True).serve(env)
    print(app.url)

```

This PR also adds:
- Support for module- and pickle-based app startup using the private `AppEnvironment._startup_fn` decorator.
- Adds a `AppEnvResolver` class to rehydrate app environments in the `fserve` executable.
- Adds support for pickling the `FastAPIAppEnvironment` with no user-facing SDK changes.

### DIFF
--- a/examples/apps/pickled_app.py
+++ b/examples/apps/pickled_app.py
@@ -11,7 +11,6 @@ env = flyte.app.AppEnvironment(
     parameters=[flyte.app.Parameter(name="foo", value="bar")],
     requires_auth=False,
     port=8080,
-    env_vars={"LOG_LEVEL": "10"},
 )
 
 state = {}
@@ -41,7 +40,7 @@ async def app_server(foo: str):
 @env.on_shutdown
 async def app_shutdown(**kwargs):
     # clear the state on shutdown
-    print("Shutting down app...")
+    print("Shutting down app, clearing state...")
     state.clear()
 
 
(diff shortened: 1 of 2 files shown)


examples/apps/pickled_app.py
```diff
@@ -0,0 +1,55 @@
...
+    # NOTE: since FastAPI cannot be pickled (because starlette.datastructures.State cannot be pickled due to
+    # circular references), we need to use a factory pattern to create the app instance. In the startup function.
+    app = fastapi.FastAPI()
+
+    @app.get("/")
+    async def root() -> str:
+        return f"Hello, World! {foo}, here is the state: {state}"
+
+    print("Starting up uvicorn server...")
+    await uvicorn.Server(uvicorn.Config(app, port=8080)).serve()
```
kumare3, line 37: does the server decorator block

cosmicbboy: yep, I believe it does block

PR #450, opened 2025-12-19
Fix refresh_token path for PKCE

EngHabu:

### DIFF
--- a/src/flyte/remote/_client/auth/_keyring.py
+++ b/src/flyte/remote/_client/auth/_keyring.py
@@ -98,12 +98,10 @@ class KeyringStore:
                  or if the system keyring is not available
         """
         for_endpoint = strip_scheme(for_endpoint)
-        access_token = ""
+        access_token: str | None = None
         try:
             refresh_token = keyring.get_password(for_endpoint, KeyringStore._refresh_token_key)
-            access_token_opt = keyring.get_password(for_endpoint, KeyringStore._access_token_key)
-            if access_token_opt:
-                access_token = access_token_opt
+            access_token = keyring.get_password(for_endpoint, KeyringStore._access_token_key)
         except NoKeyringError as e:
             logger.debug(f"KeyRing not available, tokens will not be cached. Error: {e}")
             return None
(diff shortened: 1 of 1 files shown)


src/flyte/remote/_client/auth/_keyring.py
```diff
@@ -98,6 +98,7 @@ def retrieve(for_endpoint: str) -> typing.Optional[Credentials]:
                  or if the system keyring is not available
         """
         for_endpoint = strip_scheme(for_endpoint)
+        access_token = ""
```
wild-endeavor: if we're going to initialize access token should we also initialize refresh token?

src/flyte/remote/_client/auth/_keyring.py
```diff
@@ -108,10 +109,13 @@ def retrieve(for_endpoint: str) -> typing.Optional[Credentials]:
             logger.debug(f"Failed to retrieve tokens from keyring. Error: {e}")
             return None
 
-        if not access_token:
+        if not access_token and not refresh_token:
             logger.debug("No access token found in keyring.")
             return None
 
+        if not access_token and refresh_token:
+            access_token = ""
```
wild-endeavor: confused by this... maybe it's the jet lag, wouldn't access_token already be 'not'?

src/flyte/remote/_client/auth/_keyring.py
EngHabu: refresh_token is optional

src/flyte/remote/_client/auth/_keyring.py
EngHabu: Cleaned up a bit...

src/flyte/remote/_client/auth/_keyring.py
```diff
@@ -109,8 +110,11 @@ def retrieve(for_endpoint: str) -> typing.Optional[Credentials]:
             return None
 
         if not access_token:
-            logger.debug("No access token found in keyring.")
-            return None
+            if not refresh_token:
+                logger.debug("No access token found in keyring.")
+                return None
+            else:
+                access_token = ""
```
wild-endeavor, line 117: oh Credentials has this as a str, not optional str, got it.

PR #467, opened 2025-12-23
Silence grpcio polling when using non-Flyte event loops

cosmicBboy:
This PR suppresses grpcio polling-related warnings from being emitted in the syncify API. This future-proofs suppression of these kinds of warnings when flyte grpcio calls are made when using non-Flyte-controlled event loops.

### DIFF
--- a/examples/apps/run_webhook.py
+++ b/examples/apps/run_webhook.py
@@ -1,17 +1,20 @@
 # /// script
 # requires-python = ">=3.12"
 # dependencies = [
+#     "ipdb",
 #     "fastapi",
 #     "uvicorn",
-#     "flyte==2.0.0b40",
+#     "flyte==2.0.0b42",
 # ]
 # ///
+
 import logging
 import os
 import pathlib
 from contextlib import asynccontextmanager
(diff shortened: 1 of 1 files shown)


src/flyte/syncify/_api.py
```diff
@@ -150,6 +150,17 @@ async def iterate_in_loop(self, async_gen: AsyncIterator[R_co]) -> AsyncIterator
                 yield r
             return
 
+        # Install the gRPC error handler on the caller's event loop as well.
+        # This is needed because gRPC's async polling events may be delivered to
+        # the caller's loop (e.g., FastAPI's event loop) when using .aio().
+        import flyte.errors
+
+        try:
+            caller_loop = asyncio.get_running_loop()
```
kumare3: I think we should not do it again and again, we should only do it once really, i would put a variable block

cosmicbboy: what do you mean by variable block?

kumare3: like have a thread local variable that can be used to check if things have been set

PR #468, opened 2025-12-23
rename mention of "Reference Task" -> "Remote Task"

cosmicBboy:


src/flyte/errors.py
```diff
@@ -149,13 +149,25 @@ class ActionNotFoundError(RuntimeError):
...
 
 
+# NOTE: Use RemoteTaskError instead, since "reference tasks" (from v1) are now
+# simply "remote tasks" in v2.
 class ReferenceTaskError(RuntimeUserError):
     """
     This error is raised when the user tries to access a task that does not exist.
     """
 
+    CODE = "ReferenceTaskUsageError"
```
kumare3, line 159: Want to change this too

cosmicbboy: it's fine, the code will be `RemoteTaskUsageError` when `RemoteTaskError` is raised

PR #483, opened 2025-12-29
Document OCR example

kumare3:
Maximizing GPU usage with Flyte. Process thousands of documents through a model like QwenVL using flyte 2.0 and reusable containers

### DIFF
--- a/examples/ml/ocr/batch_ocr.py
+++ b/examples/ml/ocr/batch_ocr.py
@@ -24,8 +24,7 @@ MODEL_ID = "Qwen/Qwen2.5-VL-3B-Instruct"
 
 # Base image with OCR dependencies
 ocr_image = flyte.Image.from_debian_base().with_uv_project(
-    pyproject_file=Path("pyproject.toml"),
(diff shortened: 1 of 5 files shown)


examples/ml/ocr/batch_ocr.py
```diff
@@ -0,0 +1,669 @@
...
+    config = MODEL_GPU_CONFIGS[model]
+
+    logger.info(f"Using dispatcher: {dispatcher.short_name} for {model.value}")
+    logger.info(f"GPU config: {config.gpu_type}:{config.gpu_count}, Memory: {config.memory}")
+
+    # Process chunks in parallel using the selected dispatcher
+    tasks = []
+    for idx, chunk in enumerate(chunks):
+        with flyte.group(f"chunk-{idx}"):
+            task = asyncio.create_task(dispatcher(model, chunk, batch_size))
```
samhita-alla: i don't think we need to run `create_task` before running `gather`. we just need this when using `as_completed`, right?

samhita-alla (approved):

PR #494, opened 2026-01-03
common.initialize_config used instead of local initialize_config.Remo…

debasisdwivedy:
The run method uses initialize_config from the common package. Remove local package for clarity as it is unused

### DIFF
--- a/src/flyte/cli/_run.py
+++ b/src/flyte/cli/_run.py
@@ -18,7 +18,7 @@ from . import _common as common
 from ._params import to_click_option
 
 RUN_REMOTE_CMD = "deployed-task"
-
+initialize_config = common.initialize_config
 
 @lru_cache()
 def _list_tasks(


src/flyte/cli/_run.py
```diff
@@ -14,25 +14,12 @@
 from .._code_bundle._utils import CopyFiles
 from .._task import TaskTemplate
 from ..remote import Run
-from . import _common as common
```
kumare3, line 17: Can you please use common as the alias. This makes it easy to understand where certain things are coming from here

src/flyte/cli/_run.py
```diff
@@ -14,25 +14,12 @@
...
-from ._common import CLIConfig, initialize_config
+from ._common import CLIConfig, initialize_config, PROJECT_OPTION, DOMAIN_OPTION, get_option_from_metadata,get_console,get_panel, ObjectsPerFileGroup,GroupBase,FileGroup
 from ._params import to_click_option
 
 RUN_REMOTE_CMD = "deployed-task"
 
 
-@lru_cache()
-def _initialize_config(ctx: click.Context, project: str, domain: str, root_dir: str | None = None):
```
kumare3, line 25: should this be just deleted

src/flyte/cli/_run.py
debasisdwivedy: Imported the entire package as common for namespace clarity. so instead of having the code as below:

```
from . import _common as common
from ._common import CLIConfig, initialize_config
```
Now you have :

```
from . import _common as common
```
All the functions and variable present in commons need to be accessed via common.*

src/flyte/cli/_run.py
debasisdwivedy: `_initialize_config` is never used. Instead we use `common.initialize_config`.

kumare3 (approved):

PR #500, opened 2026-01-05
Update vllm and sglang plugin dependencies

cosmicBboy:
This PR removes the vllm and sglang package dependencies from the flyteplugins-vllm and flyteplugins-sglang package dependencies. This is because:

- To use those plugins, the user doesn't actually need to have vllm or sglang installed in the local environment. Those dependencies are only required in the images that run the apps remotely.
- This simplifies and future proofs testing and maintenance. E.g. the vllm and flyte have a transitive dependency conflict (vllm uses a very old version of `protovalidate`). Therefore it's hard to install flyte and vllm in the same environment without invoking separate installation commands.


kumare3 (approved):

cosmicBboy:
one sec @kumare3, still WIP

PR #501, opened 2026-01-05
make sure app versions are consistent with same include files

cosmicBboy:
This PR makes sure that the app code bundle versions are consistent when the include files are specified and do not change.

### DIFF
--- a/src/flyte/app/_deploy.py
+++ b/src/flyte/app/_deploy.py
@@ -78,7 +78,8 @@ async def _deploy_app(
     from flyte.app._runtime import translate_app_env_to_idl
     from flyte.remote import App
 
-    if app.include and serialization_context.code_bundle.pkl is None:
+    is_pkl = serialization_context.code_bundle and serialization_context.code_bundle.pkl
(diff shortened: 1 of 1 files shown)


src/flyte/app/_deploy.py
```diff
@@ -78,16 +78,17 @@ async def _deploy_app(
     from flyte.app._runtime import translate_app_env_to_idl
     from flyte.remote import App
 
-    if app.include and serialization_context.code_bundle and serialization_context.code_bundle.pkl is None:
-        # Only bundle the include files if the code bundle is a tgz bundle. If this
-        # is a pkl bundle, assume that the AppEnvironment has a server function
-        # that will be used to serve the app. This function should contain all
-        # of the code needed to serve the app.
+    is_pkl = serialization_context.code_bundle and serialization_context.code_bundle.pkl
+    if app.include and is_pkl:
```
kumare3, line 82:
```suggestion
    if app.include and not is_pkl:
```

?

kumare3 (approved):

PR #503, opened 2026-01-05
compileall in the local builder

pingsutw:


wild-endeavor (dismissed):

src/flyte/_internal/runtime/io.py
```diff
@@ -85,8 +85,8 @@ async def upload_error(err: execution_pb2.ExecutionError, output_prefix: str) ->
             message=err.message,
             kind=errors_pb2.ContainerError.RECOVERABLE,
             origin=err.kind,
-            timestamp=err.timestamp,
-            worker=err.worker,
+            # timestamp=err.timestamp,
```
kumare3: can we delete these itself?

pingsutw: deleted

PR #505, opened 2026-01-06
Pass through initialization and auth passing

kumare3:
Usage!

```python
import flyte
import flyte.remote as remote

flyte.init_passthrough("endpoint")

with remote.auth_metadata(("key1", "value1")):
    print(remote.Run.listall(limit=2))
```


src/flyte/remote/_client/auth/_authenticators/passthrough.py
```diff
@@ -0,0 +1,79 @@
...
+        # Lazy import to avoid circular dependencies
+        from flyte.remote._auth_metadata import get_auth_metadata
+
+        # Get metadata from context
+        metadata_tuples = get_auth_metadata()
+
+        if not metadata_tuples:
+            return None
+
+        # Print the metadata for debugging (as requested)
```
jeevb: We can clean these up.

wild-endeavor (approved):

PR #508, opened 2026-01-06
combining pkce enum types

debasisdwivedy:
Converting the code below:

```
if auth_type is None:
        return "pkce"
if auth_type.lower() in _pkce_options:
        return "Pkce"
```

to

```
if auth_type is None or auth_type.lower() in _pkce_options:
        return "Pkce"
```

### DIFF
--- a/src/flyte/cli/_common.py
+++ b/src/flyte/cli/_common.py
@@ -81,9 +81,7 @@ def sanitize_auth_type(auth_type: str | None) -> str:
     """
     Convert the auth type to the mode that is used by the Flyte backend.
     """
-    if auth_type is None:
-        return "pkce"
-    if auth_type.lower() in _pkce_options:
+    if auth_type is None or auth_type.lower() in _pkce_options:
         return "Pkce"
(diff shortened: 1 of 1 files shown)


src/flyte/cli/_common.py
```diff
@@ -16,6 +16,7 @@
 import rich.box
 import rich.repr
 import rich_click as click
+import click.exceptions
```
kumare3: wy?

kumare3: also this changes the import back to `click` we want to continue using rich_click

debasisdwivedy: Reverted it back. But i think would be better as it would only import the exceptions from the click package. `rich-click` does not have `exception` module exposed, it internally calls `click.exception`. It does not show during static code analysis and might create problems at runtime.

This would even better as we are explicitly calling the exception package:

```
from click.exceptions import Exit, NoArgsIsHelpError
```

We can add it later if we find any issue

PR #513, opened 2026-01-07
adding config file exception

debasisdwivedy:
Adding exception is config file is not provided by the user and if not present in the default path.
Currently it is not clear if the config file is missing as the error generated is for missing arguments like below:

`Error invoking command: Error invoking command: Missing required parameter(s): --name (type: TEXT) `

The config file might not be present in one of the default location as below:

```
a. ./config.yaml if it exists  (current working directory)   
b. ./.flyte/config.yaml if it exists (current working directory)       
c. <git_root>/.flyte/config.yaml if it exists         
d. `UCTL_CONFIG` environment variable                   
e. `FLYTECTL_CONFIG` environment variable         
f. ~/.union/config.yaml if it exists            
g. ~/.flyte/config.yaml if it exists   
```

This might not be the case. In such a scenario the user has to be made aware that the config file needs to be added in one of the default location or provide a path for the config file.
Regards

### DIFF
--- a/src/flyte/cli/main.py
+++ b/src/flyte/cli/main.py
@@ -214,8 +214,6 @@ def main(
     cfg = config.auto(config_file=config_file)
     if cfg.source:
         logger.debug(f"Using config file discovered at location `{cfg.source.absolute()}`")
-    else:
-        raise click.ClickException(f"Config file not found.Please provide the path to the config file or create one in the default path ~/.flyte/config.yaml")
 
     ctx.obj = CLIConfig(
         log_level=log_level,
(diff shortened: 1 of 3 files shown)


src/flyte/cli/main.py
```diff
@@ -214,6 +214,8 @@ def main(
     cfg = config.auto(config_file=config_file)
     if cfg.source:
         logger.debug(f"Using config file discovered at location `{cfg.source.absolute()}`")
+    else:
```
kumare3:
```suggestion
    elif not cfg:
```
?

debasisdwivedy: Added it to _intialize module and _run module. Just want to give the user clarity about the issue in case config is missing. Adding it to main.py will fail all the test cases as the assertions will fail

PR #515, opened 2026-01-07
[wip] Fix CI for publishing plugin images

cosmicBboy:


kumare3 (dismissed):

plugins/sglang/src/flyteplugins/sglang/_app_environment.py
```diff
@@ -14,7 +14,7 @@
 from flyteplugins.sglang._constants import SGLANG_MIN_VERSION_STR
 
 DEFAULT_SGLANG_IMAGE = (
-    flyte.Image.from_debian_base(name="sglang-app-image")
+    flyte.Image.from_debian_base(name="sglang-app-image", python_version=(3, 13))
```
wild-endeavor, line 17: does this mean we'll only build one version of the sglang-app-image and it'll only be 3.13?  that's fine but will users accidentally somehow use another one?  what happens if they create an sglang app while on 3.11?

PR #526, opened 2026-01-09
[Feat] use flyteidl2 dataproxy

machichima:
use dataproxy from flyteidl2 rather than flyteidl


wild-endeavor:
We'll need to update the backend before this can be used right?

PR #527, opened 2026-01-09
Implement polars plugin, improve overall dataframe developer experience

cosmicBboy:
This PR:
- Implements polars plugin: parquet <> dataframe encoders and decoders for `polars.DataFrame` and `polars.LazyFrame`. This supports proper `sink_parquet` and `scan_parquet` with the LazyFrame encoder/decoder
- Implements better error reporting when downloading dataframes from s3 into local python runtime

### DIFF
--- a/src/flyte/io/_dataframe/dataframe.py
+++ b/src/flyte/io/_dataframe/dataframe.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import _datetime
 import asyncio
 import collections
-import os
 import pathlib
 import types
 import typing
@@ -18,10 +17,10 @@ from pydantic import BaseModel, ConfigDict, Field, PrivateAttr, model_serializer
 from typing_extensions import Annotated, TypeAlias, get_args, get_origin
 
 import flyte.storage as storage
-from flyte.storage._storage import get_credentials_error
 from flyte._logging import logger
 from flyte._utils import lazy_module
 from flyte._utils.asyn import loop_manager
+from flyte.storage._storage import get_credentials_error
 from flyte.types import TypeEngine, TypeTransformer, TypeTransformerFailedError
 from flyte.types._renderer import Renderable
 from flyte.types._type_engine import modify_literal_uris
--- a/src/flyte/storage/_storage.py
+++ b/src/flyte/storage/_storage.py
@@ -5,7 +5,7 @@ import pathlib
 import random
 import tempfile
 import typing
-from typing import AsyncGenerator, Optional, Coroutine, Any
+from typing import AsyncGenerator, Optional
 from uuid import UUID
 
 import fsspec
@@ -480,6 +480,7 @@ def get_credentials_error(uri: str, protocol: str) -> str:
             "Please set the following environment variables: AZURE_STORAGE_ACCOUNT_NAME, "
             "AZURE_STORAGE_ACCOUNT_KEY"
         )
+    raise ValueError(f"Unsupported protocol: {protocol}")
 
 
 register(_OBSTORE_SUPPORTED_PROTOCOLS, asynchronous=True)


plugins/polars/src/flyteplugins/polars/__init__.py
```diff
@@ -0,0 +1,15 @@
+__all__ = [
+    "ParquetToPolarsDecodingHandler",
+    "ParquetToPolarsLazyFrameDecodingHandler",
+    "PolarsLazyFrameToParquetEncodingHandler",
+    "PolarsToParquetEncodingHandler",
+    "register_polars_df_transformers",
```
kumare3: why do we need to expose this?

plugins/polars/src/flyteplugins/polars/__init__.py
```diff
@@ -0,0 +1,15 @@
+__all__ = [
+    "ParquetToPolarsDecodingHandler",
```
kumare3: also all of this. for the user they should see almost nothing right?

plugins/polars/src/flyteplugins/polars/df_transformer.py
```diff
@@ -0,0 +1,241 @@
+import functools
+import os
+import typing
+from pathlib import Path
+
+import flyte.storage as storage
+from flyte._logging import logger
+from flyte._utils import lazy_module
+from flyte.io._dataframe.dataframe import (
```
kumare3: this is available from `flyte.io.extend`

src/flyte/cli/_run.py
```diff
@@ -254,7 +254,7 @@ async def _execute_and_render(self, ctx: click.Context, config: common.CLIConfig
                 result = await execution_context.run.aio(self.obj, **ctx.params)
         except Exception as e:
             console.print(common.get_panel("Exception", f"[red]✕ Execution failed:[/red] {e}", config.output_format))
-            return
+            raise
```
wild-endeavor, line 257: what's the effect of this?  why is it needed?  are you trying to get the exit code to be non-zero?

src/flyte/io/_dataframe/dataframe.py
```diff
@@ -969,13 +983,17 @@ async def to_literal(
             )
 
         # Otherwise assume it's a dataframe instance. Wrap it with some defaults
+        if not ctx.has_raw_data:
+            fdf = await DataFrame.from_local(python_val)
+            return await self.to_literal(fdf, python_type, expected)
```
wild-endeavor, line 988: should fdf._literal_sd be available at this point?  just return that?

src/flyte/io/_dataframe/dataframe.py
```diff
@@ -957,6 +967,10 @@ async def to_literal(
 
             # 3. This is the third and probably most common case. The python DataFrame object wraps a dataframe
             # that we will need to invoke an encoder for. Figure out which encoder to call and invoke it.
+            if not ctx.has_raw_data:
+                fdf = await DataFrame.from_local(python_val.val)
```
wild-endeavor, line 971: ditto

src/flyte/io/_dataframe/dataframe.py
```diff
@@ -1146,7 +1164,12 @@ async def open_as(
         """
         protocol = get_protocol(sd.uri)
         decoder = self.get_decoder(df_type, protocol, sd.metadata.structured_dataset_type.format)
-        result = await decoder.decode(sd, updated_metadata)
+
+        try:
+            result = await decoder.decode(sd, updated_metadata)
+        except OSError as exc:
+            raise OSError(f"{exc}\n{get_credentials_error(sd.uri, protocol)}") from exc
```
pingsutw, line 1171: Just curious, won’t Polars raise an error saying something like "credentials are required"

src/flyte/cli/_run.py
cosmicbboy: > are you trying to get the exit code to be non-zero?

correct, it was weird to me that an exception was raised but didn't exit non-zero

src/flyte/io/_dataframe/dataframe.py
cosmicbboy: nope, this is when the user is passing in a local raw dataframe. It needs to use `from_local` to upload the dataframe and get the remote path. e.g. when user does `DataFrame.wrap_df`, it doesn't necessarily have `_literal_sd` available yet

src/flyte/io/_dataframe/dataframe.py
cosmicbboy: > nope, this is when the user is passing in a local raw dataframe. It needs to use `from_local` to upload the dataframe and get the remote path.

src/flyte/io/_dataframe/dataframe.py
cosmicbboy: correct, but I want to surface more informative error message in general

src/flyte/io/_dataframe/dataframe.py
```diff
@@ -396,6 +399,9 @@ async def all(self) -> DF:  # type: ignore
 
         return await flyte_dataset_transformer.open_as(self.literal, self._dataframe_type, self.metadata)
 
+    def all_sync(self) -> DF:  # type: ignore
+        return asyncio.run(self.all())
```
kumare3, line 403: is this right, or should we make the all method syncified?

PR #536, opened 2026-01-13
adding checks for mandatory options org,project and domain

debasisdwivedy:
Checks in SDK are not performed for `org`,`project` and `domain` values for None or "" . So when a task is run the backend gives an error as below:

"Failed to create TaskAction CR: failed to create TaskAction CR: TaskAction.flyte.org \"my-org---run-1768321423-run-1768321423\" is invalid: [spec.project: Invalid value: \"\": spec.project in body should be at least 1 chars long,

Config.yaml used:

```
task:
  #domain: development
  #project: flytesnacks
  org: my-org
```
Do not pass the argument in your flyte cli. Try: `flyte run hello.py main --name World`

You need to check that the user has provided `org`,`project` and `domain` either in config.yaml or passed the value as an option

Regards

### DIFF
--- a/src/flyte/cli/_run.py
+++ b/src/flyte/cli/_run.py
@@ -202,14 +202,17 @@ class RunTaskCommand(click.RichCommand):
                         (param_name, param.type.get_metavar(param, ctx) or param.type.name.upper() or param.type)
                     )
 
-        if ctx.obj.org is None and ctx.obj.config.task.org is None:
-            missing_options.append(("org", "string"))
+        task_cfg = getattr(getattr(ctx.obj, "config", None), "task", None)
 
-        if self.run_args.project is None and ctx.obj.config.task.project is None:
-            missing_options.append(("project", "string"))
+        if not self.run_args.local:
+            if not getattr(ctx.obj, "org", None) and not getattr(task_cfg, "org", None):
+                missing_options.append(("org", "TEXT"))
 
-        if self.run_args.domain is None and ctx.obj.config.task.domain is None:
-            missing_options.append(("domain", "string"))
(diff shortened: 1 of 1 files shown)


kumare3 (approved):

src/flyte/cli/_run.py
```diff
@@ -346,14 +372,38 @@ def _validate_required_params(self, ctx: click.Context) -> None:
...
+        task_cfg = getattr(getattr(ctx.obj, "config", None), "task", None)
+
+        if not getattr(ctx.obj, "org", None) and not getattr(task_cfg, "org", None):
+            missing_options.append(("org", "TEXT"))
+
+        if not self.run_args.run_project and not getattr(task_cfg, "project", None):
+            missing_options.append(("run-project", "TEXT"))
+
+        if not self.run_args.run_domain and not getattr(task_cfg, "domain", None):
+            missing_options.append(("run-domain", "TEXT"))
```
pingsutw, line 393: Should we add a helper function for this? This code is similar to the code above

debasisdwivedy: We can definitely make it more generic, where we have a list with all the mandatory options and a function that check if the mandatory options are present and accordingly update the `missing_options` list

I think `org`,`project` and `domain` is required for all the task below.

-  **Run and stop tasks**   
    run  
    abort  
-  **Management of various objects**  
    create  
    get  
    delete  
    update  
-  **Prefetch artifacts from remote registries**  
    prefetch  
     

Currently it is only added for run task. Could you please verify if that is the case.

In such a scenario the options checking has to be done outside of the `cli` package as it will need it for both `cli` and `sdk`

Regards

PR #538, opened 2026-01-16
add snowflake connector

samhita-alla:


plugins/connectors/src/flyteplugins/connectors/snowflake/connector.py
```diff
@@ -0,0 +1,277 @@
...
+        user = custom.get("user")
+        database = custom.get("database")
+        schema = custom.get("schema", "PUBLIC")
+        warehouse = custom.get("warehouse")
+
+        if not all([user, database, warehouse]):
+            raise ValueError("User, database and warehouse must be specified in the task configuration.")
+
+        # Extract authentication secrets from kwargs
+        private_key_content = kwargs.get("snowflake_private_key")
```
pingsutw: Could we just add snowflake_private_key to the args in `create`, like

```python
async def create(
        self,
        task_template: TaskTemplate,
        inputs: Optional[Dict[str, Any]] = None,
        snowflake_private_key,
        **kwargs,
    ) -> SnowflakeJobMetadata:
```

PR #548, opened 2026-01-20
add support for private urls in AppEndpoint

cosmicBboy:
This PR: addresses an issue with `flyte.app.AppEndpoint` where injecting private urls from upstream apps was not supported. It does so by retrieving the URL at serving time.

### DIFF
--- a/src/flyte/app/_parameter.py
+++ b/src/flyte/app/_parameter.py
@@ -5,7 +5,7 @@ import re
 import typing
 from dataclasses import dataclass, field
 from functools import cache, cached_property
-from typing import TYPE_CHECKING, List, Literal, Optional, TypeAlias
+from typing import TYPE_CHECKING, List, Literal, Optional, TypeAlias, Union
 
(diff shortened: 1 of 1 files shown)


examples/apps/app_calling_app/app.py
```diff
@@ -25,16 +27,24 @@
...
     app=app2,
     image=image,
     resources=flyte.Resources(cpu=1, memory="512Mi"),
-    requires_auth=False,
+    requires_auth=True,
+    parameters=[
+        flyte.app.Parameter(
+            name="app1_url",
+            value=flyte.app.AppEndpoint(app_name="app1-is-called-by-app2"),
+            env_var="APP1_URL",
```
kumare3, line 43: Why not private endpoints that are auto created?

cosmicbboy: Private endpoints are not exposed in the app idl, and are not accessible at deployment time.

kumare3 (approved):

PR #560, opened 2026-01-23
Support uv editable installs; fix cwd editable detection and secret mounts

redartera:
## Summary
- include uv editable path dependencies when building images in dependencies-only mode for local and remote builders
- add utils to parse uv export editable paths and build bind mounts
- fix local builder secret mount assembly to avoid concatenating mounts without separators
- fix cwd editable install selection to prefer the editable parent of the current directory before falling back to a sibling/parent heuristic

## Failure Case (cwd editable detection)
In the manual repro below, the editable dependency (`editable-lib/`) lives under the main project (`project-root/`). When running from the project root, the old logic could incorrectly pick `editable-lib/` as the cwd editable install just because its parent matched the cwd, even though the actual project root should have been selected.

Project layout:

```
project-root/
├─ pyproject.toml
├─ uv.lock
├─ main.py
└─ editable-lib/
   ├─ pyproject.toml
   └─ src/
      └─ editable_lib/
         └─ __init__.py
```

Explanation:
- `cwd` is `project-root/`.
- `editable-lib/` is a direct child of `project-root/`.
- The old loop could return `editable-lib/` due to the “parent == cwd” check, even though `project-root/` is the correct editable root.
- The new logic first checks for editable installs that are parents of (or equal to) `cwd`, then only falls back to the one-level-above heuristic.

### Automated
- not run (tests added/updated: `tests/flyte/imagebuild/test_utils.py`, `tests/flyte/imagebuild/test_docker_builder.py`)

### Manual
Project layout used for end-to-end validation (editable dep created with `uv init --lib`):

```
project-root/
├─ pyproject.toml
├─ uv.lock
├─ main.py
└─ editable-lib/
   ├─ pyproject.toml
   └─ src/
      └─ editable_lib/
         └─ __init__.py
```

`pyproject.toml` used:

```toml
[project]
name = "test-editables"
version = "0.1.0"
description = "Add your description here"
readme = "README.md"
authors = [
    { name = "redartera", email = "reda@artera.ai" }
]
requires-python = ">=3.13"
dependencies = ["editable-lib", "flyte"]

[build-system]
requires = ["uv_build>=0.8.2,<0.9.0"]
build-backend = "uv_build"

[tool.uv.sources]
editable-lib = { path = "./editable-lib", editable = true }
flyte = { git = "https://github.com/redartera/flyte-sdk", branch = "feature/image-build-editable-installs" }
```

Repro script (`main.py`, run from the same project root; `git` installed in the image so `uv sync` works with the git-based dependency):

```python
import flyte
from pathlib import Path


env = flyte.TaskEnvironment(
    "hello_world",
    image=flyte.Image.from_debian_base(
        registry="XXXXXXXXXXX",
        name="XXXXXXX",
        platform=("linux/amd64",),
    )
    .with_apt_packages("git")  # Important to install git so uv sync with the git-based dependency works
    .with_uv_project(
        pyproject_file=Path(__file__).parent / "pyproject.toml",
        uvlock=Path(__file__).parent / "uv.lock",
    )
)

@env.task
def mean(data: list[float]) -> float:
    return sum(data) / len(data)

@env.task
def main(data: list[float]) -> float:
    output = mean(data)
    output = output / 100
    if output < 0:
        raise ValueError("Output cannot be negative")
    return output
```

### DIFF
--- a/src/flyte/_internal/imagebuild/remote_builder.py
+++ b/src/flyte/_internal/imagebuild/remote_builder.py
@@ -304,8 +304,11 @@ def _get_layers_proto(image: Image, context_path: Path) -> "image_definition_pb2
             )
             layers.append(pip_layer)
         elif isinstance(layer, UVProject):
-            # Copy pyproject itself
             pyproject_dst = copy_files_to_context(layer.pyproject, context_path)
+            # Keep track of the directory containing the pyproject.toml file
+            # this is what should be passed to the UVProject image definition proto as 'pyproject'
+            pyproject_dir_dst = pyproject_dst.parent
+
             # Copy uv.lock itself
             uvlock_dst = copy_files_to_context(layer.uvlock, context_path)
 
@@ -327,13 +330,15 @@ def _get_layers_proto(image: Image, context_path: Path) -> "image_definition_pb2
                 case "install_project":
                     # Copy the entire project
                     docker_ignore_patterns = get_and_list_dockerignore(image)
-                    pyproject_dst = copy_files_to_context(layer.pyproject.parent, context_path, docker_ignore_patterns)
+                    pyproject_dir_dst = copy_files_to_context(layer.pyproject.parent, context_path, docker_ignore_patterns)
                 case _:
                     raise ValueError(f"Invalid project install mode: {layer.project_install_mode}")
 
             uv_layer = image_definition_pb2.Layer(
                 uv_project=image_definition_pb2.UVProject(
-                    pyproject=str(pyproject_dst.relative_to(context_path)),
+                    # NOTE: UVProject expects 'pyproject' to be the directory containing the pyproject.toml file
+                    # whereas it expects 'uvlock' to be the path to the uv.lock file itself.
+                    pyproject=str(pyproject_dir_dst.relative_to(context_path)),
                     uvlock=str(uvlock_dst.relative_to(context_path)),
                     options=pip_options,
                     secret_mounts=secret_mounts,


src/flyte/_utils/helpers.py
```diff
@@ -106,16 +106,20 @@ def get_cwd_editable_install() -> typing.Optional[Path]:
...
+    # First, prioritize editable installs that are parents of cwd
     for install in editable_installs:
         # child.is_relative_to(parent) is True if child is inside parent
         if cwd.is_relative_to(install):
             return install
-        else:
-            # check if the cwd is one level above the install folder
-            if install.parent == cwd:
-                # check if the install folder contains a pyproject.toml file
-                if (cwd / "pyproject.toml").exists() or (cwd / "setup.py").exists():
```
redartera, line 117: should I just perhaps add a condition checking that `(cwd / "src")` exists instead of decoupling the two loops?

pingsutw: Why is this change needed?

redartera: running `flyte run ...` would somewhat think the editable dependency is the root of the project and bundle code from there - that's because in my setup the editable dependency is just one level below the project's `pyproject.toml`.

The correct behaviour should have been to detect `project-root` as the root of the code bundle instead of `editable-lib`.

This was my layout:
```
project-root/
├─ pyproject.toml
├─ uv.lock
├─ main.py
└─ editable-lib/
   ├─ pyproject.toml
   └─ src/
      └─ editable_lib/
         └─ __init__.py
```

pingsutw: I see, but it could always explicitly set the root dir, right? I am thinking of removing get_cwd_editable_install in the future.

pingsutw: this change is fine for now

redartera: > I see, but it could always explicitly set the root dir, right? I am thinking of removing get_cwd_editable_install in the future.

Great point thanks for sharing, I wasn't aware `flyte run` had a `--root-dir` flag until now, super useful

redartera: @pingsutw this now works with the remote builder (both uv project install modes) - the change I made is https://github.com/flyteorg/flyte-sdk/pull/560/commits/ea9c4ee4466e24cfa26df19131db8907fc780da6

dansola: @pingsutw this will not work if the editable package is one level above the root directory, correct?

dansola: I wonder if we can bypass the is_relative_to guard.

PR #562, opened 2026-01-23
Add logs for networking endpoints.

mhotan:
Helps validate proxy configuration

### DIFF
--- a/src/flyte/_internal/controllers/remote/_core.py
+++ b/src/flyte/_internal/controllers/remote/_core.py
@@ -191,8 +191,12 @@ class Controller:
         self._running = True
         logger.debug("Waiting for Service Client to be ready")
         client_set = await self._client_coro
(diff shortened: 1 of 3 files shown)


src/flyte/_internal/controllers/remote/_core.py
```diff
@@ -191,8 +191,12 @@ async def _bg_worker_pool(self):
         self._running = True
         logger.debug("Waiting for Service Client to be ready")
         client_set = await self._client_coro
+        self._client_set = client_set
         self._state_service: StateService = client_set.state_service
         self._queue_service: QueueService = client_set.queue_service
+        # Log the endpoint for network debugging
+        endpoint = getattr(client_set, 'endpoint', None)
```
kumare3, line 198: we could expose this on the protocol (python protocol) instead of the getattr

PR #565, opened 2026-01-24
Use separate thread for running sync tasks

wild-endeavor:
Sync tasks broken in the following scenario:
* `_run_and_stop` in runtime runs the task, and blocks on `self.func()` in `task.execute`
* This [raise](https://github.com/flyteorg/flyte-sdk/blob/4a5f4e1bffa13ffcee00a09057018367b0d66187/src/flyte/_internal/controllers/remote/_core.py#L131) never happens because this is in the main thread and the main thread is blocked on `self.func()`.

The way around this is to create a new thread to run `self.func`

### DIFF
--- a/src/flyte/_task.py
+++ b/src/flyte/_task.py
@@ -490,7 +490,6 @@ class AsyncFunctionTaskTemplate(TaskTemplate[P, R, F]):
         assert ctx.data.task_context is not None, "Function should have already returned if not in a task context"
         ctx_data = await self.pre(*args, **kwargs)
         tctx = ctx.data.task_context.replace(data=ctx_data)
-        # claude: starting here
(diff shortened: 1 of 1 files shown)


src/flyte/_task.py
```diff
@@ -493,7 +495,40 @@ async def execute(self, *args: P.args, **kwargs: P.kwargs) -> R:
             if iscoroutinefunction(self.func):
                 v = await self.func(*args, **kwargs)
             else:
-                v = self.func(*args, **kwargs)
+                copied_ctx = contextvars.copy_context()
```
kumare3: lets move it to 
```python
await utils.asyncify(func, *args, **kwargs)
```

PR #567, opened 2026-01-25
Add wait in image builder protocol

cosmicBboy:
This adds an option to `flyte.build` to not wait for the image build to complete. Note: when the builder is `remote` and `wait=False`, the `build` method will return the url to the image build task instead of the fully qualified URI

### DIFF
--- a/src/flyte/_internal/imagebuild/docker_builder.py
+++ b/src/flyte/_internal/imagebuild/docker_builder.py
@@ -1,4 +1,3 @@
-import asyncio
 import os
 import shutil
 import subprocess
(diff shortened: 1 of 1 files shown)


src/flyte/_build.py
```diff
@@ -6,10 +6,25 @@
...
     Build an image. The existing async context will be used.
 
+    Args:
+        image: The image(s) to build.
+        dry_run: Tell the builder to not actually build. Different builders will have different behaviors.
+        force: Skip the existence check. Normally if the image already exists we won't build it.
+        wait: Wait for the build to finish. If wait is False, the function will return immediately and the build will
+            run in the background.
+    Returns:
+        The image URI. If wait is False when using the remote image builder, the function will return the build image
```
kumare3: I agree, making this an explicit object will be better. 
Like ImageBuild object that contains, build job url and uri this can be useful in other places too

pingsutw: +1 to the build object

kumare3 (approved):

PR #578, opened 2026-01-27
[Rust Controller] Handle SlowDownError (retry with backoff)

machichima:
Handle `SlowDownError` like what we had in python controller


src/flyte/_internal/controllers/remote/_core.py
```diff
@@ -362,7 +362,7 @@ async def _bg_launch(self, action: Action):
                             trace=trace,
                             input_uri=action.inputs_uri,
                             run_output_base=action.run_output_base,
-                            group=action.group if action.group else None,
+                            group=action.group.name if action.group else "",
```
machichima: Not related to the `SlowDownError`, but this will cause runtime error when executing tasks

PR #579, opened 2026-01-27
Update polars examples

cosmicBboy:


src/flyte/_internal/imagebuild/docker_builder.py
```diff
@@ -715,7 +718,10 @@ async def _build_image(self, image: Image, *, push: bool = True, dry_run: bool =
                 click.secho(f"Run command: {concat_command} ", fg="blue")
 
             try:
-                await asyncio.to_thread(subprocess.run, command, check=True)
+                if wait:
+                    await asyncio.to_thread(subprocess.run, command, check=True)
+                else:
+                    await asyncio.to_thread(subprocess.Popen, command)
```
kumare3: this will block the main thread from exiting

PR #584, opened 2026-01-28
snowflake connector: add support for DF conversion, batch inputs, and update docstrings

samhita-alla:
This PR

* adds support for retrieving pandas dataframes from the snowflake query output - expects the secrets (private key and private key passphrase) to be injected as env vars - `SNOWFLAKE_PRIVATE_KEY` and `SNOWFLAKE_PRIVATE_KEY_PASSPHRASE`
* updates docstrings for the API reference docs
* batches inputs if `batch=True` in the `Snowflake` task - this is useful for batching input queries automatically
* adds `secret_group` to the `Snowflake` task to avoid having to specify the secret as an env var. the plugin takes care of generating the env var given the group (optional) and key
* updates the snowflake example

TODO: test the example with the latest updates.

### DIFF
--- a/plugins/snowflake/README.md
+++ b/plugins/snowflake/README.md
@@ -6,9 +6,6 @@ Run Snowflake SQL queries as Flyte tasks with parameterized inputs, key-pair aut
 
 ```bash
 pip install flyteplugins-snowflake
-
-# With DataFrame support (read query results as pandas DataFrames):
-pip install flyteplugins-snowflake[dataframe]
 ```
 
 ## Quick start
@@ -97,7 +94,7 @@ insert_rows = Snowflake(
 
 ## Reading results as DataFrames
 
-Install with `pip install flyteplugins-snowflake[dataframe]`, then set `output_dataframe_type`:
+Set `output_dataframe_type` to get query results as a pandas DataFrame:
 
 ```python
 import pandas as pd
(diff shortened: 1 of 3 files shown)


plugins/snowflake/src/flyteplugins/snowflake/task.py
```diff
@@ -36,29 +43,40 @@ def __init__(
         query_template: str,
         plugin_config: SnowflakeConfig,
         inputs: Optional[Dict[str, Type]] = None,
-        output_dataframe_type: Optional[Type[DataFrame]] = None,
+        output_dataframe_type: Optional[Type] = None,
+        secret_group: Optional[str] = None,
```
pingsutw, line 47: Why do we need the secret_group? if the secret is created by `flyte create secret`, the group is always None, right

samhita-alla: i'm not sure in which instances we need a secret group but wanted to support just in case. i think some secret managers expect a group, right?

src/flyte/io/_dataframe/__init__.py
```diff
@@ -114,11 +99,6 @@ def lazy_import_dataframe_handler():
             register_bigquery_handlers()
         except DuplicateHandlerError:
             logger.debug("Transformer for bigquery is already registered.")
-    if is_imported("snowflake.connector"):
```
kumare3, line 117: thank you

PR #585, opened 2026-01-28
[Fix] Make batch size configurable in Dir `from_local`

machichima:
We discovered a large memory consumption when uploading large directory (e.g. `50GB`), the memory usage will reach `11GB` and throw timeout error. It's because in `fsspec`, the default batch size is set to the system file descriptor limit ([here](https://github.com/fsspec/filesystem_spec/blob/2576617e5cbe441bcc53b021bccd85ff3489fde7/fsspec/asyn.py#L185))

This PR did set the default to 32 same as the max concurrency set in `src/flyte/storage/_parallel_reader.py`

https://github.com/flyteorg/flyte-sdk/blob/6572a5d0b6e24badec6f7f5894158fd1ac6f99fc/src/flyte/storage/_parallel_reader.py#L22

### DIFF
--- a/src/flyte/io/_dir.py
+++ b/src/flyte/io/_dir.py
@@ -623,7 +623,7 @@ class Dir(BaseModel, Generic[T], SerializableType):
                           as an input to discoverable tasks. If not specified, the cache key will be based on
                           directory attributes.
             batch_size: Optional concurrency limit for uploading files. If not specified, the default value is
-              determined by the FLYTE_IO_MAX_CONCURRENCY environment variable (default: 32).
+              determined by the FLYTE_IO_BATCH_SIZE environment variable (default: 32).
 
         Returns:
             A new Dir instance pointing to the uploaded directory
--- a/src/flyte/storage/_storage.py
+++ b/src/flyte/storage/_storage.py
@@ -26,7 +26,7 @@ if typing.TYPE_CHECKING:
     from obstore import AsyncReadableFile, AsyncWritableFile
 
 _OBSTORE_SUPPORTED_PROTOCOLS = ["s3", "gs", "abfs", "abfss"]
-MAX_CONCURRENCY = int(os.getenv("FLYTE_IO_MAX_CONCURRENCY", str(32)))
+BATCH_SIZE = int(os.getenv("FLYTE_IO_BATCH_SIZE", str(32)))
 
 
 def _is_obstore_supported_protocol(protocol: str) -> bool:
@@ -279,7 +279,7 @@ async def put(
         to_path = ctx.raw_data.get_random_remote_path(file_name=name)
 
     if not batch_size:
-        batch_size = MAX_CONCURRENCY
+        batch_size = BATCH_SIZE
 
     file_system = get_underlying_filesystem(path=to_path)
     from_path = strip_file_header(from_path)


src/flyte/storage/_storage.py
```diff
@@ -26,6 +26,7 @@
     from obstore import AsyncReadableFile, AsyncWritableFile
 
 _OBSTORE_SUPPORTED_PROTOCOLS = ["s3", "gs", "abfs", "abfss"]
+MAX_CONCURRENCY = int(os.getenv("FLYTE_IO_MAX_CONCURRENCY", str(32)))
```
kumare3: Let's make smaller env var like _F_IO_MAX_*

kumare3 (dismissed):

src/flyte/storage/_storage.py
cosmicbboy: why not more explicit?

src/flyte/storage/_storage.py
```diff
@@ -263,18 +264,27 @@ async def _get_from_filesystem(
     return str(to_path)
 
 
-async def put(from_path: str, to_path: Optional[str] = None, recursive: bool = False, **kwargs) -> str:
+async def put(
+    from_path: str,
+    to_path: Optional[str] = None,
+    recursive: bool = False,
+    batch_size: Optional[int] = None,
```
pingsutw, line 271:
```suggestion
    batch_size: int = MAX_CONCURRENCY,
```

src/flyte/io/_dir.py
```diff
@@ -621,6 +622,8 @@ async def upload_with_cache_key() -> Dir:
             dir_cache_key: Optional precomputed hash value to use for cache key computation when this Dir is used
                           as an input to discoverable tasks. If not specified, the cache key will be based on
                           directory attributes.
+            batch_size: Optional concurrency limit for uploading files. If not specified, the default value is
+              determined by the FLYTE_IO_MAX_CONCURRENCY environment variable (default: 32).
```
pingsutw:
```suggestion
              determined by the FLYTE_IO_BATCH_SIZE environment variable (default: 32).
```

src/flyte/storage/_storage.py
pingsutw: the default value should be FLYTE_IO_BATCH_SIZE

src/flyte/storage/_storage.py
machichima: I use `FLYTE_IO_MAX_CONCURRENCY` to match the usage in: 
https://github.com/flyteorg/flyte-sdk/blob/6572a5d0b6e24badec6f7f5894158fd1ac6f99fc/src/flyte/storage/_parallel_reader.py#L22

In this case when user set `FLYTE_IO_MAX_CONCURRENCY`, this will affect on both `ObstoreParallelReader` and `put` here

src/flyte/storage/_storage.py
machichima: I set the default value below:

```py
    if not batch_size:
        batch_size = MAX_CONCURRENCY
```

As in `src/flyte/io/_dir.py`, we will pass in `batch_size` like so:

```py
storage.put(
            from_path=local_path_str, to_path=remote_destination, recursive=True, batch_size=batch_size
        )
```

When `batch_size` is `None` here, after passing in, the default value we set through `batch_size: int = MAX_CONCURRENCY` will not be applied.

machichima: > the default value should be FLYTE_IO_BATCH_SIZE

You mean we should change `FLYTE_IO_MAX_CONCURRENCY` env var above to `FLYTE_IO_BATCH_SIZE`?

The reason why I'm using this is here: https://github.com/flyteorg/flyte-sdk/pull/585/files#r2739036154

src/flyte/io/_dir.py
machichima: The reason for using `FLYTE_IO_MAX_CONCURRENCY` is here:

https://github.com/flyteorg/flyte-sdk/pull/585/files#r2739036154

WDYT?

machichima: After chatting offline, we will change to `FLYTE_IO_BATCH_SIZE`, as batch size and max concurrency is different concept. Now, the `FLYTE_IO_BATCH_SIZE` will be applied to `put`, and `FLYTE_IO_MAX_CONCURRENCY` be applied to `ObstoreParallelReader` only

machichima: updated in https://github.com/flyteorg/flyte-sdk/pull/585/commits/12ad2eb8e36c5dd093643cce8c9fac9cbc39843f

src/flyte/storage/_storage.py
machichima: ref: https://github.com/flyteorg/flyte-sdk/pull/585#discussion_r2739115736

modified in https://github.com/flyteorg/flyte-sdk/commit/12ad2eb8e36c5dd093643cce8c9fac9cbc39843f

src/flyte/storage/_storage.py
machichima: ref: https://github.com/flyteorg/flyte-sdk/pull/585#discussion_r2739115736

modified in https://github.com/flyteorg/flyte-sdk/commit/12ad2eb8e36c5dd093643cce8c9fac9cbc39843f

src/flyte/storage/_storage.py
pingsutw, line 271:
```suggestion
    recursive: bool = False,
    batch_size: int = BATCH_SIZE,
```

src/flyte/storage/_storage.py
```diff
@@ -263,18 +264,27 @@ async def _get_from_filesystem(
...
+) -> str:
     if not to_path:
         from flyte._context import internal_ctx
 
         ctx = internal_ctx()
         name = pathlib.Path(from_path).name
         to_path = ctx.raw_data.get_random_remote_path(file_name=name)
 
+    if not batch_size:
+        batch_size = BATCH_SIZE
```
pingsutw, line 282: and we can remove it

src/flyte/storage/_storage.py
machichima: See response here: https://github.com/flyteorg/flyte-sdk/pull/585#discussion_r2739042032

PR #591, opened 2026-01-29
add n8n example

cosmicBboy:
Production-grade example of serving a standalone n8n app on Union apps.

- Uses postgres (via supabase) for data persistence
- Uses podtemplate sidecar to implement javascript/python runners
- Uses flyte webhook app to trigger flyte tasks from n8n workflows

### DIFF
--- a/examples/genai/n8n/README.md
+++ b/examples/genai/n8n/README.md
@@ -25,3 +25,8 @@ Then deploy the app:
 ```bash
 python n8n_app.py
 ```
+
+This will deploy two apps and one task environment:
+1. `n8n-app` - The main n8n app. This contains a sidecar container that runs the n8n javascript and python task runners.
+2. `flyte-n8n-webhook-app` - A webhook app that allows you to trigger Flyte tasks from n8n workflows.
+3. `flyte-n8n-webhook-task` - A task environment that contains some toy Flyte task that can be triggered via the webhook.


examples/genai/n8n_app.py
```diff
@@ -0,0 +1,36 @@
+import flyte
+import flyte.app
+
+
+# TODO:
+# - Add postgres database for user data persistence
```
kumare3: just use supabase?

cosmicbboy: Yep, that's the plan!

examples/genai/n8n/flyte_webhook_app.py
```diff
@@ -0,0 +1,231 @@
...
+image = flyte.Image.from_debian_base().with_pip_packages("fastapi", "uvicorn")
+
+
+# ---------------------------------------------------------------------------
+# FastAPI lifespan: initialize Flyte passthrough auth on startup
+# ---------------------------------------------------------------------------
+@asynccontextmanager
+async def lifespan(app: FastAPI):
+    PROJECT_NAME_ENV_VAR = "FLYTE_INTERNAL_EXECUTION_PROJECT"
+    DOMAIN_NAME_ENV_VAR = "FLYTE_INTERNAL_EXECUTION_DOMAIN"
```
kumare3, line 43: we have flyte.current_domain() we should use that i dont want to use these env vars. Can i also add current_project and you can use it?

PR #594, opened 2026-01-30
bugfix: fix df local to remote

cosmicBboy:
This PR:
- fixes an issue where a `flyte.io.DataFrame` created from a local run could not be passed as input to a remote run.
- adds a `preserve_original_types` to `with_runcontext`. If set to `True`, then flyte will do a best-effort conversion of offloaded types (in particular raw dataframe types like `pd.DataFrame`) so that `run.outputs()` returns the original type.

### DIFF
--- a/examples/basics/dataframes/dataframe_inputs.py
+++ b/examples/basics/dataframes/dataframe_inputs.py
@@ -105,7 +105,7 @@ if __name__ == "__main__":
     local_task_df = local_run.outputs()[0]
 
     for dataframe in [in_mem_dataframe, local_task_df]:
-        run = flyte.run(process_df, df=dataframe)
+        run = flyte.with_runcontext(preserve_original_types=True).run(process_df, df=dataframe)
         print(run.url)
         run.wait()
         result = run.outputs()[0]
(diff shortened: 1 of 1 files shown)


examples/basics/dataframes/dataframe_inputs.py
```diff
@@ -81,36 +93,43 @@ def process_fdf_to_fdf(df: flyte.io.DataFrame) -> flyte.io.DataFrame:
...
+        run.wait()
+        result = run.outputs()[0]
+        assert isinstance(result, pd.DataFrame)
+        print(result)
+
+        flyte_dataframe = flyte.io.DataFrame.from_df(dataframe)
+        run = flyte.run(process_fdf_to_df, df=flyte_dataframe)
+        print(run.url)
+        run.wait()
+        result: pd.DataFrame = run.outputs()[0]
```
kumare3, line 119: Why are we doing this 😩 this will break everything. Anytime someone is working in an app etc they will materialize the results. How did this happen? How do you know the type? What if we don't have the dependency? What if we use remote tasks. What if I want to get the data frame only ?

cosmicbboy: > How did this happen?

The changes snuck in here, I forgot to revert changes I originally made here: https://github.com/flyteorg/flyte-sdk/pull/527

I reverted it

PR #597, opened 2026-01-31
Add first-class support for tuple and NamedTuple via pydantic

cosmicBboy:
This PR introduces a TupleTransformer and NamedTupleTransformer, which delegates serdes responsibility to the pydantic base model:

- tuple and namedtuple types are converted to pydantic BaseModels under the hood
- at deserialization time, the transformers take the json schema metadata to reconstitute the native python types
- complex nesting of subtypes (including File, Dir, dataclasses, BaseModel, etc)

## Why is this needed?

Many AI frameworks like LangGraph, LangChain, PydanticAI etc heavily rely on types like tuple, NamedTuple, and TypedDict in their type system. Rather than pickling, this PR preserves observability by serializing these types in a way that shows up in the UI (see https://github.com/flyteorg/flyte-sdk/pull/598 for TypedDict PR)


src/flyte/types/_type_engine.py
```diff
@@ -819,6 +819,465 @@ def assert_type(self, t: Type[enum.Enum], v: T):
             raise TypeTransformerFailedError(f"Value {v} is not in Enum {t}")
 
 
+class TupleTransformer(TypeTransformer[tuple]):
```
kumare3, line 906: shall we move it to its own file? I think we should decompose type engine. Its becoming too big to handle

src/flyte/types/_type_engine.py
```diff
@@ -819,6 +819,465 @@ def assert_type(self, t: Type[enum.Enum], v: T):
...
+    """
+    Transformer that handles typed tuples like tuple[int, str, float].
+
+    This transformer delegates to PydanticTransformer by wrapping the tuple
+    in a dynamically generated Pydantic BaseModel.
+    """
+
+    def __init__(self):
+        super().__init__("Typed Tuple", tuple, enable_type_assertions=False)
```
kumare3, line 915: is this ok to have 2 transformers for the same type?

src/flyte/types/_type_engine.py
kumare3, line 906: i cannot tell, but this one and the named tuple one are they repetitive, can we just inherit from some common base and have common code

kumare3 (approved):

PR #599, opened 2026-02-02
 Bring determinism to your programs

kumare3:
- Durable Sleep: Sleep once and maintain consistency in the sleep
- Durable time, simple wrapper to get the same time again.

### DIFF
--- a/tests/flyte/extras/test_frozen_time.py
+++ b/tests/flyte/extras/test_frozen_time.py
@@ -4,8 +4,9 @@ from unittest.mock import AsyncMock, patch
 
 import pytest
 
-from flyte.durable import sleep as durable_sleep, time as durable_time, now as durable_now
-
+from flyte.durable import now as durable_now
+from flyte.durable import sleep as durable_sleep
+from flyte.durable import time as durable_time
 
 # -- durable_time tests --
 
--- a/examples/basics/determinism.py
+++ b/examples/basics/determinism.py
@@ -1,19 +1,35 @@
+import asyncio
+
 import flyte
 import flyte.durable
 
 env = flyte.TaskEnvironment("determinism")
 
+
 @env.task
-async def main(n: int) -> int:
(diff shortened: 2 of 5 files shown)


pingsutw (approved):

tests/flyte/extras/test_frozen_time.py
```diff
@@ -0,0 +1,144 @@
...
+    with (
+        patch("flyte.durable._time.time.time", side_effect=[100.0, 103.0]),
+        patch("flyte.durable._time.asyncio.sleep", new_callable=AsyncMock) as mock_sleep,
+    ):
+        await durable_sleep.aio(5.0)
+        mock_sleep.assert_awaited_once_with(2.0)
+
+
+@pytest.mark.asyncio
+async def test_durable_sleep_zero_seconds():
```
machichima, line 106: Should we also add test to cover negative value for `durable_sleep`?

tests/flyte/extras/test_frozen_time.py
```diff
@@ -0,0 +1,144 @@
...
+    assert isinstance(result, float)
+
+
+def test_durable_time_sync_returns_float():
+    result = durable_time()
+    assert isinstance(result, float)
+
+
+@pytest.mark.asyncio
+async def test_durable_time_returns_current_time():
```
machichima, line 26: I think for `durable_time`, we want it to give consistent value for all replays? Is it possible for us to add a test for replay scenario?

Same comment for `durable_now`

tests/flyte/extras/test_frozen_time.py
kumare3: thats ok

tests/flyte/extras/test_frozen_time.py
kumare3: hmm let me think, i will merge and then comeback

PR #607, opened 2026-02-05
Fixes a potential subtle race condition in call-seq-generation

kumare3:


src/flyte/_internal/controllers/remote/_controller.py
```diff
@@ -143,17 +143,19 @@ def generate_task_call_sequence(self, task_obj: object, action_id: ActionID) ->
...
-        current_action_sequencer[current_task_id] = new_seq
         name = ""
         if hasattr(task_obj, "__name__"):
             name = task_obj.__name__
         elif hasattr(task_obj, "name"):
             name = task_obj.name
+
+        uniq = unique_action_name(action_id)
+        current_action_sequencer = self._parent_action_task_call_sequence[uniq]
+        current_task_id = name or id(task_obj)
```
wild-endeavor, line 154: i think i'm mostly talking to myself.
can you remind me why we did this?  why not just have a global counter per `uniq`?  every time we make any call, it's incremented.

IIRC, the reason is to try to prevent new action ids in recovery right?  for example, let's say we have this (obviously contrived situation)

```python
@env.task
async def t1(): ...

async def run_t1():
  await sleep(some random time)
  return await t1()

@env.task
async def t2(): ...

async def run_t2():
  await sleep(some random time)
  return await t2()

@env.task
async def t3(): ...

async def run_t3():
  await sleep(some random time)
  return await t3()

@env.task
def parent_task():
  at1 = asyncio.create_task(run_t1())
  at2 = asyncio.create_task(run_t2())
  at3 = asyncio.create_task(run_t3())

  asyncio.gather(at1, at2, at3)
```

By keeping a counter for each task, even if a recovery run runs them out of order, they'll get the same action IDs.

With this change, we're merging calls right?  so taking the example from today.
```python
@env.task
async def t1(): ...

async def run_t1(override_name: str):
  await sleep(some random time)
  return await t1().override(short_name=override_name)

@env.task
def parent_task():
  at1 = asyncio.create_task(run_t1(override_name="name1"))
  at2 = asyncio.create_task(run_t1(override_name="name2"))
  at3 = asyncio.create_task(run_t1(override_name="name3"))

  asyncio.gather(at1, at2, at3)
```
now that we're merging the calls, in a recovery scenario, now the results might be jumbled right?

wild-endeavor (approved):
we should merge this... just wondering about edge cases.

src/flyte/_internal/controllers/remote/_controller.py
kumare3: We maintain counters local to per parent action for every task. This is for concurrency within reusable environments.

PR #608, opened 2026-02-05
wandb: default to single run for multi-node

samhita-alla:
This PR:

* Adds a new config parameter, `rank_scope`, to distinguish between **global** and **worker** scopes (explained in the matrix below).
* Defaults to a **single W&B run for multi-node jobs**. The following matrix summarizes the behavior across run modes:

| run_mode  | rank_scope | Who initializes W&B?            | W&B Runs | Grouping  |
|-----------|------------|----------------------------------|----------|-----------|
| `"auto"`  | `"global"` | global rank 0 only               | 1        | –         |
| `"auto"`  | `"worker"` | local_rank 0 per worker          | N        | –         |
| `"shared"`| `"global"` | all ranks (shared mode)          | 1        | –         |
| `"shared"`| `"worker"` | all ranks (shared mode)          | N        | –         |
| `"new"`   | `"global"` | all ranks                        | N×M      | 1 group   |
| `"new"`   | `"worker"` | all ranks                        | N×M      | N groups  |

**N** = number of workers / nodes
**M** = number of processes per worker

* Fixes a bug in the W&B run initialization logic for distributed contexts by moving initialization into the **function context** (instead of execution time), ensuring `wandb.init()` runs correctly on each worker.
* Updates tests, examples, and the README accordingly.

### DIFF
--- a/plugins/wandb/examples/example_distributed.py
+++ b/plugins/wandb/examples/example_distributed.py
@@ -1,17 +1,5 @@
 """
-Example: Distributed Training with W&B Integration
-
 This example demonstrates all distributed training scenarios with W&B logging.
-
-Parameters:
-- run_mode: Controls how W&B runs are created
-  - "auto" (default): Smart behavior based on rank_scope
-  - "shared": All ranks share run(s) (uses W&B shared mode)
-  - "new": Each rank gets its own run (grouped in UI)
-
-- rank_scope: Controls scope for run creation and grouping
-  - "global" (default): Global scope - 1 run/group across all workers
-  - "worker": Worker scope - 1 run/group per worker
 
 Behavior Matrix (Multi-node):
 | run_mode | rank_scope | Who logs?           | W&B Runs | Grouping        |


plugins/wandb/examples/example_distributed.py
```diff
@@ -1,30 +1,42 @@
...
-
-2. Multi-node multi-GPU:
-   - run_mode="auto" (default): Local rank 0 of each worker logs → N W&B runs (1 per worker)
-   - run_mode="shared": All GPUs per worker log to shared run → N W&B runs (1 per worker)
-   - run_mode="new": Each GPU gets its own run (grouped per worker) → NxGPUs W&B runs
+This example demonstrates all distributed training scenarios with W&B logging.
+
+Parameters:
+- run_mode: Controls how W&B runs are created
+  - "auto" (default): Smart behavior based on rank_scope
```
kumare3: what does smart here mean?

samhita-alla: haha, we should be explicit here. that's what claude assumed :)

kumare3 (approved):

PR #613, opened 2026-02-05
Add first-class support for typeddict

cosmicBboy:
This PR adds support for `TypedDict`.

Note: to be merged after https://github.com/flyteorg/flyte-sdk/pull/597

## Why is this needed?

Many AI frameworks like LangGraph, LangChain, PydanticAI etc heavily rely on types like tuple, NamedTuple, and TypedDict in their type system. Rather than pickling, this PR preserves observability by serializing these types in a way that shows up in the UI

### DIFF
--- a/src/flyte/cli/_params.py
+++ b/src/flyte/cli/_params.py
@@ -536,6 +536,7 @@ class JsonParamType(click.ParamType):
     def _convert_to_typeddict(self, parsed_value: typing.Any, typeddict_type: typing.Type) -> dict:
         """Convert a parsed JSON value to a TypedDict."""
         from pydantic import create_model
+        from typing_extensions import NotRequired, Required, get_args, get_origin
 
         if not isinstance(parsed_value, dict):
             raise click.BadParameter(f"Expected dict for TypedDict type, got {type(parsed_value)}")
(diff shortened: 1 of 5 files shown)


src/flyte/types/_tuple_dict.py
```diff
@@ -0,0 +1,879 @@
...
+    def _model_to_value(self, model_instance: BaseModel, expected_type: Type) -> dict:
+        """Convert a Pydantic model instance back to a TypedDict."""
+        annotations = getattr(expected_type, "__annotations__", {})
+        result = {}
+        for name, field_type in annotations.items():
+            if hasattr(model_instance, name):
+                value = getattr(model_instance, name)
+                # Recursively convert nested values
+                converted_value = self._convert_value_from_pydantic(value, field_type)
+                result[name] = converted_value
```
andreahlert, line 730: this includes NotRequired fields as None in the output - e.g. a TypedDict with `tool_calls: NotRequired[list[dict]]` that wasn't provided comes back as `{'tool_calls': None}` instead of the key being absent. breaks downstream code doing `if 'tool_calls' in result`

should probably check `__required_keys__` and skip when value is None for optional fields

src/flyte/cli/_params.py
```diff
@@ -529,6 +533,59 @@ def _convert_to_namedtuple(self, parsed_value: typing.Any, namedtuple_type: typi
...
+        annotations = getattr(typeddict_type, "__annotations__", {})
+        required_keys = getattr(typeddict_type, "__required_keys__", frozenset(annotations.keys()))
+        field_definitions: typing.Dict[str, typing.Any] = {}
+
+        for field_name, field_type in annotations.items():
+            if field_name in required_keys:
+                field_definitions[field_name] = (field_type, ...)
+            else:
+                # Optional fields get a default of None
+                field_definitions[field_name] = (typing.Optional[field_type], None)
```
andreahlert: this passes NotRequired[T] straight into Pydantic without unwrapping, so you end up with Optional[NotRequired[T]] which crashes with PydanticSchemaGenerationError. the transformer side handles it correctly with _unwrap_typeddict_field_type but this code path doesn't call it

repro: https://gist.github.com/andreahlert/e3afa92494a180e59a296c8a51296240

src/flyte/types/_tuple_dict.py
```diff
@@ -0,0 +1,879 @@
...
+        annotations = getattr(t, "__annotations__", {})
+        required_keys = getattr(t, "__required_keys__", frozenset(annotations.keys()))
+
+        # Check all required keys are present
+        for key in required_keys:
+            if key not in v:
+                raise TypeTransformerFailedError(f"Missing required key '{key}' for TypedDict {t}")
+
+    def _create_wrapper_model(
+        self, t: Type, _model_cache: Optional[Dict[Type, Type[BaseModel]]] = None
```
andreahlert, line 599: minor: base class signature is _create_wrapper_model(self, t) so the cache param never gets passed from get_literal_type/to_literal/to_python_value. works fine for normal cases but the cache won't persist for self-referential TypedDicts

PR #618, opened 2026-02-05
Use enum names instead of values in literals

EngHabu:

### DIFF
--- a/src/flyte/types/_type_engine.py
+++ b/src/flyte/types/_type_engine.py
@@ -969,9 +969,10 @@ class EnumTransformer(TypeTransformer[enum.Enum]):
                     parsed."
             )
 
+        values = [v.value for v in t]  # type: ignore
+        if not isinstance(values[0], str):
+            raise TypeTransformerFailedError("Only EnumTypes with name of value are supported")
         names = [v.name for v in t]  # type: ignore
-        if not isinstance(names[0], str):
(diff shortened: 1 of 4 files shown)


src/flyte/types/_type_engine.py
```diff
@@ -864,15 +973,21 @@ def get_literal_type(self, t: Type[T]) -> LiteralType:
...
+        names = [v.name for v in t]  # type: ignore
+        if not isinstance(names[0], str):
+            raise TypeTransformerFailedError("Only EnumTypes with name of string are supported")
+        return LiteralType(enum_type=types_pb2.EnumType(values=names))
 
     async def to_literal(self, python_val: enum.Enum, python_type: Type[T], expected: LiteralType) -> Literal:
         if isinstance(python_val, str):
             # this is the case when python Literals are used as enums
-            if python_val not in expected.enum_type.values:
+            if python_val.__getattribute__("name"):
```
AdilFayyaz, line 981: I'm not sure about this but from what I understood, if someone declares a Literal["a","b"], this would be dynamically converted to a LiteralEnum. But wouldn't the python_val always remain a string, e.g. 'a' or 'b'? If so, wouldn't the `.__getattribute__("name")` fail?

pingsutw: Literal will become pickle for now

pingsutw (approved):

PR #620, opened 2026-02-06
fix df local sync in nested types

cosmicBboy:
fixes deadlock issue where dataframes were being uploaded using an asyncify upload file function while pydantic and dataclass serialization was using the loop manager. The solution is to pre-invoke the lazy uploaders (currently implemented by the flyte dataframe type to upload raw dfs to object store from local) before calling `model_dump_json` in pydantic and dataclass transformers.

### DIFF
--- a/src/flyte/types/_type_engine.py
+++ b/src/flyte/types/_type_engine.py
@@ -77,17 +77,26 @@ async def _invoke_lazy_uploaders(obj: typing.Any) -> None:
     Args:
         obj: The object to process (can be a Pydantic model, dataclass, or collection)
     """
+    if obj is None:
+        logger.debug("Object is None, skipping lazy uploaders.")
+        return
+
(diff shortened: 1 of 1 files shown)


src/flyte/types/_type_engine.py
```diff
@@ -381,6 +441,11 @@ async def to_literal(
         python_type: Type[BaseModel],
         expected: LiteralType,
     ) -> Literal:
+        # Pre-process the model to invoke any lazy uploaders on nested Flyte IO types.
+        # This ensures uploads happen in the syncify context where gRPC clients work correctly,
+        # and prevents deadlocks when @model_serializer tries to run async code via loop_manager.
+        await _invoke_lazy_uploaders(python_val)
```
kumare3, line 464: This is very slow, can we just do it when local?

PR #621, opened 2026-02-06
Introducing TUI: Flyte local runs

kumare3:
This PR introduces a local mode on steriods!

you can now visualize your runs using an experimental TUI. Flyte caching, reports, groups are all accessible in the local TUI with input / output tracking.

Screen Recording 2026-02-05 at 10 17 55 PM

### DIFF
--- a/examples/basics/large_slow_run.py
+++ b/examples/basics/large_slow_run.py
@@ -25,4 +25,4 @@ async def parallel_main(n: int):
 @env.task
 async def main(n: int):
     for i in range(n):
-       await sleeper()
+        await sleeper()


kumare3:
> Hey @kumare3, pulled the branch and tested locally (Python 3.12, Textual 7.5.0). Cache tracking works well, cached calls show ~0.00s in the detail panel so you immediately see what's hitting cache.
> 
> I noticed `main` shows up twice in the tree though:
> 
> ```
> [ok] my-env.main
>   └── [ok] my-env.main
>         └── [ok] my-env.sleeper
> ```
> 
> Looks like `_run_local` records a start for the root action, and then `get_action_outputs` records another one for the same task with a generated action ID. First time I saw it I thought something was running twice.
> 
> Also if the workflow crashes bypassing the controller's try/except, the root node stays showing `[>]` (running) while the subtitle says "failed". You could catch the worker error in `on_worker_state_changed` and mark any still-running nodes as failed.
> 
> `TaskCallSequencer` refactor is clean, makes sense to share between both controllers. Now I kinda want this for remote runs too.

Try again been pushing

wild-endeavor (approved):

PR #627, opened 2026-02-07
Persistence of local runs

kumare3:
First
```bash
flyte create config --local-persistence
```
Then, this or flyte.run will be recorded. you can also set in 
`flyte.init(local_persistence=True)`
```
flyte run --local [--tui] 
```
And now you can checkout and browse all old runs

```
flyte start tui
```
https://github.com/user-attachments/assets/60cd7513-ef0a-4f45-8b67-ac214bb81982

### DIFF
--- a/src/flyte/_persistence/_db.py
+++ b/src/flyte/_persistence/_db.py
@@ -47,11 +47,17 @@ CREATE TABLE IF NOT EXISTS runs (
 """
 
 
+_RUNS_INDEXES = [
+    "CREATE INDEX IF NOT EXISTS idx_runs_action_start ON runs (action_name, start_time)",
+    "CREATE INDEX IF NOT EXISTS idx_runs_status_start ON runs (status, start_time)",
+]
+
 _RUNS_MIGRATIONS = [
     "ALTER TABLE runs ADD COLUMN has_report INTEGER DEFAULT 0",
     "ALTER TABLE runs ADD COLUMN context TEXT",
     "ALTER TABLE runs ADD COLUMN group_name TEXT",
     "ALTER TABLE runs ADD COLUMN log_links TEXT",
+    *_RUNS_INDEXES,
 ]
 
 
@@ -105,6 +111,8 @@ class LocalDB:
         conn = await aiosqlite.connect(db_path)
         await conn.execute(_TASK_CACHE_DDL)
         await conn.execute(_RUNS_DDL)
+        for idx_stmt in _RUNS_INDEXES:
+            await conn.execute(idx_stmt)
         await conn.commit()
         LocalDB._conn = conn
         # Also open a sync connection for sync callers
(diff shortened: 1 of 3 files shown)


src/flyte/cli/_create.py
```diff
@@ -284,6 +292,10 @@ def config(
     if image_builder:
         image["builder"] = image_builder
 
+    local: Dict[str, Any] = {}
+    if local_persistence:
+        local["persistence"] = True
+
     if not admin and not task:
         raise click.BadParameter("At least one of --endpoint or --org must be provided.")
```
machichima, line 300: I think we do not need this check if we only set `local`? As those are for remote

src/flyte/_persistence/_run_store.py
```diff
@@ -0,0 +1,204 @@
...
+                (error, time.time(), run_name, action_name),
+            )
+            conn.commit()
+
+    # -- Read methods (sync) --
+
+    @staticmethod
+    def list_runs_sync() -> list[RunRecord]:
+        """List top-level runs (action_name='a0') ordered by start_time DESC."""
+        with LocalDB._write_lock:
```
machichima: `list_runs_sync`, `list_actions_for_run_sync`, and `get_action_sync` are using `SELECT` only, I think they do not need `write_lock`?

src/flyte/_persistence/_db.py
```diff
@@ -0,0 +1,174 @@
...
+                await LocalDB._initialize_async()
+            else:
+                LocalDB._initialize_sync_inner()
+
+    @staticmethod
+    async def _initialize_async():
+        db_path = LocalDB._get_db_path()
+        conn = await aiosqlite.connect(db_path)
+        await conn.execute(_TASK_CACHE_DDL)
+        await conn.execute(_RUNS_DDL)
```
machichima, line 113: nit: maybe we can add index after this to optimize queries (also in `initialize_sync`)

src/flyte/cli/_create.py
kumare3: that is interesting, lets do this as a follow up

src/flyte/_persistence/_run_store.py
kumare3: great call out (vibing)

src/flyte/_persistence/_db.py
kumare3: like

kumare3: which index do we need, i dont see any queries that need an index right now right?

kumare3: we can add them as we need them?

kumare3: i found one place, for list runs and sorting it by time would help. So i am adding it now

wild-endeavor (approved):
is this backwards compatible or will we need to clear the local db?

PR #628, opened 2026-02-07
Support local app serving

cosmicBboy:
# Summary
This PR implements local app serving for AppEnvironment, enabling developers to serve Flyte apps on localhost for rapid local development and testing. It introduces a symmetry with the existing `flyte.with_runcontext(mode="local") ` pattern by adding `flyte.with_servecontext(mode="local")` for apps.

Key changes:
- New `_LocalApp` class and local serving infrastructure in `_serve.py` (server-decorator, command/subprocess, and FastAPI-based apps)
- `AppEnvironment.endpoint` now transparently resolves to localhost when the app is served locally
`ephemeral_ctx` / `ephemeral_ctx_sync` context managers on both _LocalApp and remote App for scoped activate/deactivate lifecycle
- `_LocalApp` `activate` and `deactivate` methods mirror the methods in remote `App` object.
  - `activate`: if the server is not running locally, starts it, and optionally waits for the server to be active.
  - `deactivate`: stop the server locally
- `--local` flag on the flyte serve CLI command
- Signal handler and atexit cleanup to ensure child processes are torn down on `Ctrl-C` or process exit
- 6 example scripts and comprehensive unit tests

### DIFF
--- a/examples/apps/local_command_app.py
+++ b/examples/apps/local_command_app.py
@@ -30,7 +30,7 @@ if __name__ == "__main__":
     flyte.init()
 
     # Serve the app locally (non-blocking)
-    local_app = flyte.with_servecontext(mode="local").serve(app_env)
+    local_app = flyte.with_servecontext(mode="local", health_check_path="/").serve(app_env)
 
     # Wait for the app to be ready
     local_app.activate(wait=True)


src/flyte/remote/_app.py
```diff
@@ -174,19 +175,41 @@ async def activate(self, wait: bool = False) -> App:
...
     @syncify
-    async def deactivate(self, wait: bool = False):
+    async def deactivate(self, wait: bool = False) -> App:
         """
         Stop the app
         :param wait: Wait for the app to reach the deactivated state
         """
         if self.is_deactivated():
-            return
+            return self
```
kumare3, line 184: Why do we need to mimic remote for local?

kumare3: How can we just handle this in serve?

src/flyte/_serve.py
```diff
@@ -14,12 +25,232 @@
...
+        except (URLError, OSError):
+            pass
+
+        return False
+
+    def is_deactivated(self) -> bool:
+        """
+        Check if the app is currently deactivated.
+        """
+        return self._process is None or self._process.poll() is not None
```
andreahlert: This is always `True` for thread-based apps since `_process` is never set. `FastAPIAppEnvironment.__post_init__` sets `_server = self._fastapi_app_server`, so it goes through `_serve_local_with_server_func` (daemon thread, no subprocess). The `assert local_app.is_deactivated()` in `local_app_ephemeral_server.py` always passes regardless of whether the server actually stopped.

`_is_running()` already handles both thread and process correctly but `is_deactivated` does not. Same gap in `deactivate()` - the terminate/kill block is conditional on `self._process is not None`, so for thread-based apps it only cleans up the registry while uvicorn keeps serving.

src/flyte/_serve.py
```diff
@@ -14,12 +25,232 @@
...
+    original = _ORIGINAL_SIGINT_HANDLER if signum == signal.SIGINT else _ORIGINAL_SIGTERM_HANDLER
+    if callable(original):
+        original(signum, frame)
+    elif original == signal.SIG_DFL:
+        # Re-raise with default disposition
+        signal.signal(signum, signal.SIG_DFL)
+        os.kill(os.getpid(), signum)
+
+
+def _install_signal_handlers() -> None:
```
kumare3, line 129: best effort, but its cool. We will have to figure out how to clean up in case they leak.

kumare3 (approved):

kumare3:
Looks great

src/flyte/remote/_app.py
andreahlert: the idea is that calling code doesn't have to care if it's local or remote, you just call activate/deactivate either way. but yeah right now it's just duck typing by coincidence. might be worth a Protocol to make it intentional, or just keep the local handle minimal with endpoint + stop().

src/flyte/_serve.py
andreahlert: start_new_session=True on Popen would help for subprocesses, kills the whole group. for thread-based apps (fastapi path) there's no shutdown mechanism at all right now though - deactivate() skips them since _process is None. probably needs a threading.Event or similar.

PR #629, opened 2026-02-07
feat: Add Hugging Face Datasets plugin

andreahlert:
## Summary

Port of the [flytekit-huggingface](https://github.com/flyteorg/flytekit/tree/master/plugins/flytekit-huggingface) plugin to flyte-sdk v2, enabling native support for `datasets.Dataset` as a Flyte DataFrame type.

- **DataFrameEncoder/Decoder** for `datasets.Dataset` with Parquet serialization
- **Cloud storage support** (S3, GCS, Azure) via fsspec-compatible storage options
- **Anonymous S3 fallback** for public datasets (mirrors Polars plugin pattern)
- **Column filtering** on both encode and decode via type annotations
- **Auto-registration** via `flyte.plugins.types` entry point

Follows the same patterns as the existing Polars plugin.

### Usage Example

```python
import flyte
import datasets

env = flyte.TaskEnvironment("hf-example")

@env.task
async def create_dataset() -> datasets.Dataset:
    return datasets.Dataset.from_dict({
        "text": ["hello", "world"],
        "label": [0, 1],
    })

@env.task
async def process(ds: datasets.Dataset) -> int:
    return len(ds)
```

## Test plan

- [x] Type recognition tests (Dataset, with columns, with format annotations)
- [x] Handler registration and property tests
- [x] Encode/decode roundtrip tests
- [x] DataFrame wrapper integration test
- [x] Raw task I/O test via `flyte.TaskEnvironment`
- [x] Column subsetting on decode
- [x] Various data types (int, float, str, bool)
- [x] Empty dataset roundtrip
- [x] Storage options tests (S3, GCS, Azure, anonymous, unknown protocol)

### DIFF
--- a/plugins/huggingface/README.md
+++ b/plugins/huggingface/README.md
@@ -1,0 +1,14 @@
+# Hugging Face Datasets Plugin
+
+This plugin provides native support for **Hugging Face Datasets** in Flyte, enabling seamless serialization and deserialization of `datasets.Dataset` objects through Parquet format.
+
+The plugin supports:
+- `datasets.Dataset` - Hugging Face's dataset type for structured data
+- Automatic column filtering based on type annotations
+- Efficient Parquet-based storage for large datasets
+
+To install the plugin, run the following command:
+
+```bash
+pip install --pre flyteplugins-huggingface
+```
--- a/plugins/huggingface/README.md
+++ b/plugins/huggingface/README.md
@@ -1,14 +1,64 @@
 # Hugging Face Datasets Plugin
 
-This plugin provides native support for **Hugging Face Datasets** in Flyte, enabling seamless serialization and deserialization of `datasets.Dataset` objects through Parquet format.
+Native support for `datasets.Dataset` as a Flyte DataFrame type, with Parquet serialization and column filtering.
 
-The plugin supports:
-- `datasets.Dataset` - Hugging Face's dataset type for structured data
-- Automatic column filtering based on type annotations
-- Efficient Parquet-based storage for large datasets
-
-To install the plugin, run the following command:
+## Installation
 
 ```bash
 pip install flyteplugins-huggingface
+```
+
+## Usage
+
+```python
+import flyte
+import datasets
+
+env = flyte.TaskEnvironment(
+    name="hf-example",
+    image=flyte.Image.from_debian_base().with_pip_packages(
+        "flyteplugins-huggingface",
+    ),
+)
+
+
+@env.task
+async def create_dataset() -> datasets.Dataset:
+    return datasets.Dataset.from_dict({
+        "text": ["hello", "world", "foo"],
+        "label": [0, 1, 0],
+    })
+
+
+@env.task
+async def filter_positive(ds: datasets.Dataset) -> datasets.Dataset:
+    return ds.filter(lambda x: x["label"] == 1)
+
+
+@env.task
+async def main() -> int:
+    ds = await create_dataset()
+    filtered = await filter_positive(ds)
+    return len(filtered)
+
+
+if __name__ == "__main__":
+    flyte.init_from_config()
+    run = flyte.run(main)
+    print(run.url)
+```
+
+## Column filtering
+
+Use type annotations to load only specific columns:
+
+```python
+from typing import Annotated
+from collections import OrderedDict
+
+@env.task
+async def load_text_only(
+    ds: Annotated[datasets.Dataset, OrderedDict(text=str)],
+) -> list:
+    return ds["text"]
 ```
--- a/plugins/huggingface/README.md
+++ b/plugins/huggingface/README.md
@@ -10,5 +10,5 @@ The plugin supports:
 To install the plugin, run the following command:
 
 ```bash
-pip install --pre flyteplugins-huggingface
+pip install flyteplugins-huggingface
 ```
--- a/plugins/huggingface/README.md
+++ b/plugins/huggingface/README.md
@@ -1,6 +1,6 @@
 # Hugging Face Datasets Plugin
 
-Native support for `datasets.Dataset` as a Flyte DataFrame type, with Parquet serialization and column filtering.
+Native support for HuggingFace Datasets in Flyte: prefetch datasets from the Hub to remote storage and pass `datasets.Dataset` between tasks with automatic Parquet serialization.
 
 ## Installation
 
@@ -8,7 +8,24 @@ Native support for `datasets.Dataset` as a Flyte DataFrame type, with Parquet se
 pip install flyteplugins-huggingface
 ```
 
-## Usage
+## Prefetch from HuggingFace Hub
+
+Stream a dataset from the Hub directly to Flyte's remote storage:
+
+```python
+import flyte
+from flyteplugins.huggingface import hf_dataset
+
+flyte.init(endpoint="my-flyte-endpoint")
+
+run = hf_dataset(repo="stanfordnlp/imdb", split="train")
+run.wait()
+data_dir = run.outputs()[0]  # flyte.io.Dir with parquet files
+```
+
+## Type transformer
+
+Pass `datasets.Dataset` between tasks with automatic serialization:
 
 ```python
 import flyte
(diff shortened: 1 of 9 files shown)


andreahlert:
@pingsutw @cosmicBboy @kumare3 could you take a look? This ports the huggingface plugin from flytekit to v2, same approach as the polars plugin.

plugins/huggingface/src/flyteplugins/huggingface/df_transformer.py
```diff
@@ -0,0 +1,170 @@
+import functools
+import os
+import typing
+from pathlib import Path
+
+import flyte.storage as storage
+from flyte._logging import logger
+from flyte._utils import lazy_module
```
kumare3: let's not use `_` imports. @pingsutw maybe we need to add it to `extend`

plugins/huggingface/src/flyteplugins/huggingface/df_transformer.py
```diff
@@ -0,0 +1,170 @@
+import functools
+import os
+import typing
+from pathlib import Path
+
+import flyte.storage as storage
+from flyte._logging import logger
```
kumare3: Logger is actually available - https://github.com/flyteorg/flyte-sdk/blob/28396f21fa1024ed76c0b43225d2a3bda3c3963e/src/flyte/__init__.py#L20

plugins/huggingface/src/flyteplugins/huggingface/df_transformer.py
```diff
@@ -0,0 +1,170 @@
...
+
+        if structured_dataset_type.columns:
+            columns = [c.name for c in structured_dataset_type.columns]
+            existing = set(df.features.keys())
+            to_remove = [c for c in existing if c not in columns]
+            if to_remove:
+                df = df.remove_columns(to_remove)
+
+        filesystem = storage.get_underlying_filesystem(path=path)
+        storage_options = get_hf_storage_options(protocol=filesystem.protocol)
```
kumare3: We should probably use `xet data transfer now right?` if possible.

plugins/huggingface/src/flyteplugins/huggingface/df_transformer.py
andreahlert: Good catch, switched to `from flyte import logger` since it's already exported in `__init__.py`. Fixed in b5f2178.

plugins/huggingface/src/flyteplugins/huggingface/df_transformer.py
andreahlert: Not really, from my understanding xet is for HF Hub transfers (load_dataset, push_to_hub), but here the storage_options are for writing parquet to Flyte's own storage (S3/GCS/Azure), same pattern as the Polars plugin. For Hub-side acceleration that would be at the task level via env vars like the image_classification example does with `HF_XET_HIGH_PERFORMANCE=1`.

plugins/huggingface/src/flyteplugins/huggingface/df_transformer.py
andreahlert: Thanks, will update once `lazy_module` is exposed through the public API.

plugins/huggingface/src/flyteplugins/huggingface/df_transformer.py
```diff
@@ -0,0 +1,170 @@
...
+    datasets = lazy_module("datasets")
+
+
+def get_hf_storage_options(protocol: typing.Optional[str], anonymous: bool = False) -> typing.Dict[str, typing.Any]:
+    """Get fsspec-compatible storage options for HuggingFace datasets.
+
+    HuggingFace datasets uses fsspec for remote I/O, so storage options
+    follow the fsspec/s3fs/gcsfs/adlfs conventions.
+    """
+    from flyte._initialize import get_storage
```
kumare3: this method should not be used outside

plugins/huggingface/src/flyteplugins/huggingface/df_transformer.py
```diff
@@ -0,0 +1,170 @@
...
+    if not protocol:
+        return {}
+
+    try:
+        storage_config = get_storage()
+    except InitializationError:
+        storage_config = None
+
+    match protocol:
+        case "s3":
```
kumare3: this looks like its using s3?

plugins/huggingface/src/flyteplugins/huggingface/df_transformer.py
```diff
@@ -0,0 +1,116 @@
...
+)
+from flyteidl2.core import literals_pb2, types_pb2
+
+if typing.TYPE_CHECKING:
+    import datasets
+else:
+    datasets = lazy_module("datasets")
+
+
+def get_hf_storage_options(protocol: typing.Optional[str], anonymous: bool = False) -> typing.Dict[str, typing.Any]:
```
kumare3: why is this function needed?

plugins/huggingface/src/flyteplugins/huggingface/df_transformer.py
```diff
@@ -0,0 +1,116 @@
...
+
+        if structured_dataset_type.columns:
+            columns = [c.name for c in structured_dataset_type.columns]
+            existing = set(df.features.keys())
+            to_remove = [c for c in existing if c not in columns]
+            if to_remove:
+                df = df.remove_columns(to_remove)
+
+        filesystem = storage.get_underlying_filesystem(path=path)
+        storage_options = get_hf_storage_options(protocol=filesystem.protocol)
```
kumare3: i dont see any hf storage options, i think you can drop this entirely

plugins/huggingface/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+The plugin supports:
+- `datasets.Dataset` - Hugging Face's dataset type for structured data
+- Automatic column filtering based on type annotations
+- Efficient Parquet-based storage for large datasets
+
+To install the plugin, run the following command:
+
+```bash
+pip install --pre flyteplugins-huggingface
```
kumare3: No --pre

andreahlert: Done

plugins/huggingface/src/flyteplugins/huggingface/df_transformer.py
andreahlert: Done.

andreahlert:
> Can you add an example?
> 
> I am thinking either hf:// file system do we need it Cc @samhita-alla

I think we do, at least for the read side. HF Hub auto-converts all datasets to Parquet now, so hf:// would let Flyte track what was actually consumed instead of users doing opaque load_dataset calls. The fsspec backend already exists.

Immutability is a concern since Hub data can change, but for pinned dataset revisions (hf://datasets/squad@v1.0) it works.

Write side is trickier and probably not needed for a first pass.

Open to what is best in your opinion.

samhita-alla:
hey @andreahlert, thank you for creating this PR! sharing my thoughts on how we can improve the plugin from a UX perspective:

---
the two primary use cases i see are:
- loading a dataset from HF Hub: to use in training or processing tasks
- creating a dataset in a task: to pass between tasks for training and optionally publish to HF Hub

for loading, i think this should follow the same pattern as our [HF model prefetch utility](https://github.com/flyteorg/flyte-sdk/blob/main/src/flyte/prefetch/_hf_model.py) i.e. stream from the hub rather than fully materializing the dataset into memory. the output should be a reference (e.g. `Dir`) stored in cloud storage, which downstream tasks can then stream from.

for creating, when a task outputs a `datasets.Dataset`, we should:
- stream-write it to remote storage
- optionally support pushing to HF Hub 

i don’t think a type transformer alone will be useful here.

---

i know the PR mirrors what’s in the v1 plugin, but i think this approach would make it genuinely useful for users. not sure if you have the bandwidth to work on this. if not, totally understandable. we can leave the PR open and i can take a pass at it later!

andreahlert:
Hey everybody. Updated based on @samhita-alla  feedback.

Took a step back and rethought the approach. The plugin now has a hf_dataset() prefetch that streams parquet from HF Hub to Flyte storage (same pattern as flyte.prefetch.hf_model), batched writes in the encoder, and an example showing the full flow. 

Not sure if I got the prefetch design exactly right so would appreciate your take on it. 

Hub push is left out for now, figured that's better as a follow-up. WDYT?

samhita-alla:
closing this in favor of https://github.com/flyteorg/flyte-sdk/pull/992

PR #630, opened 2026-02-08
fix: Handle list[Model], dict[str, Model] and Optional[Model] in nested type detection

andreahlert:
## Motivation

PR #426 added support for nested Pydantic/dataclass schema detection via `$ref`, but as noted in flyteorg/flyte#6887 (item 1), it does not handle `list` and `map` containers with nested models.

When a schema contains `list[NestedModel]`, `dict[str, NestedModel]`, or `Optional[NestedModel]`, the `$ref` inside `items`, `additionalProperties`, or `anyOf` was not resolved, causing `KeyError` or incorrect type inference (falling back to `str`).

### Centralized `$ref` resolution

Instead of inlining `$ref` resolution at every call site, this PR follows the same approach as flytekit PR #3375: `_get_element_type` now accepts an optional `schema` parameter and resolves `$ref` centrally. This means any container type (`list`, `dict`, `Optional`, or nested combinations) automatically gets `$ref` support.

### New helper functions

- **`_resolve_ref`**: Centralized `$ref` path resolution to schema definition
- **`_resolve_property_type`**: Resolves a JSON schema property to a `(name, type)` tuple, handling `$ref`, `anyOf`, and all type variants
- **`_resolve_single_type`**: Resolves the inner type of `Optional[T]` from the first `anyOf` element, properly handling `$ref`, `array`, `object`, and primitive types
- **`_resolve_typed_property`**: Resolves a property with a known type string (array, object, primitives)
- **`_is_optional_anyof`**: Validates that an `anyOf` truly represents `Optional[T]` (exactly 2 elements, second is `null`) instead of blindly assuming any `anyOf` with `$ref` is Optional

### Bug fixes

- **`Optional[T]` default values**: Fields not in the schema's `required` list now get `default=None` in the generated dataclass, so constructing with omitted optional fields works correctly
- **`Optional[List[Model]]` / `Optional[Dict[str, Model]]`**: Previously would crash with `KeyError` because the code accessed `property_val["items"]` instead of `anyOf[0]["items"]` - now handled correctly via `_resolve_single_type`
- **`Union[TypeA, TypeB]`**: Previously any `anyOf` with `$ref` was blindly wrapped in `Optional` - now properly validated
- **Optional field inclusion**: Properties not in `required` are included in `property_order` so fields with defaults (like `Optional[X] = None`) are not skipped during schema traversal

### `__init__` wrapper

Extended dict-to-object conversion to handle `list[dict]` -> `list[Model]` and `dict[str, dict]` -> `dict[str, Model]`.

## Tests

Added `test_nested_lists_maps.py` with 10 test cases covering:
- `list[NestedModel]` roundtrip (Pydantic + dataclass)
- `dict[str, NestedModel]` roundtrip (Pydantic + dataclass)
- `Optional[NestedModel]` with value and with `None`
- `Optional[NestedModel]` omitted (default to `None`)
- Mixed initialization (dict + object in same list)
- Empty collections (`[]`, `{}`)

All existing type engine tests pass with zero regressions.

Ref: flyteorg/flyte#6887

### DIFF
--- a/src/flyte/types/_type_engine.py
+++ b/src/flyte/types/_type_engine.py
@@ -2208,6 +2208,21 @@ def _get_element_type(
         # Element type of Optional[int] is [integer, None]
         return typing.Optional[_get_element_type({"type": element_type[0]}, schema)]  # type: ignore
 
+    # Handle nested array (e.g., List[List[int]])
+    if element_type == "array":
+        inner_items = element_property.get("items", {})
+        return typing.List[_get_element_type(inner_items, schema)]  # type: ignore
+
+    # Handle nested object / dict (e.g., List[Dict[str, int]])
+    if element_type == "object":
+        additional = element_property.get("additionalProperties")
+        if additional:
+            return typing.Dict[str, _get_element_type(additional, schema)]  # type: ignore
+        # Nested dataclass-like object with a title
+        if element_property.get("title"):
+            return convert_mashumaro_json_schema_to_python_class(element_property, element_property["title"])
+        return dict
+
     if element_type == "string":
         return str
     elif element_type == "integer":
--- a/tests/flyte/type_engine/pydantic/test_nested_lists_maps.py
+++ b/tests/flyte/type_engine/pydantic/test_nested_lists_maps.py
@@ -38,6 +38,18 @@ class Profile(BaseModel):
     address: Optional[Tag] = None
 
 
+class NestedLists(BaseModel):
+    matrix: List[List[int]]
+
+
+class NestedDicts(BaseModel):
+    lookup: List[Dict[str, int]]
+
+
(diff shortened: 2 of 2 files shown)


andreahlert:
@wild-endeavor @kumare3 Hey! Picked up item 1 from #6887 - the list/map gap in PR #426.

`list[NestedModel]`, `dict[str, NestedModel]` and `Optional[NestedModel]` were all hitting `KeyError` because `$ref` wasn't being resolved in array items, additionalProperties, or anyOf.

Bonus bug: optional fields with defaults were silently skipped because `property_order` only iterated over `required`, so `Optional[Model] = None` never made it into the generated class.

kumare3:
cc @AdilFayyaz can you PTAL?

PR #632, opened 2026-02-08
feat: Add Google Gemini plugin for Flyte

andreahlert:
## Summary

This PR adds a new plugin for integrating Google's Gemini API with Flyte, following the same patterns as the existing OpenAI and Anthropic plugins.

### Features

- **`function_tool` decorator**: Convert Flyte tasks to Gemini function declarations automatically
- **Automatic type conversion**: Python type hints to JSON schema for tool input validation
- **`Agent` class**: Configure Gemini agents with model, instructions, and tools
- **`run_agent` function**: Execute agent loops that handle function calling and responses
- **Support for sync and async tasks**
- **Manual function calling control**: Uses AFC (Automatic Function Calling) disabled mode for explicit tool execution management

### Usage Example

```python
import flyte
from flyteplugins.gemini import function_tool, run_agent

env = flyte.TaskEnvironment("gemini-agent")

@env.task
async def get_weather(city: str) -> str:
    """Get the current weather for a city."""
    return f"Weather in {city}: sunny, 72F"

@env.task
async def agent_task(prompt: str) -> str:
    tools = [function_tool(get_weather)]
    return await run_agent(
        prompt=prompt,
        tools=tools,
        model="gemini-2.5-flash",
    )
```

### Files Added

- `plugins/gemini/` - Plugin package
  - `pyproject.toml` - Package configuration
  - `README.md` - Documentation
  - `src/flyteplugins/gemini/agents/_function_tools.py` - Core implementation
  - `tests/test_agents.py` - Test suite (19 tests)
- `examples/genai/gemini_agent.py` - Example workflow

## Test Plan

- [x] All 19 unit tests pass
- [x] Type conversion tests (string, int, float, bool, list, dict, Optional)
- [x] Function schema extraction tests
- [x] FunctionTool creation tests (regular functions, Flyte tasks, async)
- [x] Agent configuration tests
- [x] Tool execution tests (sync and async)
- [x] run_agent tests (tool calls, errors, max iterations, finish reasons)

## Why Gemini?

Google Gemini is one of the leading LLM providers alongside OpenAI and Anthropic. Having first-class support enables users to:
- Choose the best model for their use case
- Easily switch between providers
- Use Gemini's unique capabilities (multimodal, large context windows, cost efficiency)

This plugin mirrors the Anthropic plugin's API design, making it easy to switch between providers.

### DIFF
--- a/plugins/gemini/src/flyteplugins/gemini/agents/_function_tools.py
+++ b/plugins/gemini/src/flyteplugins/gemini/agents/_function_tools.py
@@ -20,73 +20,6 @@ from google.genai import types
 
 logger = logging.getLogger(__name__)
 
-# Type mapping from Python types to JSON schema types
-TYPE_MAP: dict[type, str] = {
-    str: "string",
-    int: "integer",
-    float: "number",
-    bool: "boolean",
-    list: "array",
-    dict: "object",
-}
-
-
-def _python_type_to_json_schema(py_type: type) -> dict[str, typing.Any]:
-    """Convert a Python type hint to a JSON schema definition."""
-    origin = typing.get_origin(py_type)
-    args = typing.get_args(py_type)
-
-    # Handle Optional types (Union[X, None])
-    if origin is typing.Union:
-        non_none_args = [a for a in args if a is not type(None)]
-        if len(non_none_args) == 1:
-            return _python_type_to_json_schema(non_none_args[0])
-        return {"anyOf": [_python_type_to_json_schema(a) for a in non_none_args]}
-
-    # Handle list[X]
-    if origin is list:
-        if args:
-            return {"type": "array", "items": _python_type_to_json_schema(args[0])}
-        return {"type": "array"}
-
-    # Handle dict[K, V]
-    if origin is dict:
-        return {"type": "object"}
-
-    # Handle basic types
-    if py_type in TYPE_MAP:
-        return {"type": TYPE_MAP[py_type]}
-
-    # Default to string for unknown types
-    return {"type": "string"}
-
-
-def _get_function_schema(func: typing.Callable) -> dict[str, typing.Any]:
-    """Extract JSON schema from a function's type hints."""
-    sig = inspect.signature(func)
-    hints = typing.get_type_hints(func)
-
-    properties: dict[str, typing.Any] = {}
-    required: list[str] = []
-
-    for name, param in sig.parameters.items():
-        if name in ("self", "cls"):
-            continue
-
-        param_type = hints.get(name, str)
-        prop_schema = _python_type_to_json_schema(param_type)
-        properties[name] = prop_schema
-
-        # Check if parameter is required (no default value)
-        if param.default is inspect.Parameter.empty:
-            required.append(name)
-
-    return {
-        "type": "object",
-        "properties": properties,
-        "required": required,
-    }
-
 
 @dataclass
 class FunctionTool:
@@ -122,8 +55,7 @@ class FunctionTool:
         """
         if self.task is not None:
             if self.is_async:
-                return await self.task(**kwargs)
-            return await asyncio.to_thread(self.task, **kwargs)
+                return await self.task.aio(**kwargs)
         if self.is_async:
             return await self.func(**kwargs)
         return await asyncio.to_thread(self.func, **kwargs)
@@ -139,6 +71,10 @@ def function_tool(
 
     This function converts a Python function, @flyte.trace decorated function,
     or Flyte task into a FunctionTool that can be used with Gemini's function calling API.
+
+    The input_schema is derived via the Flyte type engine, producing JSON schema.
+    This ensures that Literal types, dataclasses, FlyteFile, and other Flyte-native
+    types are represented correctly.
 
     For @flyte.trace decorated functions, the tracing context is preserved
     automatically since functools.wraps maintains the original function's metadata.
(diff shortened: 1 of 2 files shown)


plugins/gemini/src/flyteplugins/gemini/agents/_function_tools.py
```diff
@@ -0,0 +1,367 @@
...
+    if py_type in TYPE_MAP:
+        return {"type": TYPE_MAP[py_type]}
+
+    # Default to string for unknown types
+    return {"type": "string"}
+
+
+def _get_function_schema(func: typing.Callable) -> dict[str, typing.Any]:
+    """Extract JSON schema from a function's type hints."""
+    sig = inspect.signature(func)
```
kumare3: you should not need to do this, this is already done as part of the AsyncFunctionTemplate - native interface

plugins/gemini/src/flyteplugins/gemini/agents/_function_tools.py
```diff
@@ -0,0 +1,367 @@
...
+        # @flyte.trace uses functools.wraps, so __name__, __doc__ and type hints
+        # are preserved. The tracing activates automatically in a task context.
+        actual_func = func
+        task = None
+        native_interface = None
+        report = False
+
+    tool_name = name or actual_func.__name__
+    tool_description = description or (actual_func.__doc__ or f"Execute {tool_name}")
+    input_schema = _get_function_schema(actual_func)
```
kumare3: you already have native_interface

plugins/gemini/src/flyteplugins/gemini/agents/_function_tools.py
```diff
@@ -0,0 +1,367 @@
...
+    async def execute(self, **kwargs) -> typing.Any:
+        """Execute the tool with the given arguments.
+
+        Async functions are awaited directly. Sync functions are run in a
+        thread executor to avoid blocking the event loop.
+        """
+        if self.task is not None:
+            if self.is_async:
+                return await self.task(**kwargs)
+            return await asyncio.to_thread(self.task, **kwargs)
```
kumare3: this is not needed and dangerous. this will start non daemon threads which will block. you dont need to do it, this is all solved by flyte, it gives you an async interface. doing the to_thread is not recommended. 

i think we should just call the async execute method. it takes care of the rest

plugins/gemini/src/flyteplugins/gemini/agents/_function_tools.py
andreahlert: I checked `NativeInterface` before going this route. It stores Python types as `Dict[str, Tuple[Type, Any]]` but doesn't expose a JSON schema conversion. Gemini's `FunctionDeclaration` needs `parameters_json_schema` as a JSON schema dict, so some conversion layer is needed regardless.

This is the same approach used in the Anthropic plugin (`plugins/anthropic/.../agents/_function_tools.py`, lines 34-66). Also, `function_tool()` needs to support regular callables and `@flyte.trace` decorated functions too, where `NativeInterface` isn't available.

Happy to refactor if there's a method I'm missing that already does this conversion.

plugins/gemini/src/flyteplugins/gemini/agents/_function_tools.py
andreahlert: `native_interface` is only set when `func` is an `AsyncFunctionTaskTemplate`. For plain callables it's `None`, so we'd still need the `inspect` fallback. The Anthropic plugin stores `native_interface` the same way but generates the schema from the raw function for this reason.

One thing I could do is use `native_interface.inputs` when available and fall back to `inspect` otherwise. Would that be the direction you'd prefer?

plugins/gemini/src/flyteplugins/gemini/agents/_function_tools.py
andreahlert: I hear you on avoiding unnecessary threads. I looked at how the SDK handles this internally, and `AsyncFunctionTaskTemplate.execute()` itself uses `run_sync_with_loop()` for sync functions, which also spins up a thread with its own event loop (`src/flyte/_utils/asyncify.py`).

The reason for `to_thread` here: for sync Flyte tasks, `task(**kwargs)` goes through `__call__` -> `controller.submit_sync()` which returns a blocking `Future.result(None)`. Without `to_thread`, that blocks the event loop while the agent loop needs to stay async to handle multiple tool calls.

That said, if there's a cleaner async path I should be using instead, I'm happy to switch. The Anthropic plugin uses the same `to_thread` pattern, so if there's a better approach we could align both.

plugins/gemini/src/flyteplugins/gemini/agents/_function_tools.py
```diff
@@ -0,0 +1,307 @@
...
+        )
+
+    async def execute(self, **kwargs) -> typing.Any:
+        """Execute the tool with the given arguments.
+
+        Async functions are awaited directly. Sync functions are run in a
+        thread executor to avoid blocking the event loop.
+        """
+        if self.task is not None:
+            if self.is_async:
```
kumare3, line 57: Not async?

kumare3: .aio is available on non async tasks too

kumare3: We updated the anthropic plugin, sorry for the confusion, can you change like that and we can merge this

PR #633, opened 2026-02-08
feat: Add DuckDB plugin

andreahlert:
## Summary

- Add a DuckDB connector plugin for running SQL queries against DuckDB as Flyte tasks
- DuckDB is an embedded analytical database (like SQLite for OLAP) that runs locally and synchronously
- Follows the same architecture and patterns as the Snowflake plugin

## Features

- In-memory and file-based database support via `DuckDBConfig(database_path=...)`
- Parameterized SQL queries with typed inputs
- Extension installation and loading (httpfs, json, spatial, etc.)
- Query results returned as pandas DataFrames via temporary parquet files
- Automatic cleanup of temporary result files on delete

## Design

Unlike Snowflake/BigQuery, DuckDB runs locally with no remote service, so the connector pattern is adapted:
- Queries execute synchronously in `create()` (wrapped in `run_in_executor` for async compat)
- `get()` always returns SUCCEEDED since queries complete in `create()`
- `delete()` cleans up temporary parquet result files
- No credentials, polling, or dashboard links needed

## Test plan

- [x] 25 unit tests covering task creation, config serialization, SQL generation, connector create/get/delete, parameterized queries, extensions, and end-to-end flow
- [x] All tests pass (`pytest plugins/duckdb/tests/ -v`)
- [x] Linting passes (`ruff check plugins/duckdb/`)
- [x] Plugin structure matches Snowflake plugin layout


plugins/duckdb/src/flyteplugins/duckdb/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+```python
+import flyte
+from flyte.io import DataFrame
+from flyteplugins.duckdb import DuckDB, DuckDBConfig
+
+config = DuckDBConfig()
+
+count_rows = DuckDB(
+    name="count_rows",
+    query_template="SELECT COUNT(*) AS total FROM 'data.parquet'",
```
kumare3: Where is the data.parquet coming from?
Is this an input of type `flyte.io.DataFrame`

Then i would love to support
```python
count_rows = DuckDB(
    name="count_rows",
    query_template="SELECT COUNT(*) AS total FROM '{input}'",
    plugin_config=config,
    input=DataFrame,   # This can be implicit
    output_dataframe_type=DataFrame,
)
```

Then you can pass a parquet, a pandasDataframe, a spark DataFrame or anything to it

plugins/duckdb/src/flyteplugins/duckdb/connector.py
```diff
@@ -0,0 +1,150 @@
...
+        result_uri: Path to the temporary parquet file containing query results.
+        has_output: Indicates if the query produces output.
+    """
+
+    query_id: str
+    result_uri: Optional[str] = None
+    has_output: bool = False
+
+
+class DuckDBConnector(AsyncConnector):
```
kumare3: You dont want a connector. connector is useful for connecting to remote services. For this it should just be a Dataframe type or a task plugin

plugins/duckdb/src/flyteplugins/duckdb/connector.py
```diff
@@ -0,0 +1,150 @@
+import asyncio
```
kumare3: delete this entire file

plugins/duckdb/src/flyteplugins/duckdb/task.py
```diff
@@ -0,0 +1,83 @@
...
+        database_path: Path to a DuckDB database file, or ":memory:" for an in-memory database.
+        extensions: Optional list of DuckDB extensions to install and load before executing
+            the query (e.g., ["httpfs", "parquet"]).
+    """
+
+    database_path: str = ":memory:"
+    extensions: Optional[List[str]] = None
+
+
+class DuckDB(AsyncConnectorExecutorMixin, TaskTemplate):
```
kumare3: No need of connector mixin. Rather this should implement the execute method. It should be like ContainerTask (not all of it just the execute function)
https://github.com/flyteorg/flyte-sdk/blob/28396f21fa1024ed76c0b43225d2a3bda3c3963e/src/flyte/extras/_container.py#L235

kumare3: Once you do this, you can delete the connector. You need this way, because you need guaranteed memory and isolation. connector is ok for shared resources, this is usually api calls

kumare3 (commented):
I think you got it a little wrong

plugins/duckdb/tests/test_task.py
```diff
@@ -0,0 +1,325 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Execute: multi-query
+# ---------------------------------------------------------------------------
+
+
+class TestMultiQuery:
+    @pytest.mark.asyncio
+    async def test_create_insert_select(self):
```
kumare3, line 266: Can you have a test with flyte.io.dataframe

andreahlert: Yep. Is that ok?

samhita-alla:
i'll review the PR later this week. thanks for porting it over!

PR #640, opened 2026-02-10
Added support to handle nested Pydantic types

AdilFayyaz:
Summary
                                                      
  - Fix `_get_element_type()` in the type engine to correctly resolve nested and recursive Pydantic types that were previously flattened to str
  - Add $ref resolution so nested models or enums inside containers (List, Dict) are resolved via $defs lookup instead of defaulting to str
  - Rewrite anyOf handling to support $ref variants (e.g. Optional[Inner]), fixing KeyError: 'type' crashes
  - Add recursive array/object branches so nested containers (List[List[int]], Dict[str, Dict[str, int]]) resolve to their correct Python types
  - Handle null JSON schema type, returning NoneType instead of str
  - Add example demonstrating all nested Pydantic type patterns running as parallel Flyte tasks

  Test plan

  - pytest `tests/flyte/type_engine/pydantic/test_nested_structs_in_pydantic.py` : 16 new tests covering - 
    - Nested arrays: List[List[int]] schema resolution + roundtrip, List[List[Inner]] roundtrip
    - Nested dicts: List[Dict[str, int]] schema + roundtrip, Dict[str, Dict[str, int]] schema + roundtrip, List[Dict[str, Inner]] roundtrip, Dict[str, Dict[str, Inner]] roundtrip
    - anyOf with $ref: List[Optional[Inner]] schema resolution (no KeyError), List[Optional[str]] schema resolution, List[Optional[Inner]] roundtrip via guess_python_type
    - null type: {"type": "null"} resolves to NoneType
    - Enums in containers: List[ModelWithEnum] roundtrip, Dict[str, ModelWithEnum] roundtrip
    - Complex combined: model with List[List[Inner]], Dict[str, Inner], List[Dict[str, int]], List[ModelWithEnum], Optional[Inner] roundtrip with data and with None optional

All previous tests pass

### DIFF
--- a/examples/basics/types/pydantic_nested_types.py
+++ b/examples/basics/types/pydantic_nested_types.py
@@ -1,0 +1,123 @@
+import asyncio
+from enum import Enum
+from typing import Dict, List, Optional
+
+from pydantic import BaseModel
+
+import flyte
+
+env = flyte.TaskEnvironment(name="inputs_pydantic_nested_types")
+
+
+class Status(str, Enum):
+    ACTIVE = "active"
+    INACTIVE = "inactive"
+
+
+class Inner(BaseModel):
+    name: str
+    value: int
+
+
+class ModelWithEnum(BaseModel):
+    label: str
+    status: Status
+
+
+@env.task
+async def nested_lists(matrix: List[List[int]]) -> str:
+    return f"Matrix {len(matrix)}x{len(matrix[0])}: {matrix}"
+
+
+@env.task
+async def list_of_dicts(records: List[Dict[str, int]]) -> str:
+    return f"Records ({len(records)} entries): {records}"
+
+
+@env.task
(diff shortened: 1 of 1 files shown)


tests/flyte/type_engine/pydantic/test_nested_structs_in_pydantic.py
```diff
@@ -0,0 +1,311 @@
...
+
+# -- Models --
+
+
+class Status(str, Enum):
+    ACTIVE = "active"
+    INACTIVE = "inactive"
+
+
+class Inner(BaseModel):
```
pingsutw, line 19: For the follow-up, let’s make sure that FlyteFile, FlyteDirectory, and DataFrame also work in the nested model

AdilFayyaz: Yeah will work on that next.

PR #642, opened 2026-02-10
check if app spec is updated on re-deploy

cosmicBboy:
This PR changes the behavior of re-deploying an app where no changes have been made. Instead of erroring out, it will pass through gracefully.

### DIFF
--- a/src/flyte/remote/_app.py
+++ b/src/flyte/remote/_app.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from typing import AsyncIterator, Literal, Mapping, Tuple, cast
 
+import google.protobuf.json_format
 import grpc
 import rich.repr
 from flyteidl2.app import app_definition_pb2, app_payload_pb2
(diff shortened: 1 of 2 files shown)


src/flyte/remote/_app.py
```diff
@@ -268,6 +246,23 @@ async def delete(
...
+    @staticmethod
+    async def _app_specs_are_equal(
+        old_app_spec: app_definition_pb2.Spec, updated_app_spec: app_definition_pb2.Spec
+    ) -> bool:
+        """
+        Compare two app specs and return True if they are the same.
+
+        If the updated_app_spec doesn't have any ingress defined, use the old app spec's ingress.
+        """
+        old_app_spec_json = google.protobuf.json_format.MessageToDict(old_app_spec)
```
kumare3, line 258: i prefer this --- https://github.com/flyteorg/flyte-sdk/blob/07dc8c7e3a582a7b4106ad759f8c84d2f9b65aea/src/flyte/_internal/runtime/reuse.py#L33

This ensures deterministic outputs

cosmicbboy: SerializeToString will not work because we need to handle the "ingress" field, see below

kumare3 (approved):

PR #652, opened 2026-02-11
Add support for cloning an existing image

jeevb:


src/flyte/_internal/imagebuild/docker_builder.py
```diff
@@ -569,10 +569,6 @@ async def build_image(self, image: Image, dry_run: bool = False, wait: bool = Tr
             uri = await self._build_from_dockerfile(image, push=True, wait=wait)
             return ImageBuild(uri=uri, remote_run=None)
 
-        if len(image._layers) == 0:
-            logger.warning("No layers to build, returning the image URI as is.")
-            return ImageBuild(uri=image.uri, remote_run=None)
```
pingsutw, line 574: Why do we need to remove it for the local builder?

pingsutw: won't it trigger rebuild if len(layer) is 0

jeevb: For consistency. Basically, if you `.clone()`, semantically, you should always rebuild I thought. If you don't want to rebuild/push to a new registry, then you shouldn't `.clone()`, right? Am I thinking of this wrongly?

pingsutw: make senses

PR #658, opened 2026-02-12
stateless code sandbox

samhita-alla:
## Add `flyte.sandbox.create()` for containerized execution 

This PR introduces a lightweight sandbox runtime for executing arbitrary code inside an ephemeral stateless container, along with improvements to file handling.

### Sandbox execution API

`flyte.sandbox.create()` runs arbitrary Python code or shell commands inside a one-off Docker container.

The container is built on demand from the declared `packages` / `system_packages`, executed once, and discarded.

#### Auto-IO mode (default): minimal boilerplate

- Triggered when `code` is provided and `auto_io=True`
- Flyte auto-generates an argparse preamble
- Declared inputs become local variables
- Declared scalar outputs are written automatically to `/var/outputs/`
- No manual I/O handling required

```python
sandbox = flyte.sandbox.create(
    name="double",
    code="result = x * 2",
    inputs={"x": int},
    outputs={"result": int},
)

result = await sandbox.run.aio(x=21)  # -> 42
```
#### Verbatim mode: full script control

- Triggered when code is provided and `auto_io=False`
- Script runs exactly as written
- No CLI args injected
- Script handles all I/O manually via `/var/inputs/` and `/var/outputs/`

```python
sandbox = flyte.sandbox.create(
    name="etl",
    code="""
import json, pathlib

data = json.loads(pathlib.Path("/var/inputs/payload").read_text())
pathlib.Path("/var/outputs/total").write_text(
    str(sum(data["values"]))
)
""",
    inputs={"payload": File},
    outputs={"total": int},
    auto_io=False,
)
```

#### Command mode: run arbitrary commands

- Triggered when command is provided
- Executes any binary, shell pipeline, or test runner

```python
sandbox = flyte.sandbox.create(
    name="test-runner",
    command=["/bin/bash", "-c", pytest_cmd],
    arguments=["_", "/var/inputs/solution.py", "/var/inputs/tests.py"],
    inputs={"solution.py": File, "tests.py": File},
    outputs={"exit_code": str},
)
```

### Execution
Call `.run()` on the returned sandbox to build the image and execute.

## Container improvements (`_container.py`)

- Adds support for parsing `File` input from `kwargs`

## Deterministic remote file paths (`_file.py`) - `File.named_remote(name)`
Creates a File reference whose remote path is derived deterministically from name. Unlike `File.new_remote()`, which generates a random path on each call, this method produces the same path for the same name within a task execution, making it safe across retries.

- First attempt uploads to the path
- Retries resolve to the identical location
- Avoids duplicate uploads

If extraction fails, it falls back to the run base directory.

```python
file = File.named_remote("report.csv")

async with file.open("wb") as f:
    await f.write(b"...")
```

## New error type
`InvalidPackageError`: Introduced to signal invalid or unavailable packages during sandbox image build.

### DIFF
--- a/examples/sandbox/code_sandbox.py
+++ b/examples/sandbox/code_sandbox.py
@@ -61,6 +61,7 @@ sum_sandbox = flyte.sandbox.create(
     code="total = sum(range(n + 1)) if conditional else 0",
     inputs={"n": int, "conditional": bool},
     outputs={"total": int},
+    cache="disable",
 )
 
 # Example 2 — code mode (auto-inject): third-party packages (numpy)
@@ -86,6 +87,7 @@ stats_sandbox = flyte.sandbox.create(
     },
     outputs={"mean": float, "std": float, "window_end": datetime.datetime},
     packages=["numpy"],
+    cache="disable",
 )
 
 # Example 3 — verbatim mode: complete Python script, full control
--- a/src/flyte/sandbox/_code_sandbox.py
+++ b/src/flyte/sandbox/_code_sandbox.py
@@ -256,7 +256,13 @@ class _Sandbox:
 
             python_args = " ".join(cli_args)
             python_cmd = f"python $1 {python_args}" if python_args else "python $1"
-            bash_cmd = f"set -o pipefail && {python_cmd}; _exit=$?; echo $_exit > /var/outputs/exit_code"
+            bash_cmd = (
+                "set -o pipefail && "
+                "echo '--- script ---' && cat \"$1\" && "
+                "echo '--- running ---' && "
+                f"{python_cmd}; "
+                "_exit=$?; echo $_exit > /var/outputs/exit_code"
+            )
 
             return ContainerTask(
                 name=task_name,


src/flyte/extras/__init__.py
```diff
@@ -9,14 +9,21 @@
...
 
 from flyte.durable._time import durable_sleep, durable_time
 
 from ._container import ContainerTask
+from ._sandbox import ImageConfig, InvalidPackageError, Sandbox, sandbox_environment
 
 __all__ = [
     "ContainerTask",
+    "ImageConfig",
+    "InvalidPackageError",
```
kumare3: move to flyte.errors

src/flyte/extras/__init__.py
```diff
@@ -9,14 +9,21 @@
...
+    "Sandbox",
```
kumare3: @cosmicBboy lets talk about this and the other thing we have working on

src/flyte/_internal/runtime/convert.py
```diff
@@ -232,14 +237,14 @@ async def convert_outputs_to_native(interface: NativeInterface, outputs: Outputs
...
         return next(iter(kwargs.values()))
     else:
-        # Return as tuple if multiple outputs, using the order from the proto literals list
-        # (which preserves the original serialization order) rather than interface.outputs.keys()
-        # (which may have different order due to protobuf map not preserving order)
-        output_order = [named_literal.name for named_literal in outputs.proto_outputs.literals]
-        return tuple(kwargs[k] for k in output_order)
+        # Return as tuple if multiple outputs are defined in the interface,
+        # to match the order of outputs in the interface
+        return tuple(kwargs[k] for k in interface.outputs.keys())
```
samhita-alla, line 242: @cosmicBboy what prompted you to make this change? `outputs.proto_outputs.literals` doesn’t preserve the order whereas `interface.outputs.keys()` does.

cosmicbboy: when I was making the change to add support for named tuples https://github.com/flyteorg/flyte-sdk/pull/597/changes#diff-55d849b75cb68188430758dc4d7d15239e201b7a6fcdfd8b46bbc9cdb66cc3b1 for some reason `interface.outputs.keys()` was not preserving order. Can we re-test the namedtuple examples with your reversion to make sure namedtuples still works?

- https://github.com/flyteorg/flyte-sdk/blob/main/examples/basics/types/namedtuple_types.py
- https://github.com/flyteorg/flyte-sdk/blob/main/examples/basics/types/namedtuple_types_complex.py

cosmicbboy: I just tested those examples on this branch and it works 👍🏾

PR #664, opened 2026-02-12
Generalize custom type detection in nested Pydantic models

AdilFayyaz:
## Summary:                               

  - When `guess_python_type` reconstructs a Python type from a Flyte LiteralType (JSON schema), nested custom types with registered TypeTransformers were being rebuilt as generic dataclasses instead of being recognized as their original types. This is because the schema reconstruction code had hardcoded checks for File and Dir only and never consulted TypeEngine._REGISTRY for other registered types.
  - Added `schema_match()` method to TypeTransformer that auto-matches for BaseModel subclasses by comparing JSON schema fingerprints (title, type, required fields). No user override needed, registering a transformer for BaseModel subclass is sufficient.
  - Added `_match_registered_type_from_schema()` helper that iterates all registered transformers to find a match.
  - Replaced all hardcoded `File.schema_match()/Dir.schema_match()` call sites in `generate_attribute_list_from_dataclass_json_mixin` and `_get_element_type` with the generalized helper. This also fixes a bug where File/Dir were not detected when referenced via $ref in the schema.

## Test Plan
- pytest `tests/flyte/type_engine/pydantic/test_custom_type_in_nested_pydantic.py` 10 new tests covering `_match_registered_type_from_schema` (match, no match, File, Dir), guess_python_type structure verification (direct, List, Dict, nested List, mixed with File+Dir), and default schema_match behavior

### DIFF
--- a/src/flyte/types/_type_engine.py
+++ b/src/flyte/types/_type_engine.py
@@ -2204,8 +2204,7 @@ def _get_element_type(
     if not isinstance(element_property, dict):
         return typing.Any
 
-    matched_type = _match_registered_type_from_schema(element_property)
-    if matched_type is not None:
+    if (matched_type := _match_registered_type_from_schema(element_property)) is not None:
         return matched_type
 
     # Handle $ref for nested models and enums
@@ -2222,8 +2221,7 @@ def _get_element_type(
             if ref_schema.get("enum"):
                 return str
             # Check if the $ref matches a registered custom type
-            matched_type = _match_registered_type_from_schema(ref_schema)
-            if matched_type is not None:
+            if (matched_type := _match_registered_type_from_schema(ref_schema)) is not None:
                 return matched_type
             # if defs not in the schema, they need to be propogated into the resolved schema
             if "$defs" not in ref_schema and defs:


src/flyte/types/_type_engine.py
```diff
@@ -2200,20 +2197,16 @@ def _get_element_type(
...
     if not isinstance(element_property, dict):
         return typing.Any
 
-    if File.schema_match(element_property):
-        return File
-    elif Dir.schema_match(element_property):
-        return Dir
+    matched_type = _match_registered_type_from_schema(element_property)
+    if matched_type is not None:
+        return matched_type
```
pingsutw: nit:
```suggestion
    if (matched_type := _match_registered_type_from_schema(element_property)) is not None:
        return matched_type
```

src/flyte/types/_type_engine.py
```diff
@@ -2228,6 +2221,10 @@ def _get_element_type(
             # Guard the nested enum elements inside containers
             if ref_schema.get("enum"):
                 return str
+            # Check if the $ref matches a registered custom type
+            matched_type = _match_registered_type_from_schema(ref_schema)
```
pingsutw: ditto

pingsutw:
lgtm, some minor nits

PR #666, opened 2026-02-13
Implement user-facing `HashFunction` 

cosmicBboy:
This pull request introduces advanced examples and core support for content-based caching of DataFrames in Flyte using custom hash functions. It adds three new example scripts demonstrating how to enable cache key computation for pandas, Polars, and Flyte DataFrames, and updates the core `flyte.io.DataFrame` implementation to support hash-based caching via a flexible API.

**Key highlights:**
- New examples show how to use custom hash functions for caching with pandas, Polars, and Flyte DataFrames, including both annotation-based and explicit API usage.
- Core changes to `flyte.io.DataFrame` and related modules add support for passing custom hash functions and storing hash values, enabling cache-aware workflows.

---

### New Examples for Content-Based Caching

* Added `examples/advanced/hash_pandas_dataframe.py`: Demonstrates using `HashFunction` with `typing.Annotated` for pandas DataFrames, showing cache hits by comparing task outputs.
* Added `examples/advanced/hash_polars_dataframe.py`: Demonstrates similar content-based caching for Polars DataFrames using custom hash functions and type annotations.
* Added `examples/advanced/hash_flyte_dataframe.py`: Shows how to use a custom hash function with `flyte.io.DataFrame` via the new `hash_method` parameter in `from_local`, enabling cache key computation for Flyte DataFrames.

### Core API and Implementation Enhancements

* Updated `flyte.io.DataFrame`:
  - Added `hash` field and internal `_hash_method` attribute to store and propagate hash values for cache key computation. [[1]](diffhunk://#diff-05d6fef8567ad079a359de94576b23aeeb68e0adf531dfdef5f68558aac84c3aR80) [[2]](diffhunk://#diff-05d6fef8567ad079a359de94576b23aeeb68e0adf531dfdef5f68558aac84c3aR90)
  - Modified `from_local` and `from_local_sync` methods to accept a `hash_method` parameter, allowing users to specify a `HashMethod` or precomputed hash string for content-based caching. [[1]](diffhunk://#diff-05d6fef8567ad079a359de94576b23aeeb68e0adf531dfdef5f68558aac84c3aR238) [[2]](diffhunk://#diff-05d6fef8567ad079a359de94576b23aeeb68e0adf531dfdef5f68558aac84c3aR250-R291) [[3]](diffhunk://#diff-05d6fef8567ad079a359de94576b23aeeb68e0adf531dfdef5f68558aac84c3aR311-R318) [[4]](diffhunk://#diff-05d6fef8567ad079a359de94576b23aeeb68e0adf531dfdef5f68558aac84c3aR327)
  - Added a module-level `_dataframe_hash_cache` to maintain hashes for raw DataFrames, supporting cache-aware behavior even when type annotations are not present.

* Internal imports and typing improvements:
  - Imported `HashMethod` in `dataframe.py` and adjusted type signatures accordingly.
  - Added `HashFunction` to the public API in `flyte/io/__init__.py`.
  - Minor typing and import improvements for maintainability.

---

#### With `flyte.io.DataFrame`

```python
@env.task
async def produce_dataframe() -> DataFrame:
    """
    Producer task that creates a DataFrame with a custom hash method.

    The hash is computed from the DataFrame content using hash_pandas_dataframe,
    which allows downstream tasks to use content-based caching.
    """
    df = pd.DataFrame(SAMPLE_DATA)

    # Create a HashFunction from our custom hash function
    hash_method = HashFunction.from_fn(hash_pandas_dataframe)

    # Use from_local to create a DataFrame with the custom hash
    return await DataFrame.from_local(df, hash_method=hash_method)

@env.task(cache=Cache(behavior="override", version_override="v1"))
async def consume_dataframe(input_df: DataFrame) -> str:
    """
    Consumer task that processes the DataFrame.

    This task is cached based on the hash of the input DataFrame. If called with
    a DataFrame that has the same hash, the cached result will be returned.
    """
    ...
```

#### With third-party dataframe type e.g. `pd.DataFrame`

```python
from flyte.io import HashFunction

HashedPandasDataFrame = Annotated[pd.DataFrame, HashFunction.from_fn(hash_pandas_dataframe)]

@env.task
async def produce_dataframe() -> HashedPandasDataFrame:
    """
    Producer task that creates a pandas DataFrame.

    The return type annotation includes HashFunction, which tells Flyte to compute
    a content-based hash from the DataFrame using our custom hash function.
    """
    df = pd.DataFrame(SAMPLE_DATA)
    df["action_name"] = flyte.ctx().action.run_name
    return df
```

---

These changes make it much easier to implement cache-aware, content-based workflows with DataFrames in Flyte, supporting both explicit and annotation-based hash specification for pandas, Polars, and Flyte DataFrames.

### DIFF
--- a/plugins/polars/tests/test_df_transformer.py
+++ b/plugins/polars/tests/test_df_transformer.py
@@ -84,18 +84,20 @@ def test_annotate_extraction_lazyframe():
     from flyte.io._dataframe.dataframe import extract_cols_and_format
 
     xyz = typing.Annotated[pl.LazyFrame, "myformat"]
-    a, b, c, d = extract_cols_and_format(xyz)
+    a, b, c, d, e = extract_cols_and_format(xyz)
     assert a is pl.LazyFrame
     assert b is None
     assert c == "myformat"
     assert d is None
+    assert e is None
 
     # Without annotation, extract_cols_and_format returns empty string
-    a, b, c, d = extract_cols_and_format(pl.LazyFrame)
+    a, b, c, d, e = extract_cols_and_format(pl.LazyFrame)
     assert a is pl.LazyFrame
     assert b is None
     assert c == ""  # No format annotation
     assert d is None
+    assert e is None
 
 
 def test_retrieving_encoder():


wild-endeavor (commented):
Already merged but I think we should've kept this simpler.

* I don't think we should change the HashMethod protocol... that took a memoryview for a reason and has certain implications (a no copy scan of bytes).  I don't think we need to change HashMethod protocol cuz I don't think adding a `hash_method` argument the way it's been added is really helpful:  if a user wanted to get a hash out of an object as a whole, there's no need to do the dance with `HashFunction.from_fn(h)` - they can just as easily call `h` directly, and then pass the string value as the final hash_value.  This saves a bit of copy/paste yeah but not too much.  I was thinking that hash method would be more useful if it was a property of the Encoder, whose responsibility it is to actually read the bytes anyway.
* The `_dataframe_hash_cache` is a nice feature yes it's not just a problem faced by dataframes - it's also pertinent to Files/Dirs.  I feel like this should be a type_engine level optimization not a feature on this transformer.

PR #667, opened 2026-02-13
Orchestration sandbox.

kumare3:


src/flyte/sandbox/_api.py
```diff
@@ -0,0 +1,276 @@
...
+        image=image,
+        cache=cache,
+        retries=retries,
+        _user_source=source_code,
+        _user_input_names=input_names,
+        _user_functions=functions,
+    )
+
+
+def orchestrator(
```
cosmicbboy: I think this should be called orchestrator_from_string

src/flyte/sandbox/_api.py
```diff
@@ -0,0 +1,276 @@
...
+    max_memory: int = 50 * 1024 * 1024,
+    max_stack_depth: int = 256,
+    type_check: bool = True,
+    name: Optional[str] = None,
+    cache: CacheRequest = "disable",
+    retries: int = 0,
+) -> Callable[[Callable], SandboxedTaskTemplate]: ...
+
+
+def task(
```
cosmicbboy: I think this should be called orchestrator, to match the naming and semantics of env.sandbox.orchestrator

PR #674, opened 2026-02-15
adding progress bar for asyncio operations

debasisdwivedy:
Hi,

Adding progress bars using `tqdm` package for asynchronous operations


pyproject.toml
```diff
@@ -26,6 +26,7 @@ dependencies = [
     "aiolimiter>=1.2.1",
     "flyteidl2==2.0.2",
     "packaging",
+    "tqdm>=4.67.3",
```
kumare3, line 29: Really prefer if you use rich progress if at all. No new dependencies please

src/flyte/remote/_data.py
```diff
@@ -234,7 +235,7 @@ async def upload_dir(dir_path: Path, verify: bool = True, prefix: str | None = N
         if file.is_file():
             uploaded_files.append(_upload_single_file(cfg, file, verify=verify, basedir=prefix))
 
-    urls = await asyncio.gather(*uploaded_files)
+    urls = await tqdm.gather(*uploaded_files,desc=f"Uploading files to {dir_path}")
```
kumare3, line 238: Rendering this becomes a big challenge - we need to think about the rendering and overhead architecture of runtime vs cli

kumare3:
Thank you, but this is a hard one to take as it introduces new dependencies and not sure if plays well, with runtime and cli

PR #681, opened 2026-02-17
Update CI - use github.ref_name instead of env var

wild-endeavor:
Would have prevented this: 
https://github.com/flyteorg/flyte-sdk/actions/runs/22114974068/job/63920600743

Also `working-directory: ${{ matrix.workdir }}` is not a thing for the main package release job.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -38,6 +38,7 @@ jobs:
         run: |
           uv run python -m twine upload --verbose dist/*
       - name: Sleep until pypi is available
+        if: ${{ github.event_name == 'release' }}
         id: pypiwait
         run: |
           # from 'v1.2.3' get '1.2.3' and make sure it's not an empty string


pingsutw (dismissed):

maint_tools/build_default_image.py
```diff
@@ -47,16 +47,19 @@ async def build_flyte_connector_image(
                 pyproject_file=(Path(__file__).parent.parent / "plugins/bigquery/pyproject.toml"),
                 pre=True,
                 project_install_mode="install_project",
+                extra_args="--no-sources",
```
pingsutw, line 50: To ignore
```toml
[tool.uv.sources]
flyte = { path = "../../", editable = true }
```

pingsutw (dismissed):

maint_tools/build_default_image.py
pingsutw: edidable installed flyte is already in `Image.from_debian_base() `

PR #687, opened 2026-02-18
set workdir to root earlier in image build process

dansola:
When `get_uv_editable_install_mounts` runs, it will add the target mount like so `target={editable_dep.relative_to(project_root)}`. If we set the WORKDIR in the footer, the wrong `project_root` is used as it's assumed to be `/`. This breaks the use or relative paths for editable installs like 
```
[tool.uv.sources]
flyte-env = { path = "../flyte-env", editable = true }
```


pingsutw:
fixed it here https://github.com/flyteorg/flyte-sdk/pull/719

PR #695, opened 2026-02-20
example: interactive sandbox AppEnvironment

cosmicBboy:
An example prototype for interactive sandboxes

### DIFF
--- a/examples/apps/interactive_sandbox_app/fastapi_app.py
+++ b/examples/apps/interactive_sandbox_app/fastapi_app.py
@@ -1,14 +1,11 @@
 """FastAPI app for running commands with seccomp + Landlock sandboxing."""
 
-import os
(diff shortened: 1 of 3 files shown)


examples/apps/interactive_sandbox_app/fastapi_app.py
```diff
@@ -0,0 +1,343 @@
...
+from sandbox import (
+    SAFE_ENV,
+    check_landlock_support,
+    check_seccomp_support,
+    get_sandbox_info,
+    run_command_sandboxed,
+    run_python_sandboxed,
+)
+
+app = FastAPI(
```
kumare3, line 18: interesting idea, but definitely not in python, should be in golang or rust. you dont want it to get affected by random installs etc.

PR #698, opened 2026-02-22
Pytorch elastic hanging in case of cuda oom or non-rank0 exceptions

kumare3:
## Summary

  Fix PyTorch elastic training jobs hanging for up to 90 minutes when a worker hits CUDA OOM by adding NCCL timeout
  configuration and fixing several bugs in the `execute()` lifecycle.

  ### Problem

  When a DDP worker hits CUDA OOM and skips a collective (e.g. `loss.backward()`), the surviving workers block in NCCL
  all-reduce with PyTorch's default timeouts:
  - **Collective timeout**: 600s (10 min) — how long workers wait in a stuck all-reduce
  - **Heartbeat timeout**: 1800s (30 min) — how long before the heartbeat monitor SIGABRTs the process
  - **Restart cycles**: `max_restarts` × (collective + heartbeat) — with defaults: 3 × 2400s = **~2 hours**

  Additionally, `execute()` never called `pre()` or `post()`, dropped `*args` in local mode, and didn't set up the task
  context properly.

  ### Changes

  **`plugins/pytorch/src/flyteplugins/pytorch/task.py`**

  - Add three NCCL timeout knobs to `Elastic` config:
    - `nccl_heartbeat_timeout_sec` (default 300s, PyTorch default is 1800s)
    - `nccl_async_error_handling` (default False, opt-in for faster crash-exit on stuck collectives)
    - `nccl_collective_timeout_sec` (default None/PyTorch default, opt-in to reduce the 600s collective timeout)
  - `pre()` propagates these as env vars (`TORCH_NCCL_HEARTBEAT_TIMEOUT_SEC`, `TORCH_NCCL_ASYNC_ERROR_HANDLING`,
  `FLYTE_NCCL_COLLECTIVE_TIMEOUT_SEC`) — respects existing env vars
  - `launcher_entrypoint()` patches `torch.distributed.constants` AND `torch.distributed.distributed_c10d` to override the
  collective timeout before user code calls `init_process_group()` (both modules must be patched due to Python's `from X
  import Y` binding semantics)
  - Fix `execute()`: call `pre()`/`post()`, set up task context, fix local mode `*args`
  - Catch `ChildFailedError` and re-raise with human-readable error messages (e.g. SIGABRT → "NCCL collective timed out —
  common symptom of CUDA OOM causing ranks to desync")
  - Improve docstrings: document restart cycle cost formula, when to use `max_restarts=0` vs higher values

  **`plugins/pytorch/tests/test_task.py`**

  - 17 new tests (19 total) covering all NCCL timeout knobs, env var propagation, and the `launcher_entrypoint` monkey-patch

  **`examples/plugins/pytorch/cuda_oom_hang.py`** (new)

  - Reproducer that triggers CUDA OOM on rank 1 to desync DDP collectives
  - Demonstrates aggressive timeout settings: `nccl_collective_timeout_sec=60`, `nccl_heartbeat_timeout_sec=60`,
  `nccl_async_error_handling=True`, `max_restarts=0`

  ### Results

  Tested on 2x T4 GPUs with the reproducer:

  | Setting | Time to failure |
  |---------|----------------|
  | Before (PyTorch defaults) | ~90 min hang |
  | After (plugin defaults) | ~15 min |
  | After (aggressive settings) | ~2 min |

### DIFF
--- a/plugins/pytorch/src/flyteplugins/pytorch/task.py
+++ b/plugins/pytorch/src/flyteplugins/pytorch/task.py
@@ -5,7 +5,6 @@ from dataclasses import dataclass, field
 from typing import Any, Dict, Literal, Optional, Union
 
 import flyte
-import flyte.report
 from cloudpickle import cloudpickle
 from flyte._context import internal_ctx
 from flyte._logging import logger
@@ -102,6 +101,10 @@ class Elastic:
             in the collective for this long before the NCCL watchdog fires. This is the
             first phase of failure detection. PyTorch default is 600s (10 min). Set to
             None to use PyTorch default.
+        nccl_enable_monitoring (bool): When True, sets TORCH_NCCL_ENABLE_MONITORING=1
+            to activate NCCL's built-in monitoring thread. The monitoring thread
+            checks each worker's heartbeat counter and sends SIGABRT when it stalls,
+            which is what drives ``nccl_heartbeat_timeout_sec``. Defaults to True.
     """
 
     nnodes: Union[int, str]
@@ -114,6 +117,7 @@ class Elastic:
     nccl_heartbeat_timeout_sec: Optional[int] = 300
     nccl_async_error_handling: bool = False
     nccl_collective_timeout_sec: Optional[int] = None
+    nccl_enable_monitoring: bool = True
 
 
 def launcher_entrypoint(tctx: TaskContext, fn: bytes, kwargs: dict):
@@ -126,7 +130,7 @@ def launcher_entrypoint(tctx: TaskContext, fn: bytes, kwargs: dict):
     )
 
     # Override the default NCCL collective timeout before the user calls
-    # init_process_group().  We must patch both the constants module AND
+    # init_process_group(). We must patch both the constants module and
     # the distributed_c10d module because `from constants import
     # default_pg_nccl_timeout` creates a separate name binding.
     nccl_timeout = os.environ.get("FLYTE_NCCL_COLLECTIVE_TIMEOUT_SEC")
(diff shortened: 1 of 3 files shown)


pmahindrakar-oss (dismissed):

plugins/pytorch/src/flyteplugins/pytorch/task.py
```diff
@@ -119,36 +277,86 @@ async def pre(self, *args: P.args, **kwargs: P.kwargs) -> Dict[str, Any]:
...
+
+        # Propagate the collective timeout to worker subprocesses via env var.
+        # The launcher_entrypoint reads this and overrides the PyTorch default
+        # before user code calls init_process_group().
+        if (
+            self.plugin_config.nccl_collective_timeout_sec is not None
+            and "FLYTE_NCCL_COLLECTIVE_TIMEOUT_SEC" not in os.environ
+        ):
+            os.environ["FLYTE_NCCL_COLLECTIVE_TIMEOUT_SEC"] = str(self.plugin_config.nccl_collective_timeout_sec)
+
```
samhita-alla, line 313: should we add a config param for `TORCH_NCCL_ENABLE_MONITORING`? it works in conjunction with `TORCH_NCCL_HEARTBEAT_TIMEOUT_SEC`.

> If set to 1,enable monitoring thread which aborts the process when the ProcessGroupNCCL Watchdog thread gets stuck and no heartbeat is detected after TORCH_NCCL_HEARTBEAT_TIMEOUT_SEC. This can happen due to calling CUDA/NCCL APIs that may hang. It is Useful to prevent jobs being stuck for a prolonged time than necessary tying up cluster resources.

kumare3: I guess and by default we always enable it right?

samhita-alla: yes

PR #704, opened 2026-02-24
codegen plugin

samhita-alla:
LLM-driven single-file Python code generation plugin

- Adds `flyteplugins-codegen`, a plugin that uses LLMs to generate, test, and iterate on single-file Python scripts inside isolated sandboxes
- Supports two execution backends: LiteLLM mode (any model with structured output support) and Claude Agent SDK mode (autonomous Claude agent)

### AutoCoderAgent - the main entry point:
- Takes a prompt + optional samples/schema/constraints, generates a Python script, writes pytest tests, builds a sandbox image, runs tests, diagnoses failures, and iterates until tests pass
- `result.run()` / `result.as_task()` execute the generated code against real data in an isolated container
- Supports data extraction from DataFrames and Files, auto-generates Pandera schemas, and infers input/output types

```python
import flyte
from flyte.io import File
from flyte.sandbox import sandbox_environment
from flyteplugins.codegen import AutoCoderAgent

agent = AutoCoderAgent(model="gpt-4.1", name="summarize-sales", resources=flyte.Resources(cpu=1, memory="1Gi"))

env = flyte.TaskEnvironment(
    name="my-env",
    secrets=[flyte.Secret(key="openai_key", as_env_var="OPENAI_API_KEY")],
    image=flyte.Image.from_debian_base().with_pip_packages(
        "flyteplugins-codegen",
    ),
    depends_on=[sandbox_environment],  # Required
)

@env.task
async def process_data(csv_file: File) -> tuple[float, int, int]:
    result = await agent.generate.aio(
        prompt="Read the CSV and compute total_revenue, total_units and row_count.",
        samples={"sales": csv_file},
        outputs={"total_revenue": float, "total_units": int, "row_count": int},
    )
    return await result.run.aio()
```

### Agent mode (`backend=claude`):
- Runs an autonomous Claude agent that writes files, runs tests (intercepted into sandboxes), and fixes issues
- Hooks checkpoint intermediate artifacts to named remote storage for durable execution across retries

```python
agent = AutoCoderAgent(
    name="my-task",
    model="claude-sonnet-4-5-20250929",
    backend="claude",  # Requires ANTHROPIC_API_KEY as a Flyte secret
)

result = await agent.generate.aio(
    prompt="...",
    samples={"input": my_file},
    outputs={"result": str},
)
```


plugins/codegen/examples/example_csv_processing.py
```diff
@@ -0,0 +1,140 @@
...
+            addl_layer=PythonWheels(
+                wheel_dir=Path(__file__).parent.parent / "dist",
+                package_name="flyteplugins-codegen",
+                pre=True,
+            ),
+        )
+        .with_apt_packages("git")
+        .with_pip_packages("git+https://github.com/flyteorg/flyte-sdk.git@86f88fece16d956e28667d3f0d8d49108c8cdd68")
+    ),
+    depends_on=[sandbox_environment],
```
kumare3, line 44: this is odd right? we should say AutoCodeAgent.environment no?

plugins/codegen/src/flyteplugins/codegen/execution/agent_sdk.py
```diff
@@ -0,0 +1,677 @@
...
+
+    # Allow safe workspace ops
+    if prog in SAFE_PREFIXES:
+        return "allow"
+
+    # Deny others
+    return "deny"
+
+
+async def code_gen_eval_agent_sdk(
```
kumare3: what does sdk mean here?

kumare3:
I would change this 
```
depends_on=[sandbox_environment],  # Required
```

to 
```
    depends_on=[agent.environment],
```

plugins/codegen/examples/example_csv_processing.py
samhita-alla: this env is just for sandboxes and isn’t related to the plugin, hence the name `sandbox_environment`.

plugins/codegen/src/flyteplugins/codegen/execution/agent_sdk.py
samhita-alla: claude agent SDK

plugins/codegen/examples/example_csv_processing.py
samhita-alla: `depends_on=[agent.environment]` - @kumare3 don’t think this is correct. the env isn't agent-related; it’s meant for sandbox/container tasks.

cosmicbboy: yeah, this should just be `sandbox_environment`, since it's the pre-defined task environment here https://github.com/flyteorg/flyte-sdk/blob/main/src/flyte/sandbox/_code_sandbox.py#L26-L29

plugins/codegen/README.md
```diff
@@ -0,0 +1,427 @@
...
+# Override specific inputs
+total_revenue, total_units, transaction_count = result.run(threshold=0.5)
+```
+
+## Data handling
+
+When you pass `samples`, the plugin automatically:
+
+1. **Converts DataFrames to CSVs** and uploads as `File` objects
+2. **Infers Pandera schemas** — conservative type + nullability checks inferred from the sample data (no value constraints)
```
cosmicbboy, line 244: 😍

plugins/codegen/README.md
```diff
@@ -0,0 +1,427 @@
...
+3. **Applies natural-language constraints** — if `constraints` are provided, each one is parsed by the LLM into a Pandera check (e.g., `"quantity must be positive"` → `pa.Check.gt(0)`) and added to the schema
+4. **Extracts comprehensive context** — column stats, distributions, patterns, sample rows
+5. **Includes everything in the prompt** — the serialized schemas and data context are injected into the LLM prompt so the generated code is aware of exact column types, nullability and validation rules
+
+Pandera is used purely for **prompt enrichment**, not runtime validation. The generated code itself doesn't import Pandera — it just benefits from the LLM knowing the precise data structure. The schemas are also stored on `result.generated_schemas` for inspection.
```
cosmicbboy, line 249: interesting, never considered pandera to be used this way. I think in a future iteration it would be cool to be able to actually run the validation, the user would have to add it as a dependency in the sandbox

samhita-alla: yeah, i agree, we can have it as an opt-in feature.

the idea here is similar to the code mode strategy -- in addition to injecting the schema in natural language, the plugin also generates a pandera schema with the constraints baked in (if the user provides any) and pass it to the LLM, which can then use it for code generation.

PR #707, opened 2026-02-24
Fix(deploy): improve duplicate environment error message to hint at --root-dir

AdilFayyaz:
## Summary:                                                                                                                                           
                                                                                                                                                     
  When a project uses a src/ layout and flyte deploy is run from the project root without `--root-dir src`, the same module gets imported twice under different names (my_module.envs and src.my_module.envs), creating two distinct TaskEnvironment objects with the same name. This produces a cryptic "Duplicate environment name" error with no guidance on the cause.

  This PR:
  1. Improves the error messages in _deploy.py to explain the dual-import root cause and suggest passing `--root-dir src`
  2. Adds a new example `(examples/project_structures/uv_project_src_layout/)` that reproduces the issue

  **To reproduce with the new example:**
  `cd examples/project_structures/uv_project_src_layout/my_project && uv sync`

  `flyte deploy --dry-run --recursive src/my_module`              # fails with improved error
  `flyte deploy --dry-run --recursive --root-dir src src/my_module`  # succeeds

### DIFF
--- a/tests/flyte/test_deploy.py
+++ b/tests/flyte/test_deploy.py
@@ -4,12 +4,16 @@ import inspect
 import sys
 import types
 from dataclasses import replace
-from unittest.mock import Mock, patch
+from unittest.mock import AsyncMock, Mock, patch
 
 import pytest
 
 import flyte
+from flyte._build import ImageBuild
 from flyte._deploy import (
+    DeploymentPlan,
+    _build_image_bg,
+    _build_images,
     _check_duplicate_env,
     _get_documentation_entity,
     _recursive_discover,
@@ -17,6 +21,7 @@ from flyte._deploy import (
     plan_deploy,
 )
 from flyte._docstring import Docstring
+from flyte._internal.imagebuild.image_builder import ImageCache
 from flyte._internal.runtime.types_serde import transform_native_to_typed_interface
 from flyte.models import NativeInterface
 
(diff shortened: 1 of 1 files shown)


examples/project_structures/uv_project_src_layout/my_project/src/my_module/main.py
```diff
@@ -0,0 +1,18 @@
+import logging
+import pathlib
+
+import flyte
+from my_module.workflows.b import process_list
+
+if __name__ == "__main__":
+    # root_dir must point to the directory containing pyproject.toml.
```
pingsutw: Should we point to src/ instead?

AdilFayyaz: Updated, however we were testing this for flyte deploy which is a different execution flow. This is only for testing programmatically.

src/flyte/_deploy.py
```diff
@@ -486,8 +486,17 @@ def _recursive_discover(planned_envs: Dict[str, Environment], env: Environment)
...
     if env.name in planned_envs:
         if planned_envs[env.name] is not env:
-            # Raise error if different TaskEnvironment objects have the same name
-            raise ValueError(f"Duplicate environment name '{env.name}' found")
+            # Two distinct TaskEnvironment objects share the same name. This is
+            # most commonly caused by the same module being imported twice under
+            # different names — for example, when `flyte deploy` is run from the
+            # project root of a src-layout project without --root-dir, causing
+            # `my_module.envs` and `src.my_module.envs` to both be loaded.
+            raise ValueError(
```
kumare3: i think we should check the `__file__` and then say that same module is being imported. I think often times it might just be copy paste error in that case, you just want to say that duplicate names are not allowed

src/flyte/_deploy.py
```diff
@@ -501,13 +537,31 @@ def plan_deploy(*envs: Environment, version: Optional[str] = None) -> List[Deplo
     if envs is None:
         return [DeploymentPlan({})]
     deployment_plans = []
-    visited_envs: Set[str] = set()
+    visited_envs: Dict[str, Environment] = {}
     for env in envs:
         if env.name in visited_envs:
-            raise ValueError(f"Duplicate environment name '{env.name}' found")
+            existing_env = visited_envs[env.name]
```
kumare3: why is this repeated code? they look identical. can we just use one? And if we cannot can we move them to a common function

AdilFayyaz: Thanks for pointing it out. I've updated the code block to call `_check_duplicate_env()` since the logic was the same.

PR #708, opened 2026-02-25
fix: accept enum names as valid CLI inputs in EnumTransformer

pingsutw:
## Summary

- `EnumTransformer.assert_type` was validating string inputs against enum **values** (e.g. `"red"`), but the transformer serializes regular enums by **name** (e.g. `"RED"`) everywhere else (`get_literal_type`, `to_literal`, `to_python_value`)
- This caused `flyte run --param '["RED"]'` to fail with `Value RED is not in Enum <enum 'Color'>` even though `"RED"` is a valid enum name
- Fix `assert_type` to accept both enum names and values for string inputs, consistent with the rest of the transformer

## Test plan

- [x] `pytest tests/flyte/type_engine/test_enum.py` — all 4 tests pass, including two new tests:
  - `test_enum_assert_type_accepts_name`: verifies `"RED"` is accepted for `Color.RED = "red"`
  - `test_enum_in_list_accepts_name`: verifies `["RED", "BLUE"]` works for `List[Color]` (the original failing case)
- [x] Manually verified: `flyte run kevin_examples/test.py test_enum --c '["RED"]'` now succeeds

```python
import enum
import typing

import flyte


class Color(enum.Enum):
    RED = "red"
    GREEN = "green"
    BLUE = "blue"


env = flyte.TaskEnvironment(name="test")


@env.task
def test_enum(c: typing.List[Color]) -> str:
    return f"Color is {c[0].value}"


if __name__ == "__main__":
    flyte.init_from_config(images="docker.io/pingsutw/test:latest")
    result = flyte.run(test_enum, c=[Color.RED])
    print(result.url)
```

### DIFF
--- a/src/flyte/types/_type_engine.py
+++ b/src/flyte/types/_type_engine.py
@@ -1038,8 +1038,13 @@ class EnumTransformer(TypeTransformer[enum.Enum]):
         raise ValueError(f"Enum transformer cannot reverse {literal_type}")
 
     def assert_type(self, t: Type[enum.Enum], v: T):
-        val = v.value if isinstance(v, enum.Enum) else v
-        if val not in [t_item.value for t_item in t]:
+        if isinstance(v, enum.Enum):
+            if not isinstance(v, t):
+                raise TypeTransformerFailedError(f"Value {v} is not in Enum {t}")
+            return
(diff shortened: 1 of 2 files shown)


AdilFayyaz (approved):
LGTM!

PR #714, opened 2026-02-25
fix(anthropic): route async Flyte task execution through task.aio()

AdilFayyaz:
## Summary:                                                                                           
  - FunctionTool.execute() now calls task.aio() for async Flyte tasks instead of calling self.func directly, ensuring nested task invocations go through Flyte's controller properly                  
  - Added test_function_tool_execute_async_flyte_task_uses_aio to verify the aio() code path is taken

  ## Test plan:
  - uv run pytest tests/test_agents.py - all 59 tests pass
  - New test mocks task.aio and asserts it is called with the correct kwargs

### DIFF
--- a/plugins/anthropic/tests/test_agents.py
+++ b/plugins/anthropic/tests/test_agents.py
@@ -450,6 +450,25 @@ async def test_function_tool_execute_flyte_task():
     assert result == 12
 
 
+@pytest.mark.asyncio
+async def test_function_tool_execute_async_flyte_task_uses_aio():
+    """Verify task.aio() is called (not self.func) for async Flyte tasks."""
+    env = flyte.TaskEnvironment("test-exec-aio-path")
+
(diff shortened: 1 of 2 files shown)


plugins/anthropic/tests/test_agents.py
```diff
@@ -450,6 +450,25 @@ def multiply(a: int, b: int) -> int:
     assert result == 12
 
 
+@pytest.mark.asyncio
+async def test_function_tool_execute_async_flyte_task_uses_aio():
+    """Verify task.aio() is called (not self.func) for async Flyte tasks."""
+    env = flyte.TaskEnvironment("test-exec-aio-path")
+
+    @env.task
+    async def add_async(a: int, b: int) -> int:
```
kumare3, line 459: add a sync function test and the same code should just work

PR #719, opened 2026-02-26
fix(uv): copy editable dep pyproject.toml files and compute common source_dir

pingsutw:
## Summary

- In `dependencies_only` mode for `UVProject`, copy only the `pyproject.toml` of each editable dependency (instead of the full directory tree) to keep the build context lean
- Track all copied `pyproject.toml` parent directories and compute their common ancestor via `os.path.commonpath` to set `source_dir` on the `UVProject` proto
- Normalize `dst_path` returned by `copy_files_to_context` using `os.path.normpath` to avoid path comparison issues

## Test plan

- [ ] Build an image using a `UVProject` with `project_install_mode="dependencies_only"` and at least one editable dependency defined in `[tool.uv.sources]`
- [ ] Verify the build context only contains the editable deps' `pyproject.toml` files (not their full source trees)
- [ ] Verify `source_dir` in the proto points to the correct common parent directory
- [ ] Run the updated `albatross` example (`bird_feeder`) end-to-end to confirm it resolves the `seeds` editable dependency correctly


wild-endeavor (approved):
Can we add in a main py

PR #720, opened 2026-02-26
fix: skip ignore files inside standard-ignored directories

pingsutw:
## Summary

- Add `_is_under_standard_ignored_dir()` to `GitIgnore` to detect when a `.gitignore`/`.flyteignore` is nested inside a directory matched by `STANDARD_IGNORE_PATTERNS` (e.g. `.venv`, `__pycache__`, `dist`, `.ruff_cache`)
- Add `.ruff_cache` and `.mypy_cache` to `STANDARD_IGNORE_PATTERNS`

## Test plan

```
flyte -vvv run examples/basics/hello.py main
```

### DIFF
--- a/src/flyte/_code_bundle/_ignore.py
+++ b/src/flyte/_code_bundle/_ignore.py
@@ -4,9 +4,9 @@ import subprocess
 import tarfile as _tarfile
 from abc import ABC, abstractmethod
 from fnmatch import fnmatch
+from functools import cached_property
 from pathlib import Path
 from shutil import which
-from functools import cached_property
 from typing import List, Optional, Type
 
 from flyte._logging import logger


src/flyte/_code_bundle/_ignore.py
```diff
@@ -36,10 +37,22 @@ class GitIgnore(Ignore):
...
         super().__init__(root)
         self.has_git = which("git") is not None
-        self.git_root = self._get_git_root()
-        self.ignore_file_paths = self._find_ignore_files()
-        self.ignored_files = self._list_ignored_files()
-        self.ignored_dirs = self._list_ignored_dirs()
+
+    @cached_property
+    def git_root(self) -> Optional[Path]:
+        return self._get_git_root()
```
pingsutw, line 43: lazy call these function since we don't need it when copy style is load-module

PR #723, opened 2026-02-26
fix(flyte serve): resolve include paths relative to app dir when bundling for flyte serve

AdilFayyaz:
## Summary:                                                                                      
  - When an AppEnvironment specifies include paths that escape the app script's directory (e.g. "../shared_utils.py"), `_deploy_app` was using `app_file.parent` as the tar bundle root. This caused tar entries with `../` components that Python 3.12+ `tarfile.extractall()` rejects, silently crashing the container on startup.
  - Fixed by resolving include paths to absolute paths (anchored at the app script's directory) and using `serialization_context.root_dir` (the project root, autodetected from cwd()) as `from_dir` when building the bundle - ensuring all entries are valid relative paths under a common ancestor.
  - Added an example `(examples/apps/root_dir_bug/)` that reproduces the scenario: an app in a subdirectory including a shared utility from the parent `apps/` directory.

  ## Test plan:
  - `pytest tests/flyte/app/test_app_code_bundling.py` - all 14 tests pass, including:
  - Updated 3 existing tests to assert resolved absolute paths (matching what `build_code_bundle produces` at runtime)
  - New regression test `test_deploy_app_include_outside_app_dir_uses_bundle_root` directly exercises the `../` scenario and asserts both the resolved include path and that `from_dir == root_dir`
  - Deploy `examples/apps/root_dir_bug/app.py` to a live cluster and confirm the container starts and `GET /` returns `{"message": "Hello, World!"}` without a `500`

### DIFF
--- a/tests/flyte/app/test_app_code_bundling.py
+++ b/tests/flyte/app/test_app_code_bundling.py
@@ -328,9 +328,13 @@ async def test_deploy_app_include_only_builds_code_bundle(temp_app_directory):
         mock_build_bundle.assert_called_once()
         call_args = mock_build_bundle.call_args
 
-        # Verify the files passed are the include files
-        assert call_args[0][0] == ("app.py", "utils.py")
-        # Verify the from_dir is the app's parent directory
+        # Include paths are resolved to absolute paths relative to the app dir
+        expected_files = (
+            str((temp_app_directory / "app.py").resolve()),
+            str((temp_app_directory / "utils.py").resolve()),
+        )
+        assert call_args[0][0] == expected_files
+        # from_dir is bundle_root = serialization_context.root_dir
         assert call_args[1]["from_dir"] == temp_app_directory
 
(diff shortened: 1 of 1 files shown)


kumare3:
cc @cosmicBboy

src/flyte/app/_deploy.py
```diff
@@ -88,13 +88,24 @@ async def _deploy_app(
         # Only bundle when not pickling. If this is a pkl bundle, assume that
         # the AppEnvironment has a server function that will be used to serve
         # the app. This function should contain all of the code needed to serve the app.
-        app_file = Path(app._app_filename)
-        app_root_dir = app_file.parent
+        app_dir = Path(app._app_filename).parent
+        bundle_root = serialization_context.root_dir or app_dir
```
kumare3, line 92: nice catch. lgtm, but @cosmicBboy

PR #727, opened 2026-02-27
feat: Add vim keys j,k navigation support for TUI menu

SharpKoi:
## Summary
- Add key bindings `j` and `k` for cursor up and down to `FlyteTUIApp>ActionTreeWidget` and `ExploreTUIApp>RunsTable`.
- Reference: https://github.com/flyteorg/flyte/issues/6945#issue-3994509953

## Test Plan
N/A

### DIFF
--- a/src/flyte/cli/_tui/_explore.py
+++ b/src/flyte/cli/_tui/_explore.py
@@ -176,6 +176,7 @@ class ExploreScreen(Screen):
         table = self.query_one("#runs-table", RunsTable)
         table.cursor_type = "row"
         table.populate()
+        table.focus()
 
     def _repopulate(self) -> None:
         table = self.query_one("#runs-table", RunsTable)


src/flyte/cli/_tui/_app.py
```diff
@@ -87,6 +87,11 @@ class ActionTreeWidget(Tree[str]):
     The first real action becomes the visible top-level node.
     """
 
+    BINDINGS: ClassVar[list[BindingType]] = [
+        Binding("j", "cursor_down", "Down", show=False),
+        Binding("k", "cursor_up", "Up", show=False),
```
machichima:
```suggestion
        Binding("j", "cursor_down", "Cursor Down", show=False),
        Binding("k", "cursor_up", "Cursor Up", show=False),
```

nit: match the built-in description https://github.com/Textualize/textual/blob/741b24b5f34a5c55e931ee09e4d3f8ed0d5440f7/src/textual/widgets/_tree.py#L549-L550

src/flyte/cli/_tui/_explore.py
```diff
@@ -53,6 +53,11 @@ def _fmt_time(ts: float | None) -> str:
 class RunsTable(DataTable):
     """Table of all persisted runs."""
 
+    BINDINGS: ClassVar[list[BindingType]] = [
+        Binding("j", "cursor_down", "Down", show=False),
+        Binding("k", "cursor_up", "Up", show=False),
```
machichima: ditto

machichima:
Seems like in the explore page, the "Status" panel is focused first, so initially the jk does not work. Are we able to fix this to focus on the explore list by default? Also would be great to be able to use jk when choosing the status

https://github.com/user-attachments/assets/4f537a16-f3db-45b7-ba7e-96550d8b5a3f

kumare3:
can we keep the arrow keys too?

machichima (dismissed):
LGTM. Thank you!

SharpKoi:
> Seems like in the explore page, the "Status" panel is focused first, so initially the jk does not work. Are we able to fix this to focus on the explore list by default? Also would be great to be able to use jk when choosing the status

Hi @machichima, thanks for suggestions and approval.
I'm actually working on adding j, k navigation to status menu, which is a little bit tricky because adding key bindings to class `ExploreScreen` or `Select` won't work to the status menu. The status menu is defined by a private class `SelectOverlay`, of which the actions of cursor up and down are limited in the scope. 
Here I propose 3 workarounds but not sure which one is best for our case, and whether it’s worth doing this(access private class, override classes, ...etc) just to support j/k navigation in the status menu:
### 1. Static Key Bindings
```python
from textual.widgets._select import SelectOverlay
class StatusSelectOverlay(SelectOverlay):
    BINDINGS: ClassVar[list[BindingType]] = [
        Binding("down,j", "cursor_down", "Cursor Down", show=False),
        Binding("up,k", "cursor_up", "Cursor Up", show=False),
    ]

class StatusSelect(Select):
    BINDINGS = [
        Binding("enter,space,down,j,up,k", "show_overlay", "Show menu", show=False),
    ]

    def compose(self) -> ComposeResult:
        from textual.widgets._select import SelectCurrent
        """Compose Select with overlay and current value."""
        yield SelectCurrent(self.prompt)
        yield StatusSelectOverlay(type_to_search=self._type_to_search).data_bind(
            compact=Select.compact
        )
```
Consistent with other components who use key bindings, but we need to inherit and override 2 classes.

### 2. Handle Keys in `ExploreScreen`
```python
def on_key(self, event: KeyEvent):
    from textual.widgets._select import SelectOverlay
    
    focused = self.app.focused
    if isinstance(focused, SelectOverlay):
        if event.key == "j":
            focused.action_cursor_down()
        if event.key == "k":
            focused.action_cursor_up()
        event.prevent_default()
```
There's no class needed to be override, but it's not consistent with other components who use key bindings, and adding key handler to parent widget just for selecting status with j,k is not elegant.

### 3. Handle Keys in Custom `Select`
```python
class StatusSelect(Select):
    BINDINGS = [
        Binding("enter,space,up,k,down,j", "show_overlay", "Show menu", show=False),
    ]

    def on_key(self, event: KeyEvent):
        from textual.widgets._select import SelectOverlay
        
        focused = self.app.focused
        if isinstance(focused, SelectOverlay):
            if event.key == "j":
                focused.action_cursor_down()
            if event.key == "k":
                focused.action_cursor_up()
            event.prevent_default()
```
Handle j,k navigation in the scope of the child widget `StatusSelect`, but it's still not consistent with other components using static key bindings, and it still needs inheriting a class.

---

Note that both workarounds 1 and 3 need to replace `Select` with `StatusSelect` in `ExploreScreen.compose()`:
```python
yield StatusSelect(
    [("All", "all"), ("Running", "running"), ("Succeeded", "succeeded"), ("Failed", "failed")],
    value="all",
    id="status-filter",
    allow_blank=False,
    type_to_search=False
)
```
and all the three need to set `type_to_search=False` to avoid key conflicts.

machichima:
@SharpKoi It would be better to not use the internal methods from the library (`textual.widgets._select`) and only use their public APIs to prevent breaking the abstraction boundary. I think we can go with the following method:


```py
from textual.widgets import Select, OptionList
from textual.binding import Binding

class StatusSelect(Select):
    # Define 'j' and 'k' bindings and map them to custom actions
    BINDINGS = [
        Binding("j", "cursor_down", "Cursor Down", show=False),
        Binding("k", "cursor_up", "Cursor Up", show=False),
    ]

    def __init__(self, *args, **kwargs):
        # type_to_search must be set to False, otherwise 'j'/'k' 
        # will be consumed as search inputs instead of key bindings.
        kwargs.setdefault("type_to_search", False)
        super().__init__(*args, **kwargs)

    def action_cursor_down(self) -> None:
        """Find the internal OptionList and trigger its downward movement when 'j' is pressed."""
        if self.expanded:
            self.query_one(OptionList).action_cursor_down()
        else:
            # If the menu is not yet expanded, pressing 'down' 
            # aligns with the native behavior: open the menu.
            self.action_show_overlay()

    def action_cursor_up(self) -> None:
        """Find the internal OptionList and trigger its upward movement when 'k' is pressed."""
        if self.expanded:
            self.query_one(OptionList).action_cursor_up()
```

PR #731, opened 2026-02-27
feat: embed image build run URL in TaskMetadata for clickable image URI

AdilFayyaz:
## Summary
When a task's image is built by the remote image builder, the build happens as a Flyte run. Previously that
  run's identity was discarded after the build completed. This PR propagates it through the deployment pipeline
  so the Union frontend can render the image URI as a clickable link pointing to the build run in the console.

  This PR also upgrades to flyteidl2==2.0.7, which adds a structured common.RunIdentifier image_build_run message containing org, project, domain, and name on the TaskMetadata.

 ## Changes

  - ImageCache gains a build_run_ids field (Dict[str, RunIdentifierData]) mapping environment name → structured
  run identifier. RunIdentifierData is a new Pydantic model mirroring the RunIdentifier proto fields, chosen
  over storing a URL string to match the new IDL and to survive JSON serialization during transport.
  - _build_image_bg extracts org, project, domain, name from remote_run.pb2.action.id.run and returns a
  RunIdentifierData alongside the image URI instead of a URL string.
  - _build_images collects those identifiers into ImageCache.build_run_ids.
  - get_proto_task looks up the RunIdentifierData for the task's environment and stamps it into
  TaskMetadata.image_build_run as a RunIdentifier proto during serialization, making it available to the
  frontend via the stored TaskSpec.
  - No image_build_run is set for locally-built images since there is no remote run.
  - Backward compatible — ImageCache instances serialized before this change deserialize with build_run_ids
  defaulting to {}.

 ## Test plan

  - test_image_cache_build_run_ids_roundtrip — RunIdentifierData survives to_transport → from_transport
  - test_image_cache_build_run_ids_defaults_empty — build_run_ids defaults to {} when not provided
  - test_image_cache_old_serialized_form_still_deserializes — old caches without the field still load cleanly
  - test_get_proto_task_sets_image_build_url — image_build_run is correctly populated in the protobuf when a run
   identifier is present
  - test_get_proto_task_no_image_build_url_without_cache — image_build_run is absent when no image cache is
  present
  - test_get_proto_task_no_image_build_url_when_env_not_in_build_run_urls — image_build_run is absent when the
  env key is missing from build_run_ids
  - test_build_image_bg_captures_remote_run_url — RunIdentifierData is extracted correctly from
  remote_run.pb2.action.id.run
  - test_build_image_bg_no_url_for_local_build — None remote run produces None run identifier
  - test_build_images_stores_build_run_urls_in_cache — run identifier flows end-to-end into
  ImageCache.build_run_ids
  - test_build_images_no_build_run_urls_for_local_build — local build leaves build_run_ids empty

### DIFF
--- a/tests/flyte/internal/runtime/test_task_serde.py
+++ b/tests/flyte/internal/runtime/test_task_serde.py
@@ -450,7 +450,11 @@ def test_get_proto_task_sets_image_build_url():
 
     cache = ImageCache(
         image_lookup={"test_env_build_url": "registry/my-image:sha256abc"},
-        build_run_ids={"test_env_build_url": RunIdentifierData(org="my-org", project="my-project", domain="development", name="abc123")},
+        build_run_ids={
+            "test_env_build_url": RunIdentifierData(
+                org="my-org", project="my-project", domain="development", name="abc123"
+            )
+        },
     )
     context = SerializationContext(
         project="test-project",
--- a/tests/flyte/test_deploy.py
+++ b/tests/flyte/test_deploy.py
@@ -258,7 +258,9 @@ async def test_build_images_stores_build_run_urls_in_cache():
         cache: ImageCache = await _build_images(plan)
 
     assert cache.image_lookup["my-env"] == "registry/my-image:sha256abc"
-    assert cache.build_run_ids["my-env"] == RunIdentifierData(org="my-org", project="my-project", domain="development", name="abc123")
+    assert cache.build_run_ids["my-env"] == RunIdentifierData(
+        org="my-org", project="my-project", domain="development", name="abc123"
+    )
 
 
 @pytest.mark.asyncio


src/flyte/_deploy.py
```diff
@@ -401,13 +404,15 @@ async def _build_images(deployment: DeploymentPlan, image_refs: Dict[str, str] |
...
+        for env_name, image_uri, build_url in final_images:
             status.success(f"Built image for environment {env_name}: {image_uri}")
             image_identifier_map[env_name] = image_uri
+            if build_url is not None:
+                build_run_urls[env_name] = build_url
     else:
         final_images = []
 
-    return ImageCache(image_lookup=image_identifier_map)
+    return ImageCache(image_lookup=image_identifier_map, build_run_urls=build_run_urls)
```
kumare3: woah this is a big change right, but i guess limited to sdk

src/flyte/_deploy.py
```diff
@@ -401,13 +404,15 @@ async def _build_images(deployment: DeploymentPlan, image_refs: Dict[str, str] |
     if images:
         with status.group(f"Building {len(images)} image{'s' if len(images) > 1 else ''}..."):
             final_images = await asyncio.gather(*images)
-        for env_name, image_uri in final_images:
+        for env_name, image_uri, build_url in final_images:
             status.success(f"Built image for environment {env_name}: {image_uri}")
             image_identifier_map[env_name] = image_uri
+            if build_url is not None:
+                build_run_urls[env_name] = build_url
```
kumare3: can we add it to apps too later?

kumare3 (approved):

src/flyte/_deploy.py
AdilFayyaz: we did have to cut a release of flyteidl2 to support this

src/flyte/_deploy.py
AdilFayyaz: Yup

PR #735, opened 2026-03-01
distributed training with eval

samhita-alla:


examples/ml/train_with_eval.py
```diff
@@ -0,0 +1,356 @@
...
+    """
+    print(f"[eval] Triggered at {trigger_time.isoformat()}")
+    print(f"[eval] Training run: {training_run_name}")
+
+    # 1. Look up the training run and check its status
+    training_run = flyte.remote.Run.get(name=training_run_name)
+
+    if training_run.phase.is_terminal:
+        print("[eval] Training run has reached a terminal state — deactivating eval trigger.")
+        flyte.remote.Trigger.update(name="lightning-eval", task_name=f"{eval_env.name}.run_eval", active=False)
```
kumare3: lets not play around with triggers imo

kumare3: we should just create a run, when callbacks are called and some conditions is met like epoch mod 10

PR #738, opened 2026-03-02
feat: add project/domain/created_at/updated_at filters to Run.listall and Action.listall

AdilFayyaz:
## Summary                                                                                                   

  - Added project and domain parameters to `Run.listall()` so callers can query runs across different projects without re-calling flyte.init(). Previously these were hardcoded to the global config.
  - Added `created_at` and `updated_at` time range filter parameters to both `Run.listall()` and `Action.listall()`, backed by a new `TimeFilter` dataclass.
  - Added `TimeFilter` to the `flyte.remote` public namespace so users can import it directly.
  - Wired `--created-after`, `--created-before`, `--updated-after`, `--updated-before` options into flyte get run, using the existing DateTimeType which accepts ISO dates, now, today, and relative expressions like now - 1 day.
  - All new parameters are optional - existing call sites are unaffected.

## Test plan

  - TimeFilter dataclass stores after/before bounds correctly and defaults both to None
  - time_filtering() produces no filters when both bounds are None
  - time_filtering() maps after → GREATER_THAN_OR_EQUAL and before → LESS_THAN proto filter functions
  - Timestamp values are serialized as ISO 8601 UTC strings
  - Run.listall() without project/domain falls back to global config (no behavior change)
  - Run.listall(project=..., domain=...) correctly overrides the ProjectIdentifier in the gRPC request
  - Run.listall(created_at=TimeFilter(...)) and updated_at=TimeFilter(...) appear in request.filters with
  correct field names and functions
  - Time filters compose correctly alongside existing filters (e.g. in_phase)
  - Action.listall() with created_at/updated_at sends correct filters on ListActionsRequest
  - TimeFilter is importable from flyte.remote
  - Manual smoke test against a live environment: Run.listall(project="<project>", updated_at=TimeFilter(after=<yesterday>)) returns expected runs
  - CLI: flyte get run --updated-after "now - 1 day" and flyte get run --created-after 2026-03-01 --in-phase succeeded pass correct filters through to the API

### DIFF
--- a/src/flyte/cli/_get.py
+++ b/src/flyte/cli/_get.py
@@ -69,16 +69,20 @@ def project(cfg: common.CLIConfig, name: str | None = None):
     "--created-after",
     type=_params.DateTimeType(),
     default=None,
-    help="Show runs created at or after this datetime (UTC). Accepts ISO dates, 'now', 'today', or relative expressions like 'now - 1 day'.",
+    help="Show runs created at or after this datetime (UTC). Accepts ISO dates, 'now', 'today', or 'now - 1 day'.",
 )
-@click.option("--created-before", type=_params.DateTimeType(), default=None, help="Show runs created before this datetime (UTC).")
+@click.option(
+    "--created-before", type=_params.DateTimeType(), default=None, help="Show runs created before this datetime (UTC)."
+)
 @click.option(
     "--updated-after",
     type=_params.DateTimeType(),
     default=None,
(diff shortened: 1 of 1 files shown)


src/flyte/remote/_action.py
```diff
@@ -156,6 +156,8 @@ async def listall(
         in_phase: Tuple[ActionPhase | str, ...] | None = None,
         filters: str | None = None,
```
pingsutw: nit: can we also remove `filters` here since we never use it

pingsutw (approved):

PR #739, opened 2026-03-02
fix: Populate creator field when deploying apps

jmonty42:
## Summary

The server requires `spec.creator.principal` (an `EnrichedIdentity` with a `User` or `Application`) when creating apps. The SDK was not setting this field, causing app deployments to fail with:

```
INVALID_ARGUMENT: validation error:
- app.spec.creator.principal: exactly one field is required in oneof [required]
```

## Changes

- **`src/flyte/app/_deploy.py`**: After building the app IDL and before sending the create request, fetch the current authenticated user via `identity_service.UserInfo()` and populate `spec.creator` with an `EnrichedIdentity` containing the user's subject. Wrapped in try/except so deploy continues gracefully if the identity call fails.
- **`tests/flyte/app/test_app_deploy.py`**: New tests verifying creator is set from user info, and that deploy continues with a warning if the identity call fails.

## Test Plan

- 2 new unit tests pass
- All 23 existing app tests pass

### DIFF
--- a/src/flyte/app/_deploy.py
+++ b/src/flyte/app/_deploy.py
@@ -79,9 +79,11 @@ async def _deploy_app(
     """
     Deploy the given app.
     """
-    import flyte.errors
     from flyteidl2.auth import identity_pb2 as auth_identity_pb2
-    from flyteidl2.common import identifier_pb2, identity_pb2 as common_identity_pb2
+    from flyteidl2.common import identifier_pb2
+    from flyteidl2.common import identity_pb2 as common_identity_pb2
+
+    import flyte.errors
     from flyte.app._runtime import translate_app_env_to_idl
     from flyte.remote import App
 
--- a/tests/flyte/app/test_app_deploy.py
+++ b/tests/flyte/app/test_app_deploy.py
@@ -5,7 +5,6 @@ from flyteidl2.app import app_definition_pb2
 from flyteidl2.auth import identity_pb2
 
 from flyte.app._deploy import _deploy_app
-from flyte.app._app_environment import AppEnvironment
 from flyte.models import SerializationContext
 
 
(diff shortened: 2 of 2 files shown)


src/flyte/app/_deploy.py
```diff
@@ -123,6 +127,20 @@ async def _deploy_app(
         if dryrun:
             return app_idl
         ensure_client()
+
+        # Populate creator with the current authenticated user's identity.
+        # The server requires this field on app creation.
+        try:
+            user_info = await get_client().identity_service.UserInfo(auth_identity_pb2.UserInfoRequest())
```
laurabarton, line 134: I'm wondering if maybe we want to use the remote helper here: https://github.com/flyteorg/flyte-sdk/blame/f33af084a7980430e2a935da8f2ba80b2aaa7579/src/flyte/remote/_user.py#L29

Here is an example of it being used: https://github.com/flyteorg/flyte-sdk/blob/f33af084a7980430e2a935da8f2ba80b2aaa7579/examples/apps/run_webhook_fastapi.py#L75

src/flyte/app/_deploy.py
```diff
@@ -123,6 +127,20 @@ async def _deploy_app(
...
+            creator = common_identity_pb2.EnrichedIdentity(
+                user=common_identity_pb2.User(
+                    id=identifier_pb2.UserIdentifier(subject=user_info.subject),
+                ),
+            )
+            app_idl.spec.creator.CopyFrom(creator)
+        except Exception as exc:
+            logger.warning(f"Could not fetch user identity for app creator field: {exc}")
```
laurabarton, line 142: Do we want to continue if UserInfo fails? Since the field is required, if we continue, it'll fail in the same way as it is now. That is probably fine, but maybe we want to provide a more descriptive error to the user that says something about not being able to get Identity information.

jmonty42: I'm going back and forth with Claude on this. The intention was that if the user identity wasn't found then it would behave as it originally did. However, after seeing Nathan's failures with the e2e tests, Claude is suggesting that this shouldn't be done in the flyte-sdk but rather in the go services. I'm closing this PR.

PR #740, opened 2026-03-02
add raw data path to apps

cosmicBboy:
This pull request introduces support for configuring and accessing the raw data path in Flyte apps, enabling applications to read the raw data path from the app context. The changes allow the raw data path to be set via command-line options, environment variables, or programmatically, and ensure it is correctly propagated in local and remote serving scenarios. An example FastAPI app is also added to demonstrate this new feature.

**Raw Data Path Support**

* Added a new `raw_data_path` attribute to the `AppContext` dataclass and exposed it via the `ctx()` function, allowing apps to access the raw data path. Also implemented the `set_raw_data_path()` function to set this value and propagate it via environment variables for thread safety. (`src/flyte/app/_context.py`)
* Updated the serving context (`_Serve`) and its factory function (`with_servecontext`) to accept and propagate the `raw_data_path` parameter, defaulting to `/tmp/flyte/raw_data` for local mode. The serving logic ensures the raw data path is set in the correct thread context for request handlers. (`src/flyte/_serve.py`)* Extended the CLI (`serve.py`) to accept a `--raw-data-path` option and set the raw data path in the app context when provided. (`src/flyte/_bin/serve.py`)
* Ensured that the container command for remote serving includes a template variable for the raw data path, allowing the backend to substitute the correct value at runtime. (`src/flyte/app/_app_environment.py`)

**Documentation and Example**

* Added an example FastAPI app (`app_with_raw_data_path.py`) that demonstrates accessing the raw data path from the app context and returning it in an API response. (`examples/apps/app_with_raw_data_path.py`)

### DIFF
--- a/src/flyte/_bin/serve.py
+++ b/src/flyte/_bin/serve.py
@@ -206,8 +206,15 @@ def _bind_parameters(
 async def _serve(
     app_env: AppEnvironment,
     materialized_parameters: dict[str, str | flyte.io.File | flyte.io.Dir],
+    raw_data_path: str | None = None,
 ):
     import signal
+
+    if raw_data_path:
+        from flyte.app._context import set_raw_data_path
+
+        set_raw_data_path(raw_data_path)
+        logger.info(f"Set raw_data_path in AppContext: {raw_data_path}")
 
     logger.info("Running app via server function")
     assert app_env._server is not None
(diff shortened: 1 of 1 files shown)


src/flyte/_bin/serve.py
```diff
@@ -328,6 +330,13 @@ def main(
 
     os.environ[RUNTIME_PARAMETERS_FILE] = parameters_file
 
+    logger.info(f"RAW DATA PATH: {raw_data_path}")
+    if raw_data_path:
+        from flyte.app._context import set_raw_data_path
+
+        set_raw_data_path(raw_data_path)
```
pingsutw, line 349: Instead of updating the env var in set_raw_data_path, would it be better to just pass raw_data_path to _serve()?

cosmicbboy: done, can you re-+1?

src/flyte/_serve.py
```diff
@@ -478,6 +486,12 @@ def _serve_local_with_server_func(
         local_app = _LocalApp(app_env=app_env, _serve_obj=self, host=host, port=port)
 
         def _run():
+            # Contextvars don't propagate to new threads. Set raw_data_path in this
+            # thread's context so ctx().raw_data_path works in request handlers.
+            from flyte.app._context import set_raw_data_path
+
+            set_raw_data_path(self._raw_data_path or "")
```
pingsutw, line 493: How does it work if we run two apps locally? Both of them will override the same env vars, right?

cosmicbboy: yep, but it should be the same raw data path in `/tmp`

PR #741, opened 2026-03-02
feat: add with_code_bundle to Image

pingsutw:
```python
from dep import foo

import flyte

env = flyte.TaskEnvironment(
    name="full_build",
    # with_code_bundle() automatically copies source code from root_dir into the image
    # when copy_style="none" is set in with_runcontext() or flyte deploy. When copy_style
    # is "loaded_modules" or "all", it's a no-op since the code is bundled separately.
    image=flyte.Image.from_debian_base().with_code_bundle(),
)


@env.task
def square(x) -> int:
    return x ** foo()


@env.task
def main(n: int) -> list[int]:
    return list(flyte.map(square, range(n)))


if __name__ == "__main__":
    flyte.init_from_config()
    run = flyte.with_runcontext(copy_style="none", version="x").run(main, n=10)
    print(run.url)
```

## Summary
- Add `Image.with_code_bundle()` method that configures automatic source code baking into Docker images when `copy_style="none"` is set in `with_runcontext()` or `flyte deploy`
- When `copy_style` is `"loaded_modules"` or `"all"`, the layer is a no-op since code is bundled separately
- Supports two copy styles: `"loaded_modules"` (only imported Python modules) and `"all"` (entire root_dir)
- Adds `resolve_code_bundle_layer()` integration in `_run_remote()`, `_run_hybrid()`, and `deploy.apply()` before image builds
- Includes safety-net `RuntimeError` in `docker_builder` if a `CodeBundleLayer` reaches the build step unresolved
- Updates the `full_build` example to use the new simplified API

## Test plan
- [x] `test_with_code_bundle_defaults` — creates CodeBundleLayer with correct defaults
- [x] `test_with_code_bundle_all` — stores `"all"` copy style
- [x] `test_with_code_bundle_custom_dst` — stores custom destination
- [x] `test_resolve_code_bundle_no_layers` — returns same object when no CodeBundleLayer present
- [x] `test_resolve_code_bundle_strips_when_not_none` — strips layers when copy_style != "none"
- [x] `test_resolve_code_bundle_strips_when_all` — strips layers when copy_style is "all"
- [x] `test_resolve_code_bundle_hash_stability` — stripped image URI matches base image URI
- [x] `test_resolve_code_bundle_all_copy_style_none` — replaces with CopyConfig for "all" + "none"
- [x] `test_resolve_code_bundle_loaded_modules_copy_style_none` — creates staging dir for "loaded_modules" + "none"
- [x] All 37 existing tests in `test_image.py` pass (no regressions)

### DIFF
--- a/src/flyte/_code_bundle/_utils.py
+++ b/src/flyte/_code_bundle/_utils.py
@@ -323,7 +323,7 @@ def add_imported_modules_from_source(source_path: str, destination: str, modules
 
 def copy_code_bundle_to_context(
     root_dir: pathlib.Path,
-    copy_style: str,
+    copy_style: CopyFiles,
     context_path: pathlib.Path,
     ignore_patterns: Optional[List[str]] = None,
 ) -> pathlib.Path:


src/flyte/_internal/imagebuild/utils.py
```diff
@@ -57,6 +57,38 @@ def copy_files_to_context(src: Path, context_path: Path, ignore_patterns: list[s
     return dst_path
 
 
+def copy_code_bundle_to_context(
```
kumare3: I think we should move this code to the code bundling place. we can call it - copy_code_bundle_files_only. This way all code bundle code will be in the same place

kumare3:
Why do we need this? Looks good, just we should keep all code bundle code with code bundle module

PR #747, opened 2026-03-04
fix: accept enum names as valid CLI inputs in EnumParamType

AdilFayyaz:
## Summary                                                                                                                                          

  - PR #618 established the convention of using enum names (e.g. RED) instead of values throughout EnumTransformer (`get_literal_type`, `to_literal, to_python_value`), and added `examples/basics/enums.py`
  - PR #708 extended this to assert_type so validation also accepted names
  - EnumParamType in `cli/_params.py` was missed - it was still building click choices from enum values (e.g. red-value, green-value), causing `flyte run --color=GREEN` to fail immediately at argument parsing before reaching the transformer
  - Fix: use `e.name` for choices and `enum_type[name]` (bracket lookup by name) in convert

 ## Test command
  `uv run flyte run examples/basics/enums.py standalone_enum_echo --color=GREEN --size=SMALL`

### DIFF
--- a/tests/cli/test_params.py
+++ b/tests/cli/test_params.py
@@ -1,5 +1,4 @@
 import enum
-from enum import StrEnum
 
 import click
 import pytest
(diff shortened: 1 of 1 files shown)


pingsutw (dismissed):
lgtm, thanks, can we add a unit test here https://github.com/flyteorg/flyte-sdk/tree/main/tests/cli

kumare3 (dismissed):

PR #750, opened 2026-03-05
feat: add support for `AWS_CONFIG_FILE` env vars

ljstrnadiii:


src/flyte/storage/_config.py
```diff
@@ -110,6 +123,24 @@ def for_sandbox(cls) -> S3:
         }
         return S3(**final_kwargs)
 
+    def _build_s3_credential_provider_from_config_file(
+        self,
+        aws_profile: str,
+        aws_config_file: str,
+        region: str | None,
+    ) -> typing.Any:
+        import boto3
```
kumare3, line 134: :(, dont want botocore at all, but i guess this is completely user specific, can you please document it in the docs

PR #770, opened 2026-03-09
Add update secret cli command

katrogan:
testing:

```
flyte create secret katrina-test
Enter secret value: 

...

flyte update secret katrina-test
Enter secret value:
[16:59:43] Waiting for secret katrina-test to be fully deleted...  
Secret katrina-test updated successfully!
Note: Secret replication may take a few moments to propagate across the cluster.
```


katrogan:
This waits to check for deletion propagation according to what the cluster reports

PR #772, opened 2026-03-10
add --debug flag and debug run context

cosmicBboy:
This PR exposes a `flyte run --debug ...` option and `flyte.with_runcontext(debug=True)` so that users can use the vscode debugger directly from the CLI/python.

### DIFF
--- a/tests/flyte/test_debug.py
+++ b/tests/flyte/test_debug.py
@@ -6,9 +6,9 @@ from types import SimpleNamespace
 from unittest.mock import MagicMock, patch
 
 from click.testing import CliRunner
+from flyteidl2.core.execution_pb2 import TaskLog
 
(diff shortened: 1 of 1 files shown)


src/flyte/_debug/client.py
```diff
@@ -0,0 +1,67 @@
...
+    ``message`` equals ``"Vscode server is ready"``, indicating that the
+    code-server is actually accepting connections.
+    """
+    for attempt in details.pb2.attempts:
+        server_ready = any(ev.message == VSCODE_READY_MESSAGE for ev in attempt.cluster_events)
+        if not server_ready:
+            continue
+        for log in attempt.log_info:
+            if log.name == VSCODE_DEBUGGER_LOG_NAME:
+                return log.uri
```
pingsutw, line 29: We just need to check the ready flag. (`log.ready`)

PR #790, opened 2026-03-13
feat: consolidate image and bundle caches into LocalDB

pingsutw:
## Summary
- Adds a persistent SQLite disk cache for **image existence checks** and **code bundle uploads**, consolidated into the shared `LocalDB` database (`~/.flyte/local-cache/cache.db`)
- Image cache: after the first successful remote registry check (~4s for `docker manifest inspect`), subsequent runs hit the local SQLite cache (~0ms). Cache key is a SHA-256 of repository, tag, and architecture.
- Bundle cache: if `list_files_to_bundle` returns the same digest as a previous run, the upload is skipped entirely and the cached remote path is reused. The cache check happens before `create_bundle` so tar creation is also avoided on cache hit.
- Both caches have a **1-day TTL** with probabilistic pruning (~5% of reads) to avoid stale entries without adding overhead on every access.
- `PersistentCacheImageChecker` is inserted as the first checker in the `DockerImageBuilder` checker chain, before network-based checkers.

### Key changes
- **`_persistence/_db.py`**: Added `image_cache` and `bundle_cache` table DDLs to `LocalDB`, with a unified `_ALL_TABLE_DDLS` list for DRY initialization across sync and async paths
- **`image_builder.py`**: Removed standalone `~/.flyte/cache/images.db` — reads/writes now go through `LocalDB.get_sync()` with `_write_lock` for thread safety
- **`bundle.py`**: Removed standalone `~/.flyte/cache/bundles.db` — same consolidation pattern; uses lazy import of `LocalDB` to avoid circular imports
- **`_task_cache.py`**: Added missing `_write_lock` to `_set_sync` and `clear` for consistency with all other sync write paths

## Test plan
- [x] `test_cached` — verifies persistent cache is checked first and other checkers are skipped
- [x] `test_persistent_cache_write_and_read` — verifies write/read round-trip and arch isolation via SQLite
- [x] All existing image builder tests pass
- [x] Manual test: `LOG_LEVEL=10 python examples/basics/hello.py` — first run writes cache, second run shows "found in persistent cache" in ~0ms


src/flyte/_internal/imagebuild/image_builder.py
```diff
@@ -170,7 +209,10 @@ async def image_exists(image: Image) -> Optional[str]:
                 image_uri = await checker.image_exists(repository, tag, tuple(image.platform))
                 if image_uri:
                     logger.debug(f"Image {image_uri} in registry")
-                return image_uri
+                    # Persist to disk so future process invocations skip network checks
+                    if checker is not PersistentCacheImageChecker:
```
pingsutw, line 257: One of the problem is the persistent cache never invalidates. If an image is deleted from the registry, the cache will still say it exists, and the build will be skipped

pingsutw: but it makes UX way more better

kumare3: We can solve this, by just having a very short TTL on the cache. this is why i am suggesting using sqlite. Anyways the data is tiny and one row is enough?

pingsutw: Short TTL sounds good to me. I'll update it to use sqlite

pingsutw: Done. Add TTL and cache for code bundle.

src/flyte/_code_bundle/bundle.py
```diff
@@ -28,6 +31,48 @@
...
 _tar_file_extension = ".tar.gz"
 
+_BUNDLE_CACHE_TTL_DAYS = 30
+
+
+def _read_bundle_cache(digest: str) -> tuple[str, str] | None:
+    """Look up a previously uploaded bundle by its file digest. Returns (hash_digest, remote_path) or None."""
+    from flyte._persistence._db import LocalDB
+
+    try:
```
kumare3, line 50: we should not crash if database not found

src/flyte/_code_bundle/bundle.py
```diff
@@ -163,6 +206,16 @@ async def build_code_bundle(
     if logger.getEffectiveLevel() <= logging.INFO:
         print_ls_tree(from_dir, files)
 
+    # Check persistent cache before creating the tar bundle to avoid unnecessary work
+    if not dryrun:
```
kumare3: can you run the stress/runs_per_second to ensure it does not break

kumare3: i have a cache in _run.py. should we just add it there instead of here? it feels weird that this is using sqlite?

kumare3: I guess we will have to then repeat it for deploy and serve i guess

kumare3 (dismissed):

kumare3 (approved):

src/flyte/_code_bundle/bundle.py
```diff
@@ -28,6 +32,56 @@
 _pickled_file_extension = ".pkl.gz"
 _tar_file_extension = ".tar.gz"
 
+_BUNDLE_CACHE_TTL_DAYS = 1
```
AdilFayyaz, line 35: Should this be 30? According to the PR description?

pingsutw: we end up deciding to set it to 1. updated PR description

AdilFayyaz: in that case this is still 30: `_IMAGE_CACHE_TTL_DAYS = 30`

pingsutw: updated it, thanks

AdilFayyaz (approved):

PR #795, opened 2026-03-14
monorepo examples and editable remote building fix

dansola:
guide for monorepo styles in flyte v2 with uv.

guide is at `examples/uv_monorepo_guide/GUIDE.md`

also, a fix to the remote builder such that the whole editable dependency is copied rather than just the toml.
- example before the fix on `02_sibling_packages`: https://demo.hosted.unionai.cloud/v2/domain/development/project/daniel/runs/rv7gs4mcfjkz9gjvcl6j?tab=logs
- example after the fix on `02_sibling_packages`: https://demo.hosted.unionai.cloud/v2/domain/development/project/daniel/runs/rsvvc9qbgsf6blkf9xvr?tab=logs


kumare3:
Is this not similar to the various deploy patterns, packaging patterns etc? +1, but just the number of options is increasing, it might be good to consolidate

PR #798, opened 2026-03-16
Skip image existence check for default flyte release images

pingsutw:
## Summary
- Add `_is_flyte_default` flag to `Image` to identify default flyte release images
- Skip image existence check for default flyte images since they are built and pushed as part of flytekit releases
- Override `_is_flyte_default` to `False` in `build_default_image.py` so the build tool still builds the image
- Minor cleanup: use `not install_flyte` instead of `install_flyte is False`

## Test plan
- [x] `test_released_version_returns_prebuilt_image` — verifies released versions produce correct URI
- [x] `test_released_version_with_v_prefix` — verifies no double `vv` prefix
- [x] `test_install_flyte_false_builds_full_image` — verifies full image build when flyte not installed

### DIFF
--- a/plugins/hitl/src/flyteplugins/hitl/_event.py
+++ b/plugins/hitl/src/flyteplugins/hitl/_event.py
@@ -346,6 +346,7 @@ async def show_form(form_url: str, api_url: str, curl_body: str, type_name: str)
     await flyte.report.flush.aio()
     return html_report
 
+
 @flyte.trace
 async def wait_for_input_event(
     name: str,


src/flyte/_internal/imagebuild/image_builder.py
```diff
@@ -145,6 +145,10 @@ class ImageBuildEngine:
     @staticmethod
     @alru_cache
     async def image_exists(image: Image) -> Optional[str]:
+        if image._is_flyte_default:
+            # For flyte default images, we skip the existence check because
+            # they are built and pushed as part of flytekit releases.
+            return image.uri
```
pmahindrakar-oss, line 151: Can we do lightweight check to know whether that flyte default image exists  to make sure someone is not pulling a broken release?

pingsutw: it should always exists otherwise CI will fail to publish the release

wild-endeavor: yeah there's no real lightweight version of a check.  if we check then it's a full check, which we should skip in certain cases.

PR #799, opened 2026-03-16
 feat: respect .dockerignore in image cache key computation

AdilFayyaz:
## Motivation

CopyConfig and UVProject were hashing all files in a directory regardless of .dockerignore, so changes to excluded files (e.g. build artifacts, data files) incorrectly invalidated the image cache. Additionally, DockerIgnore.update_hash was hashing the file path instead of its content, so editing .dockerignore had no effect on the cache key.                                                                              
                            
## Summary

  - Fixed DockerIgnore.update_hash to hash file content instead of the path string
  - Added DockerfileIgnore(Ignore) to _code_bundle/_ignore.py — wraps PatternMatcher (the same matcher used by the actual Docker build context) and fits into the existing Ignore/IgnoreGroup hierarchy                                                          
  - Refactored update_hasher_for_source to accept ignore: _IgnoreLike (Protocol) instead of a raw Callable, adds directory pruning, and passes absolute paths consistent with list_all_files                                                               
  - Image._get_hash_digest() now resolves the dockerignore once (last DockerIgnore layer wins, falls back to                    
  root_dir/.dockerignore from init config) and threads it through all layer.update_hash(hasher, ignore=ignore) calls — so .dockerignore at the project root is correctly applied even when src is a subdirectory    
                                                                                                                                                            
  Test Plan                 
                                                                                                                                                            
  - uv run pytest tests/flyte/test_image.py -v — includes 7 new tests:                                                                                      
  - test_dockerignore_hash_changes_with_content: changing .dockerignore content produces a different hash
  - test_dockerignore_hash_stable_for_same_content: same content produces a stable hash                                                                   
  - test_image_uri_changes_when_dockerignore_content_changes: image URI (cache key) changes when .dockerignore changes                                    
  - test_copy_config_update_hash_respects_dockerignore: files matched by .dockerignore are excluded from the CopyConfig hash                              
  - test_copy_config_dockerignore_itself_is_hashed: changing .dockerignore content changes the CopyConfig hash                                            
  - test_uv_project_install_project_respects_dockerignore: same coverage for UVProject install_project mode 
  - test_copy_config_respects_dockerignore_from_image_layer — .dockerignore at project root (via with_dockerignore()) is applied to CopyConfig even when src is a subdirectory                                               
  - uv run pytest tests/flyte/code_bundle/test_ignore.py -v — regression check on the Ignore hierarchy

### DIFF
--- a/src/flyte/_image.py
+++ b/src/flyte/_image.py
@@ -8,7 +8,7 @@ from abc import abstractmethod
 from dataclasses import dataclass, field
 from functools import cached_property
 from pathlib import Path
-from typing import TYPE_CHECKING, ClassVar, Dict, List, Literal, Optional, Tuple, TypeVar, Union
+from typing import TYPE_CHECKING, Any, ClassVar, Dict, List, Literal, Optional, Tuple, TypeVar, Union
 
 import rich.repr
 
@@ -74,11 +74,12 @@ class Layer:
                         )
 
     @abstractmethod
-    def update_hash(self, hasher: hashlib._Hash):
+    def update_hash(self, hasher: hashlib._Hash, ignore: Optional[Any] = None):
         """
         This method should be implemented by subclasses to provide a hash representation of the layer.
 
         :param hasher: The hash object to update with the layer's data.
+        :param ignore: Optional ignore instance threaded from Image._get_hash_digest().
         """
 
     def validate(self):
@@ -112,7 +113,7 @@ class PipOption:
             pip_install_args.append(self.extra_args)
         return pip_install_args
 
-    def update_hash(self, hasher: hashlib._Hash):
+    def update_hash(self, hasher: hashlib._Hash, ignore: Optional[Any] = None):
         """
         Update the hash with the PipOption
         """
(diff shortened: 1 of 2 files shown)


src/flyte/_image.py
```diff
@@ -363,9 +371,11 @@ def validate(self):
             raise ValueError(f"Source file {self.src.absolute()} is not a file")
 
     def update_hash(self, hasher: hashlib._Hash):
+        from ._code_bundle._ignore import DockerfileIgnore
         from ._utils import update_hasher_for_source
 
-        update_hasher_for_source(self.src, hasher)
+        ignore = DockerfileIgnore(self.src) if self.src.is_dir() else None
```
pingsutw: The ignore file isn’t always in the src folder, right? It’s either in <root_dir>/.dockerignore or in the directory specified by `with_dockerignore`(<PATH>)

AdilFayyaz: Good catch, I'll fix the implementation. This will require a bit of architectural changes on update_hash which will have to pass through all layers.

kumare3 (approved):

src/flyte/_code_bundle/_ignore.py
```diff
@@ -212,6 +212,41 @@ def _is_ignored(self, path: pathlib.Path) -> bool:
         return False
 
 
+class DockerfileIgnore(Ignore):
+    """Ignores files matching patterns in a .dockerignore file in the given root directory."""
+
+    def __init__(self, root: Path):
+        super().__init__(root)
```
pingsutw, line 219: nit: should we move this class to `_image.py` since it's not related to code bundle

AdilFayyaz: Yeah it's not related to code bundle but I only kept it here since the rest of the Ignore classes were only found in `_code_bundle/_ignore.py` 
Let me know what would be cleaner and I can then update this

PR #805, opened 2026-03-17
feat(🧱 databricks): update the databricks plugin to use the new SDK

jdoiro3:
WIP

### DIFF
--- a/plugins/databricks/src/flyteplugins/databricks/connector.py
+++ b/plugins/databricks/src/flyteplugins/databricks/connector.py
@@ -1,17 +1,18 @@
 import os
-from dataclasses import dataclass, asdict
 import typing as ty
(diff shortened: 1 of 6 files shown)


plugins/databricks/src/flyteplugins/databricks/connector.py
```diff
@@ -1,77 +1,93 @@
...
+
+
+@dataclass(kw_only=True)
+class RunResource(ResourceMeta, jobs.Run):
     databricks_instance: str
-    run_id: str
 
 
-def _get_databricks_job_spec(task_template: TaskTemplate) -> dict:
+@cache_with_sig
```
jdoiro3, line 42: Probably don't need to cache this, which would mean `cache_with_sig` could be removed.

PR #806, opened 2026-03-17
fix: prevent pre-release dependency resolution and sdist conflict in image builds

AdilFayyaz:
## Motivation      
                                                                                               
  Installing a pre-release flyte with --pre caused uv to resolve pre-release versions of all transitive dependencies (e.g. grpcio beta). Additionally, uv build produces both a .whl and .tar.gz, and copying both to the build context caused uv to error with "conflicting URLs".   
                  
  ## Summary                                                                                      
   
  - Remove pre=True from with_pip_packages when installing a pre-release flyte from PyPI — the exact version pin (flyte==x.y.za1) is sufficient without --pre
  - Remove pre=True from with_local_v2() — --pre was being applied to the dependency install step, pulling in pre-release grpcio                                                          
  - Filter sdist files (*.tar.gz, *.tar.bz2, *.zip) when copying the wheel directory to the remote build context, so only .whl is included                                               
                  
  ## Test Plan                                                                                    
                  
  - Run rm -rf dist && uv build then uv run python examples/basics/hello.py          
  - Confirm remote image build completes without "conflicting URLs" error

### DIFF
--- a/src/flyte/_image.py
+++ b/src/flyte/_image.py
@@ -531,7 +531,6 @@ class Image:
     ) -> Image:
         # Would love a way to move this outside of this class (but still needs to be accessible via Image.auto())
         # this default image definition may need to be updated once there is a released pypi version
-        from packaging.version import Version
 
         from flyte._version import __version__
 


src/flyte/_image.py
```diff
@@ -1176,7 +1172,7 @@ def with_local_v2(self) -> Image:
         """
         # Manually declare the PythonWheel so we can set the hashing
         # used to compute the identifier. Can remove if we ever decide to expose the lambda in with_ commands
-        with_dist = self.clone(addl_layer=PythonWheels(wheel_dir=DIST_FOLDER, package_name="flyte", pre=True))
+        with_dist = self.clone(addl_layer=PythonWheels(wheel_dir=DIST_FOLDER, package_name="flyte"))
```
pingsutw, line 1175: ah, it's here, thanks!

PR #817, opened 2026-03-19
fix HITL plugin

cosmicBboy:


dansola (dismissed):

plugins/hitl/src/flyteplugins/hitl/_event.py
```diff
@@ -360,20 +364,21 @@ async def wait_for_input_event(
...
+                async for chunk in storage.get_stream(response_path):
+                    response = json.loads(chunk.decode())
+                    if response.get("status") == "completed":
+                        value = response["value"]
+                        logger.info(f"Event '{name}' received human input: {value}")
+                        print(f"\nReceived human input for '{name}': {value}")
+                        return value
+
+                logger.info(f"Event '{name}' waiting for human input... ({elapsed}/{timeout_seconds}s elapsed)")
+            await flyte.durable.sleep.aio(poll_interval_seconds)
```
kumare3: you dont need durable sleep, this is just an interval right

cosmicbboy: just use asyncio.sleep?

src/flyte/_image.py
```diff
@@ -1195,11 +1195,37 @@ def with_local_v2(self) -> Image:
...
-        :return: Image
+        Returns:
+            Image
         """
         # Manually declare the PythonWheel so we can set the hashing
         # used to compute the identifier. Can remove if we ever decide to expose the lambda in with_ commands
         with_dist = self.clone(addl_layer=PythonWheels(wheel_dir=DIST_FOLDER, package_name="flyte"))
+        return with_dist
+
+    def with_local_v2_plugins(self, plugins: str | list[str] | None = None) -> Image:
```
kumare3, line 1206: cc @pingsutw is this right?

cosmicbboy: this is correct, it works after testing it on this PR

pingsutw: You can use `with_uv_project`. Much easier to iterate without rebuilding the image every time. The plugin code will be in code bundle

PR #821, opened 2026-03-20
Entrypoint flag that helps with finding important tasks

kumare3:
TaskMetadata in ListTask response should have is_entrypoint flag. 
https://github.com/flyteorg/flyte/blob/f11d46660e8ac10d9fdc4f55dda635f0a7996c42/flyteidl2/task/task_definition.proto#L115


examples/basics/devbox_one.py
```diff
@@ -22,7 +22,7 @@ async def square(i: int = 3) -> int:
     return i * i
 
 
-@env.task
+@env.task(entrypoint=True)
```
pingsutw, line 25: Should we automatically set the root task as entrypoint? If users need to specify it for the task, why they can't just use label?

src/flyte/_task_environment.py
```diff
@@ -300,8 +302,14 @@ def task(
...
         :param interruptible: Optional Whether the task is interruptible, defaults to environment setting.
         :param queue: Optional queue name to use for this task. If not set, the environment's queue will be used.
+        :param entrypoint: Optional entry point for the task, defaults to False.
+        :param task_resolver: Optional TaskResolver protocol to load tasks using custom policy.
 
         :return: A TaskTemplate that can be used to deploy the task.
+
+        Args:
+            entrypoint:
+            task_resolver:
```
pingsutw:
```suggestion
        :param entrypoint: Optional entry point for the task, defaults to False.
        :param task_resolver: Optional TaskResolver protocol to load tasks using custom policy.

        :return: A TaskTemplate that can be used to deploy the task.
```

examples/basics/devbox_one.py
kumare3: what is a root task?

pingsutw (approved):

PR #824, opened 2026-03-21
neuron parallel compilation

samhita-alla:


samhita-alla:
think we can merge it, but i want to make sure the trainium example works with this change. will wait a couple more weeks before merging.

kumare3:
cc @AdilFayyaz / @samhita-alla should we get this on flyte.clustered?

AdilFayyaz:
this could work on flyte.clustered, would be a nice example

samhita-alla:
sorry, what's flyte.clustered?

AdilFayyaz:
@samhita-alla i'm working on a new K8s JobSets plugin that allows us to run multi-node clustered tasks

samhita-alla:
@AdilFayyaz nice! so should we wait for your implementation to be merged first and then add this later?

AdilFayyaz:
I think we can still have this merged and later work on examples for JobSets, but yes the idea is to move off of Kubeflow @samhita-alla

samhita-alla:
can i get a review in that case? @AdilFayyaz

PR #826, opened 2026-03-23
[wip] Add InferencePipeline: 3-stage async pipeline for batch inference

kumare3:
InferencePipeline wires together preprocessing, batched GPU inference (via DynamicBatcher), and postprocessing with proper thread pool dispatch and backpressure — so users only define three functions:

  preprocess_fn  (per-item, sync or async, optional executor)
  inference_fn   (batched, async — the DynamicBatcher process_fn)
  postprocess_fn (per-item, receives original input + result)

Key design decisions:
- No new abstraction for batching — reuses DynamicBatcher internally
- Bounded asyncio.Queue between preprocess and batcher for backpressure
- Supports both sync and async stage functions (auto-detected)
- Sync stages can be dispatched to a ThreadPoolExecutor
- postprocess_fn receives the original raw item alongside the inference result so it has full context (filename, metadata, etc.)
- Streaming via async generator (run) or batch collect (run_all)
- Exposes underlying batcher + stats for monitoring/tuning

This pattern emerged from the embed_wikipedia and multimodal-retrieval examples, where every GPU inference use case independently wired the same load → preprocess → batch-infer → postprocess pipeline. The abstraction eliminates that boilerplate while keeping full control over DynamicBatcher tuning knobs (cost budgeting, batch size, timeouts).

### DIFF
--- a/src/flyte/extras/_inference_pipeline.py
+++ b/src/flyte/extras/_inference_pipeline.py
@@ -1,0 +1,339 @@
+"""
+InferencePipeline — 3-stage async pipeline for batch inference.
+
+Wires together preprocessing, batched GPU inference (via DynamicBatcher),
+and postprocessing with proper thread pool dispatch and backpressure so
+users only define the three stage functions.
+
+Quick start::
+
+    from flyte.extras import InferencePipeline
+
+    pipeline = InferencePipeline(
+        preprocess_fn=resize_and_normalize,   # sync or async, per-item
+        inference_fn=gpu_forward_batch,        # async, batched
+        postprocess_fn=decode_labels,          # sync or async, per-item
+    )
+
+    async with pipeline:
+        results = await pipeline.run_all(image_urls)
+
+Architecture::
+
+    items ──► preprocess_fn  ──► DynamicBatcher  ──► postprocess_fn ──► outputs
+              (executor)         (internal queues)    (executor/loop)
+
+Multiple concurrent callers (e.g. from ReusePolicy concurrency) can share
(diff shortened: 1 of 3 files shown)


src/flyte/extras/_inference_pipeline.py
```diff
@@ -0,0 +1,339 @@
...
+        Preprocessing and inference overlap: while the GPU processes batch N,
+        the CPU preprocesses items for batch N+1.
+
+        Args:
+            items: Input items (sync or async iterable).
+
+        Yields:
+            Postprocessed results, one per input item, in order.
+        """
+        if not self._batcher.is_running:
```
samhita-alla, line 238: why can’t we start the batcher by default rather than requiring the user to do it?

src/flyte/extras/_inference_pipeline.py
```diff
@@ -0,0 +1,339 @@
...
+
+        cost_estimator:
+            Optional ``(PreparedT) -> int`` for cost estimation.
+
+        max_queue_size:
+            DynamicBatcher submission queue bound.
+
+        preprocess_executor:
+            Executor for sync ``preprocess_fn`` calls.  Pass a
+            ``ThreadPoolExecutor`` for CPU-bound preprocessing.
```
samhita-alla, line 115: i think threadpoolexecutor is for i/o bound tasks no?

samhita-alla: looks like `ProcessPoolExecutor` is what we should be supporting for CPU-bound tasks.

PR #832, opened 2026-03-24
[Sandbox] Implement flyte v2 sandbox cli

popojk:
## Goal
We have a flyte v2 OSS sandbox now, this PR introduces a flyte cli to start/stop it with only 1 command

## How to test
1. Install new cli with `uv pip install -e .`
2. Run `flyte start sandbox` to run the sandbox (With a `--dev` arg, you can run the sandbox cluster without flyte-binary pod , which makes it easier for local debugging)
3. Run `flyte delete sandbox` to kill the sandbox
Screenshot 2026-03-24 at 3 42 05 PM

## Todo
We should have a document to demo how to operate sandbox with cli

### DIFF
--- a/src/flyte/remote/_client/controlplane.py
+++ b/src/flyte/remote/_client/controlplane.py
@@ -84,10 +84,6 @@ class Console:
         else:
             domain = parsed.netloc or parsed.path
 
-        # TODO: make console url configurable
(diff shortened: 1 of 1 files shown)


plugins/hitl/src/flyteplugins/hitl/_event.py
popojk, line 346: This is only auto lint fix

PR #844, opened 2026-03-26
Replace GRPC with connectRPC

jeevb:


src/flyte/remote/_client/controlplane.py
```diff
@@ -175,51 +162,39 @@ def insecure(self) -> bool:
...
+        self._secrets_service = SecretServiceClient(**shared)
+        self._identity_service = IdentityServiceClient(**shared)
+        self._trigger_service = TriggerServiceClient(**shared)
 
     @classmethod
     async def for_endpoint(cls, endpoint: str, *, insecure: bool = False, **kwargs) -> ClientSet:
         rpc_retries = kwargs.pop("rpc_retries", None)
+        # Remove grpc_options from kwargs since it's gRPC-specific
+        kwargs.pop("grpc_options", None)
+        session = await create_session(endpoint, None, insecure=insecure, rpc_retries=rpc_retries, **kwargs)
```
kumare3: what is the idea? Why dont we pass the session and store a handle to the session? what happens for long lived sessions? how do the clients get updated?

src/flyte/remote/_client/controlplane.py
```diff
@@ -284,4 +259,4 @@ def console(self) -> Console:
         return self._console
 
     async def close(self, grace: float | None = None):
```
kumare3: i would love to simply delete this

src/flyte/remote/_client/auth/_channel.py
```diff
@@ -1,247 +0,0 @@
...
-    create_auth_interceptors,
-    create_proxy_auth_interceptors,
-    get_async_proxy_authenticator,
-)
-
-# Initialize gRPC AIO early enough so it can be used in the main thread
-init_grpc_aio()
-
-
-def bootstrap_ssl_from_server(endpoint: str) -> grpc.ChannelCredentials:
```
EngHabu, line 24: Talked offline, we need a way to solve the "insecure_skip_verify" use-case... hopefully the new client allows setting that flag (like in Go)

src/flyte/remote/_client/auth/_session.py
```diff
@@ -0,0 +1,213 @@
...
+        raise ValueError(
+            f"Unknown scheme '{parsed.scheme}' in endpoint '{endpoint}'. "
+            "Use http://, https://, dns:///, or bare host:port."
+        )
+
+    # Bare host:port (no scheme at all, or urlparse mis-detected one).
+    return f"{scheme}://{endpoint}"
+
+
+def _bootstrap_ssl_from_server(endpoint: str) -> bytes:
```
jeevb, line 59: @EngHabu: `pyqwest` does not appear to expose `danger_accept_invalid_certs(true)` even though `reqwest` supports it, so we will have to retain this for working with self-signed certs.

src/flyte/remote/_client/controlplane.py
```diff
@@ -175,52 +162,31 @@ def insecure(self) -> bool:
...
-
         rpc_retries = kwargs.pop("rpc_retries", None)
-        return cls(
-            await create_channel(None, api_key, insecure=insecure, rpc_retries=rpc_retries, **kwargs),
-            endpoint,
-            insecure=insecure,
-            **kwargs,
-        )
+        session_cfg = await create_session_config(None, api_key, insecure=insecure, rpc_retries=rpc_retries, **kwargs)
+        return cls(session_cfg)
```
pmahindrakar-oss, line 189: You might want to verify API key based auth

src/flyte/__init__.py
```diff
@@ -44,27 +44,6 @@
...
-    # before importing grpc
-    if "GRPC_VERBOSITY" not in os.environ:
-        os.environ["GRPC_VERBOSITY"] = "ERROR"
-        os.environ["GRPC_CPP_MIN_LOG_LEVEL"] = "ERROR"
-        # Disable fork support (stops "skipping fork() handlers")
-        os.environ["GRPC_ENABLE_FORK_SUPPORT"] = "0"
-        # Reduce absl/glog verbosity
-        os.environ["GLOG_minloglevel"] = "2"
-        os.environ["ABSL_LOG"] = "0"
-
```
pmahindrakar-oss, line 63: Whast the equivalent of this is connect

src/flyte/_initialize.py
```diff
@@ -78,15 +78,6 @@ async def _initialize_client(
...
     from flyte.remote._client.controlplane import ClientSet
 
-    # https://grpc.io/docs/guides/keepalive/#keepalive-configuration-specification
-    channel_options = [
-        ("grpc.keepalive_permit_without_calls", 1),
-        ("grpc.keepalive_time_ms", 30000),  # Send keepalive ping every 30 seconds
-        ("grpc.keepalive_timeout_ms", 10000),  # Wait 10 seconds for keepalive response
-        ("grpc.http2.max_pings_without_data", 0),  # Allow unlimited pings without data
-        ("grpc.http2.min_ping_interval_without_data_ms", 30000),  # Min 30s between pings
-    ]
```
pmahindrakar-oss, line 88: Do we need equivalents for connect ?

src/flyte/remote/_client/auth/_session.py
```diff
@@ -0,0 +1,230 @@
...
+        proxy_authenticator = get_async_proxy_authenticator(endpoint=endpoint, proxy_command=proxy_command, **kwargs)
+    auth_http_session = get_async_session(
+        ca_cert_file_path=ca_cert_file_path, proxy_authenticator=proxy_authenticator, **kwargs
+    )
+
+    # Add proxy auth interceptors
+    proxy_auth_interceptors = create_proxy_auth_interceptors(
+        endpoint, proxy_command=proxy_command, http_session=auth_http_session, **kwargs
+    )
+    interceptors.extend(proxy_auth_interceptors)
```
pmahindrakar-oss: shouldnt this be conditional on proxy_command

src/flyte/remote/_client/auth/_session.py
```diff
@@ -0,0 +1,230 @@
...
+) -> bytes | None:
+    """Determine TLS CA certificate bytes for the pyqwest transport.
+
+    Returns PEM-encoded bytes, or None to use system defaults.
+    """
+    if insecure:
+        return None
+
+    if insecure_skip_verify:
+        return await asyncio.to_thread(_bootstrap_ssl_from_server, endpoint)
```
pmahindrakar-oss, line 96: If insecure_skip_verify is enabled, we are explicitly saying we don’t want to verify the server certificate, so fetching and trusting the server cert here seems unnecessary.

src/flyte/remote/_client/controlplane.py
jeevb: Is API key auth is used by controller to run child actions? That was tested.

src/flyte/remote/_client/auth/_session.py
jeevb: Unfortunately, we don't have a hook to disable verification in the underlying http client library, so we have to keep this functionality.

src/flyte/__init__.py
jeevb: Hopefully this isn't necessary anymore. Haven't had issues, but can always add back if we're seeing noisy logs.

src/flyte/remote/_client/auth/_session.py
```diff
@@ -0,0 +1,230 @@
...
+    return None
+
+
+def _build_pyqwest_client(tls_ca_cert: bytes | None = None) -> pyqwest.Client:
+    """Build a pyqwest Client with transport defaults matching the old gRPC channel config.
+
+    These defaults are always set explicitly so behaviour doesn't silently
+    change if pyqwest changes its own defaults in a future release.
+
+    Mapping from old gRPC channel options:
```
jeevb: @pmahindrakar-oss: timeout config with python connectrpc

src/flyte/remote/_client/auth/_session.py
jeevb: `create_proxy_auth_interceptors` handles this internally - existing logic.

src/flyte/_initialize.py
jeevb: See: https://github.com/flyteorg/flyte-sdk/blob/005f556b75ee492e3dd5684dee34e930e90c79d0/src/flyte/remote/_client/auth/_session.py#L107

src/flyte/remote/_client/auth/_session.py
pmahindrakar-oss: Feel redundant to do checks in outer call for proxy_command and also inside checking for the same. 
Can we simplify here and move the check out side and add the interceptor only when proxy_command is set

pmahindrakar-oss (dismissed):

pmahindrakar-oss (dismissed):

EngHabu (dismissed):
LGTM

src/flyte/remote/_client/auth/_session.py
pmahindrakar-oss: Also do we have insecure mode where we don't look for any tls cert. That would be needed for sandbox kind of modes of deployment .

EngHabu (approved):

PR #857, opened 2026-03-27
adds papermill plugin

samhita-alla:


plugins/papermill/src/flyteplugins/papermill/resolver.py
```diff
@@ -0,0 +1,101 @@
...
+        mod = importlib.import_module(name=task_module)
+        task = getattr(mod, task_name, None)
+        if task is None:
+            raise ValueError(
+                f"Could not find NotebookTask '{task_name}' in module '{task_module}'. "
+                "Make sure the NotebookTask is assigned to a module-level variable."
+            )
+        return task
+
+    def loader_args(self, task: TaskTemplate, root_dir: pathlib.Path | None) -> list[str]:
```
kumare3, line 91: we could load it using the task_name itself. 
Ideally honestly i would love to support tasks to be like container task. Why do we need to load the Notebook task? Can we not just send the notebook path and just run that directly? you might need the task for inputs / outputs maybe.

```python
async def my_task():
    t = NotebookTask("my_nb.ipynb")
    await t(...)
```
Ideally i would like ^ this to work. if you use module level resolver this cannot work.

fiedlerNr9 (dismissed):
Works like a charm! Thanks for building this 🚀

Somehow, I want the user to see this is a Notebook task when clicking on summary. Maybe showing a Integration link with a JN icon, similiar to what we have for spark and ray. The link could just be a redirect to the report tab?

What do you think?

plugins/papermill/src/flyteplugins/papermill/resolver.py
samhita-alla: updated the resolver. it now bakes the notebook path and i/o type schemas into the task spec at registration time so it doesn't have to be a module-level resolver. this works?

```python
@env.task
async def async_inline_workflow(x: int = 3, y: float = 1.5) -> int:
    from flyteplugins.papermill import NotebookTask

    nb = NotebookTask(
        name="add_numbers_async_inline",
        notebook_path="notebooks/basic_math.ipynb",
        task_environment=env,
        inputs={"x": int, "y": float},
        outputs={"result": int},
    )
    # Use .aio() when calling a NotebookTask from an async task function
    return await nb.aio(x=x, y=y)
```

plugins/papermill/src/flyteplugins/papermill/resolver.py
```diff
@@ -0,0 +1,164 @@
...
+            result: dict = {}
+            for field_name, lt_dict in schema.items():
+                lt = LiteralType()
+                json_format.ParseDict(lt_dict, lt)
+                result[field_name] = TypeEngine.guess_python_type(lt)
+            return result
+
+        inputs = _schema_to_types(args_dict.get("input-schema", "{}"))
+        outputs = _schema_to_types(args_dict.get("output-schema", "{}"))
+        config: dict = json.loads(args_dict.get("config", "{}"))
```
kumare3, line 80: ooh this is dangerous. hmm we might have very big right, does it work always?

kumare3, line 80: I think if you want to pass it so, just pickle the entire task and send it no?

samhita-alla: it does work with the examples i tested. why would it be big? it's just the types and the papermill config, no?

samhita-alla: hmm this should work too, right? papermill doesn’t support a wide range of types: only JSON-serializable ones (`int`, `str`, `bool`, `float`, `list`, `dict`, `None`), along with `File`, `Dir`, and `DataFrame`. these should work well with the plugin.

kumare3 (approved):

PR #861, opened 2026-03-29
[Demo cluster] Skip cluster status check in dev mode

popojk:
## Goal

In dev mode, user will need to start sandbox outside the cluster. Thus, the cli should skip cluster status check in dev mode.

### DIFF
--- a/src/flyte/cli/_demo.py
+++ b/src/flyte/cli/_demo.py
@@ -192,6 +192,5 @@ def launch_demo(image_name: str, is_dev_mode: bool) -> None:
     _switch_k8s_context()
     if not is_dev_mode:
         _wait_for_console_ready(_CONSOLE_READYZ_URL)
-
-    click.echo("\nFlyte demo cluster is ready!")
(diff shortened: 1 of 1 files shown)


src/flyte/cli/_demo.py
```diff
@@ -190,7 +190,8 @@ def launch_demo(image_name: str, is_dev_mode: bool) -> None:
 
     _merge_kubeconfig(_KUBECONFIG_PATH, _CONTAINER_NAME)
     _switch_k8s_context()
-    _wait_for_console_ready(_CONSOLE_READYZ_URL)
+    if not is_dev_mode:
+        _wait_for_console_ready(_CONSOLE_READYZ_URL)
 
     click.echo("\nFlyte demo cluster is ready!")
     click.echo("UI is available at http://localhost:30080/v2")
```
pingsutw: Could we also print these message only in non-dev mode?

pingsutw (approved):

PR #883, opened 2026-04-01
 fix: surface worker errors hidden inside BaseExceptionGroup in parallel reader

AdilFayyaz:
When `obstore.get_range_async` raised a transient error (e.g. GCS 429), it was swallowed inside a BaseExceptionGroup by asyncio.TaskGroup, showing only "Filtered traceback" with no actionable message in task logs.

 ## Summary

  - Replaced logging.getLogger(__name__) with the shared flyte._logging.logger, consistentwith other storage modules
  - Added inline try/except around task = await inq.get() — logs "Worker failed before receiving a task" if the queue itself raises
  - Added inline try/except around obstore.get_range_async(...) — logs the failing file path and chunk offset before re-raising

 ## Test Plan

  - test_worker_logs_exception_on_download_failure — mocks obstore.get_range_async to raise RuntimeError("GCS 429"), asserts logger.exception is called with the failing file path, and that the exception still propagates
  - test_worker_logs_exception_before_task_received — replaces asyncio.Queue with a subclass that raises on get() only for bounded queues (targeting inq, not outq), asserts the fallback log message is emitted

### DIFF
--- a/src/flyte/storage/_parallel_reader.py
+++ b/src/flyte/storage/_parallel_reader.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import asyncio
 import dataclasses
 import io
-import logging
 import os
 import pathlib
 import sys
@@ -15,7 +14,7 @@ import aiofiles
 import aiofiles.os
 import obstore
 
-logger = logging.getLogger(__name__)
+from flyte._logging import logger
 
 if typing.TYPE_CHECKING:
     from obstore import Bytes, ObjectMeta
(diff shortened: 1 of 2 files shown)


src/flyte/storage/_parallel_reader.py
```diff
@@ -14,6 +15,8 @@
 import aiofiles.os
 import obstore
 
+logger = logging.getLogger(__name__)
```
pingsutw: can we use flyte logger
```
from flyte import logger
```

pingsutw (approved):

PR #885, opened 2026-04-01
implement intratask checkpointing in the sdk

cosmicBboy:
This pull request introduces a comprehensive suite of example scripts demonstrating how to implement checkpointing and failure recovery for machine learning training workflows using Flyte's `AsyncCheckpoint` mechanism. The examples cover a range of ML frameworks including scikit-learn, PyTorch, PyTorch Lightning, and Hugging Face Transformers, each showing how to persist and restore training state across task retries. The scripts also illustrate best practices for simulating failures and integrating checkpointing into iterative training processes.

The most important changes are:

**Scikit-learn checkpointing:**
* Added `sklearn_partial_checkpoint.py` to demonstrate incremental training with `SGDClassifier.partial_fit`, using checkpointing to persist and restore model state and training progress across retries.

**PyTorch checkpointing:**
* Added `pytorch_task_checkpoint.py` to show checkpointing of a simple PyTorch model's state dict and optimizer at each epoch, with failure simulation and recovery.

**PyTorch Lightning checkpointing:**
* Added `pytorch_lightning_checkpoint.py` to illustrate checkpointing for a LightningModule using `ModelCheckpoint` and custom callbacks, including logic for resuming from the latest checkpoint and injecting failures.

**Hugging Face Transformers checkpointing:**
* Added `huggingface_trainer_checkpoint.py` to provide an example of checkpointing with `transformers.Trainer`, integrating Flyte checkpointing with Hugging Face's checkpointing and recovery system, and simulating failures at defined intervals.

**Generic byte checkpointing:**
* Added `generic_data_checkpoint.py` as a minimal example of checkpointing arbitrary byte data, demonstrating the basic checkpoint load/save/retry pattern.

### DIFF
--- a/src/flyte/_checkpoint.py
+++ b/src/flyte/_checkpoint.py
@@ -9,7 +9,7 @@ open / open_sync in ``"wb"`` mode) so explicit `file://` checkpoint URIs work wi
 - **Sync tasks:** `checkpoint.load_sync()`, `checkpoint.save_sync(...)` (same semantics as
   `flyte.io.File.download_sync` / sync uploads — no `flyte.syncify`).
 
-`flyte.models` exposes `flyte.models.Checkpoints`, a small value type with the platform
+`flyte.models` exposes `flyte.models.CheckpointPaths`, a small value type with the platform
 `checkpoint_path` and `prev_checkpoint_path` strings. `flyte.Checkpoint` is the runtime helper
 that downloads the previous attempt's **blob** into a temp workspace and uploads a new blob to the
 task output prefix (including tarball encoding when you save a directory).
--- a/examples/advanced/multi_run.py
+++ b/examples/advanced/multi_run.py
@@ -12,7 +12,7 @@ async def double(x: int) -> int:
     print(flyte.ctx().raw_data_path)
     print(flyte.ctx().data)
     print(flyte.ctx().code_bundle)
-    print(flyte.ctx().checkpoints)
+    print(flyte.ctx().checkpoint_paths)
     return x * 2
 
 
--- a/src/flyte/_bin/runtime.py
+++ b/src/flyte/_bin/runtime.py
@@ -96,7 +96,7 @@ def main(
     from flyte._internal.imagebuild.image_builder import ImageCache
     from flyte._internal.runtime.entrypoints import load_and_run_task
     from flyte._logging import logger
-    from flyte.models import ActionID, Checkpoints, CodeBundle, RawDataPath
+    from flyte.models import ActionID, CheckpointPaths, CodeBundle, RawDataPath
 
     logger.info("Registering faulthandler for SIGUSR1")
     faulthandler.register(signal.SIGUSR1)
@@ -149,7 +149,7 @@ def main(
         resolver_args=resolver_args,
         action=ActionID(name=name, run_name=run_name, project=project, domain=domain, org=org),
         raw_data_path=RawDataPath(path=raw_data_path, path_rewrite=path_rewrite),
-        checkpoints=Checkpoints(prev_checkpoint_path=prev_checkpoint, checkpoint_path=checkpoint_path),
+        checkpoint_paths=CheckpointPaths(prev_checkpoint_path=prev_checkpoint, checkpoint_path=checkpoint_path),
         code_bundle=bundle,
         input_path=inputs,
         output_path=outputs_path,
--- a/src/flyte/_internal/controllers/_local_controller.py
+++ b/src/flyte/_internal/controllers/_local_controller.py
@@ -20,7 +20,7 @@ from flyte._persistence._recorder import RunRecorder
 from flyte._persistence._task_cache import LocalTaskCache
 from flyte._task import AsyncFunctionTaskTemplate, TaskTemplate
 from flyte._utils.helpers import _selector_policy
-from flyte.models import ActionID, Checkpoints, NativeInterface
+from flyte.models import ActionID, CheckpointPaths, NativeInterface
 from flyte.remote._task import TaskDetails
 from flyte.storage._storage import strip_file_header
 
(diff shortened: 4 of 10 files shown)


src/flyte/_checkpoint.py
```diff
@@ -0,0 +1,302 @@
...
+            await storage.put_stream(data, to_path=self._checkpoint_dest)
+            return
+
+        src = pathlib.Path(data) if isinstance(data, str) else data
+        if not src.exists():
+            raise FileNotFoundError(f"Checkpoint source path does not exist: {src}")
+        if src.is_dir():
+            await self._save_directory_as_tarball(src)
+        else:
+            await storage.put(str(src), self._checkpoint_dest, recursive=False)
```
samhita-alla: i don't think we should be using `@syncify` with async APIs in this case. we should switch to using sync APIs when `save` is a sync method.

samhita-alla: same with `load()`

src/flyte/_checkpoint.py
```diff
@@ -0,0 +1,432 @@
+"""
+Task checkpointing against Flyte checkpoint object-store prefixes (v2 SDK).
+
+:class:`Checkpoint` uses ``flyte.storage`` for I/O.
+
+- **Async tasks:** ``await checkpoint.load()``, ``await checkpoint.save(...)``.
+- **Sync tasks:** :meth:`~Checkpoint.load_sync`, :meth:`~Checkpoint.save_sync` (same semantics as
+  :meth:`flyte.io.File.download_sync` / sync uploads — no :mod:`flyte.syncify`).
```
samhita-alla:
```suggestion
- **Sync tasks:** ``checkpoint.load_sync()``, ``checkpoint.save_sync(...)``.
```

src/flyte/_checkpoint.py
```diff
@@ -0,0 +1,432 @@
...
+            if src.startswith(_REMOTE_CHECKPOINT_SCHEMES):
+                src = repair_union_prev_checkpoint_uri(src)
+            self._checkpoint_src = src
+        self._td = tempfile.TemporaryDirectory(prefix="flyte-cp-")
+        self._restored = False
+        self._is_tarball = False
+        self._had_remote_prev = False
+
+    def __del__(self) -> None:
+        td = getattr(self, "_td", None)
```
samhita-alla, line 318: can we add a comment explaining what `_td` is?

src/flyte/_checkpoint.py
```diff
@@ -0,0 +1,432 @@
...
+        return self._payload_path
+
+    async def load(self) -> Optional[pathlib.Path]:
+        if not self.prev_exists():
+            return None
+        if self._restored:
+            return self._load_return_path(is_tarball=self._is_tarball)
+
+        prev_uri = self._checkpoint_src
+        assert prev_uri is not None
```
samhita-alla: should we raise an exception instead of asserting?

samhita-alla: i don't think we need this because `if not self.prev_exists():` already checks for the existence of checkpoint source.

src/flyte/_checkpoint.py
```diff
@@ -0,0 +1,432 @@
...
+
+        download_path = _new_checkpoint_download_temp_path()
+        is_tarball = False
+        try:
+            _clear_directory_contents(self.path)
+            self.path.mkdir(parents=True, exist_ok=True)
+            _get_checkpoint_object_sync(prev_uri, download_path)
```
samhita-alla: looks like this is the only difference between the async and sync load methods. can we maybe extract a common helper?

samhita-alla: we can common helpers for `save` and `save_sync` as well.

src/flyte/_checkpoint.py
```diff
@@ -0,0 +1,432 @@
...
+        if not src.exists():
+            raise FileNotFoundError(f"Checkpoint source path does not exist: {src}")
+        if src.is_dir():
+            self._save_directory_as_tarball_sync(src)
+        else:
+            _put_checkpoint_file_sync(str(src), self._checkpoint_dest)
+
+
+# Backward compatibility
+AsyncCheckpoint = Checkpoint
```
samhita-alla: do we need this?

src/flyte/_checkpoint.py
```diff
@@ -0,0 +1,432 @@
...
+    @abstractmethod
+    def load_sync(self) -> Optional[pathlib.Path]:
+        """Load checkpoint data from the remote source (sync task code)."""
+
+    @abstractmethod
+    def save_sync(self, data: pathlib.Path | str | bytes) -> None:
+        """Save checkpoint data to the remote destination (sync task code)."""
+
+
+class Checkpoint(BaseCheckpoint):
```
samhita-alla, line 289: why are we introducing a new `Checkpoint` class when `Checkpoints` already exists? what are we using the old one for? i know it’s incomplete but curious why we aren’t using it.

examples/checkpoint/pytorch_lightning_checkpoint.py
```diff
@@ -0,0 +1,177 @@
...
+        self._flyte_checkpoint = flyte_checkpoint
+
+    @override
+    def on_train_epoch_end(self, trainer: L.Trainer, pl_module: L.LightningModule) -> None:
+        super().on_train_epoch_end(trainer, pl_module)
+        if self.dirpath:
+            self._flyte_checkpoint.save_sync(pathlib.Path(self.dirpath))
+
+
+def find_last_checkpoint(root: pathlib.Path) -> str | None:
```
samhita-alla: can we have this as a helper so users don't have to worry about retrieving the last or the latest?

src/flyte/_checkpoint.py
```diff
@@ -0,0 +1,432 @@
...
+                exists = True
+        else:
+            exists = file_system.exists(from_path)
+        if not exists:
+            raise AssertionError(f"Unable to load data from {from_path}") from oe
+        file_system = _get_anonymous_filesystem(from_path)
+        file_system.get(str(from_path), str(to_path), recursive=False)
+
+
+def _put_checkpoint_file_sync(from_local: str, to_remote: str) -> None:
```
samhita-alla: why aren't we using `File`/`Dir` abstractions in save and load? why `storage` directly? it doesn't work for bytes but should work for file/dir, right?

src/flyte/_checkpoint.py
cosmicbboy: `Checkpoints` is a thin dataclass for just holding the task context metadata

PR #893, opened 2026-04-02
Use localhost registry when flyte endpoint is localhost

pingsutw:
## Summary
- When the configured Flyte endpoint contains `localhost`, automatically use `localhost:30000` as the image registry instead of `ghcr.io/flyteorg`
- Configure the buildx builder with `--driver-opt network=host` so the BuildKit daemon can reach localhost registries on the host machine
- Adds `_get_base_registry()` function that checks the initialized config endpoint dynamically

## Test plan
- [x] Verify default registry remains `ghcr.io/flyteorg` when endpoint is not localhost
- [x] Verify registry switches to `localhost:30000` when endpoint contains `localhost`
- [x] Verify buildx builder can push to localhost registry in sandbox environment

### DIFF
--- a/tests/flyte/imagebuild/test_image_build_engine.py
+++ b/tests/flyte/imagebuild/test_image_build_engine.py
@@ -260,9 +260,7 @@ async def test_local_docker_checker_arch_mismatch(mock_exec):
         ]
     }
     mock_exec.return_value = _make_mock_process(0, json.dumps(manifest).encode(), b"")
-    result = await LocalDockerCommandImageChecker.image_exists(
-        "registry/img", "tag1", arch=("linux/amd64",)
-    )
+    result = await LocalDockerCommandImageChecker.image_exists("registry/img", "tag1", arch=("linux/amd64",))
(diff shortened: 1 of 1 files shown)


src/flyte/_image.py
```diff
@@ -456,10 +456,26 @@ def from_dict(cls, envs: Dict[str, str]) -> Env:
...
+    """
+    Returns the base registry based on the Flyte config endpoint.
+    If the endpoint contains 'localhost', use the localhost registry.
+    """
+    from flyte._initialize import _get_init_config
+
+    init_config = _get_init_config()
+    if init_config and init_config.client:
+        endpoint = init_config.client.endpoint
+        if endpoint and "localhost" in endpoint:
```
kumare3, line 474: What if someone port forwards? Maybe an env var to disable this

pingsutw: They can use `.clone()` to override the registry. the original default registry (`ghcr.io/flyteorg`) doesn't work for them as well.

PR #901, opened 2026-04-03
fix(spark): guard against code_bundle being None in pre()

rbavery:
- Fix `AttributeError: 'NoneType' object has no attribute 'destination'` in `PysparkFunctionTask.pre()` when `code_bundle` is `None`
- This occurs when a Spark task is registered with code baked into the Docker image (not via fast-register / code bundle upload)


pingsutw:
Thank you!

PR #903, opened 2026-04-03
fix bug in Image.with_source_file not bumping image hash

cosmicBboy:
This pull request improves file hashing logic and type safety in the project. The main focus is on ensuring that both files and directories are correctly incorporated into hash computations, and on improving development experience with better type checking.

**File hashing and digest computation improvements:**

* Updated the `update_hasher_for_source` function docstring to clarify that it now handles both single files and directory trees, ensuring content and relative paths are incorporated into the hash.
* Fixed the logic in `compute_digest_for_dir` to properly handle the case when the input is a single file, ensuring that file contents are hashed even when not inside a directory.
* Added a test in `test_with_source` to verify that hashing a single file reflects changes in its content, ensuring correctness for image layer hashing.

**Development environment/type checking:**

* Added `types-PyYAML` and `types-aiofiles` to the `dev` dependencies in `pyproject.toml` to improve static type checking and developer experience.

### DIFF
--- a/src/flyte/_utils/file_handling.py
+++ b/src/flyte/_utils/file_handling.py
@@ -38,7 +38,7 @@ def update_hasher_for_source(
     Incorporates a single file, or walks a directory tree, into the hasher (content + relative paths).
     :param os.PathLike source:
     :param ignore: Optional ignore instance whose is_ignored(abs_path) determines whether to skip a file.
-    :return Text:
+    :return None:
     """
 
     def compute_digest_for_file(path: os.PathLike, rel_path: os.PathLike) -> None:


src/flyte/_utils/file_handling.py
```diff
@@ -35,7 +35,7 @@ def update_hasher_for_source(
     ignore: Optional[_IgnoreLike] = None,
 ):
     """
-    Walks the entirety of the source dir to compute a deterministic md5 hex digest of the dir contents.
+    Incorporates a single file, or walks a directory tree, into the hasher (content + relative paths).
     :param os.PathLike source:
     :param ignore: Optional ignore instance whose is_ignored(abs_path) determines whether to skip a file.
     :return Text:
```
AdilFayyaz: cleanup: return type is None

cosmicbboy: done!

PR #952, opened 2026-04-09
Add MLflow examples: tracking-server app and model-selection pipeline

chrismatteson:


samhita-alla:
@chrismatteson since we don't support data persistence with apps, do you think it's a good idea to support hosting mlflow as an app? i feel like apps are better suited for stateless applications where you just want to show something or when you want to host models etc. and don't need persistent storage.

chrismatteson:
@samhita-alla - Hmm, I see your point. I could adjust this to allow an external database as a variable of some sort.

It's a useful demo, which is why I want it, but maybe it's not a good production example. I'm not sure if we feel like these examples should all be production ready.

samhita-alla:
@chrismatteson they don't have to be production-ready but i think this is an anti-pattern that we probably shouldn't be encouraging. if you can support persistence via an external database, that works. but even then, i'm not sure it's something we should be doing given that mlflow already has a hosted solution.

PR #955, opened 2026-04-09
some examples

dansola:


examples/plugins/ray_override_test.py
```diff
@@ -0,0 +1,197 @@
+"""
+Test: Can we parameterize Ray worker GPU/machine type via override(plugin_config=...)?
+
+Context (Ryan Avery question):
+- He has a Ray task with workers requesting GPUs (A10, L40s, etc.)
+- He wants to avoid duplicating tasks/envs for each GPU variant
+- override(resources=...) does NOT propagate to Ray worker pods (only affects primary container)
+- override(plugin_config=...) DOES let you swap in a new RayJobConfig at invocation time
```
samhita-alla, line 8: can we remove this?

examples/type_transformers/test_geopandas_in_dataclass.py
```diff
@@ -0,0 +1,462 @@
...
+
+            if storage.is_remote(remote_dir):
+                # Write locally first, then upload to remote storage
+                local_dir = tempfile.mkdtemp(prefix="geodf_")
+                local_path = str(Path(local_dir) / "data.parquet")
+                self._gdf.to_parquet(local_path)
+                remote_path = remote_dir.rstrip("/") + "/data.parquet"
+                from flyte._utils.asyn import loop_manager
+
+                loop_manager.run_sync(storage.put, local_path, remote_path)
```
samhita-alla, line 133: you can do this instead: https://github.com/flyteorg/flyte-sdk/blob/7aa6b974504a17f7e54c891d2c78a0e9d830a046/src/flyte/io/_file.py#L818-L823 i.e. we can use sync methods here instead of forcing async into a sync flow.

examples/type_transformers/test_geopandas_in_dataclass.py
```diff
@@ -0,0 +1,462 @@
...
+#    "geopandas",
+#    "shapely",
+#    "pyarrow",
+#    "numpy",
+#    "flyte",
+# ]
+# ///
+
+"""
+Test: geopandas GeoDataFrame nested inside a dataclass via SerializableType.
```
samhita-alla, line 12: can we add an example instead of tests? we can move the ser-des into a separate file to keep things clean and import it into the example. we could also move this into a plugin if that makes things easier.

PR #956, opened 2026-04-10
Only use localhost registry for dev-version default image

pingsutw:
## Summary
- Previously, a released Flyte SDK pointed at a localhost endpoint would pick up the localhost registry (`localhost:30000`) for the default image. Released SDKs should always pull the default image from `ghcr.io/flyteorg` — only dev versions (which don't exist on the public registry) should fall back to the localhost registry.
- Restrict the localhost-registry swap in `Image._get_default_image_for` to `dev_mode` only.
- Add tests covering released vs. dev version × localhost vs. remote endpoint, and the `install_flyte=False` path.

## Test plan
- [x] `uv run pytest tests/flyte/test_image.py -k "default_image_uses or get_base_registry"` — all pass
- [x] Released SDK pointed at a localhost demo cluster uses `ghcr.io/flyteorg` for the default image
- [x] Dev SDK pointed at localhost still uses `localhost:30000`

### DIFF
--- a/src/flyte/_image.py
+++ b/src/flyte/_image.py
@@ -603,7 +603,7 @@ class Image:
             preset_tag = f"py{python_version[0]}.{python_version[1]}-{suffix}"
         image = Image._new(
             base_image=f"python:{python_version[0]}.{python_version[1]}-slim-bookworm",
-            registry=_get_base_registry(),
+            registry=_get_base_registry() if dev_mode else _BASE_REGISTRY,
             name=_DEFAULT_IMAGE_NAME,
             python_version=python_version,
             platform=("linux/amd64", "linux/arm64") if platform is None else platform,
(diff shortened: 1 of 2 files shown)


src/flyte/_image.py
```diff
@@ -603,7 +603,7 @@ def _get_default_image_for(
             preset_tag = f"py{python_version[0]}.{python_version[1]}-{suffix}"
         image = Image._new(
             base_image=f"python:{python_version[0]}.{python_version[1]}-slim-bookworm",
-            registry=_get_base_registry(),
+            registry=_get_base_registry() if dev_mode else _BASE_REGISTRY,
```
machichima, line 606: Could we update `_get_base_registry` to also check `dev_mode` instead?

PR #957, opened 2026-04-10
Replace _is_flyte_default with _is_cloned on Image

pingsutw:
## Summary

Replace the `_is_flyte_default` flag on `Image` with `_is_cloned` (inverted semantics): the SDK only builds an image when the user has modified it via `clone()` or a `with_*` method. Unmodified default images short-circuit the existence check and their URI is taken directly from their `base_image`.

### Core changes in `src/flyte/_image.py`
- **Field rename with flipped default.** `_is_flyte_default: bool = False` → `_is_cloned: bool = False`. A fresh image is treated as "already exists in a registry" until something mutates it.
- **Early return for released default images.** `_get_default_image_for` now returns a bare-URI image (`base_image=ghcr.io/flyteorg/flyte:<preset_tag>`, `registry=_get_base_registry()`, `name=_DEFAULT_IMAGE_NAME`, `extendable=True`) for released (non-dev) Flyte versions, instead of rebuilding the layer stack. The `packaging.version.Version` import and the devrelease/prerelease gating are gone.
- **`uri` property rewritten.** Non-cloned images return their `base_image` directly. Cloned images compose `{registry}/{name}:{tag}` (with `name` now required). No more fall-throughs for partially-defined images.
- **`clone()` and `from_dockerfile` mark the produced image as cloned.** Every `with_*` helper flows through `clone()`, so any user-facing mutation automatically flips `_is_cloned=True`.
- **`resolve_code_bundle_layer` preserves `_is_cloned`** in both the strip and resolve-to-CopyConfig reconstruction paths, so hash stability holds for default (non-cloned) images.

### Builder / tooling
- `src/flyte/_internal/imagebuild/image_builder.py`: `image_exists` short-circuits on `not image._is_cloned` instead of `_is_flyte_default`.
- `maint_tools/build_default_image.py`: dropped the redundant `object.__setattr__(default_image, "_is_flyte_default", False)` override. `from_debian_base(registry=..., name=...)` now naturally goes through `clone()`, which flips `_is_cloned=True`, so the build proceeds without help.

### Tests
- `tests/flyte/test_image.py`: 8 new tests covering released default, dev default, `with_pip_packages`, explicit `clone()`, `from_base`, `from_ref_name`, `from_debian_base` with a registry override, and `from_base` + layers.
- `tests/flyte/test_image_builder.py`: 3 new tests confirming `ImageBuildEngine.image_exists` short-circuits (no backend calls) for unmodified default / `from_base` images, and that it *does* query the backend once a layer is added.

## Test plan
- [x] `uv run pytest tests/flyte/test_image.py tests/flyte/test_image_builder.py` — all passing (pre-existing `test_default_base_image` failure is unrelated)
- [ ] Released SDK pointed at a demo cluster: default `from_debian_base()` is not rebuilt; `with_pip_packages(...)` triggers a build
- [ ] `from_base("ghcr.io/...")` does not attempt a build unless modified


popojk (approved):
LGTM, thanks!

PR #958, opened 2026-04-10
Fix duration of previous runs not reported in TUI

proog128:
Fixes the duration displayed in the terminal UI for execution of steps of previous runs. The duration has been 0 before, because start and end times stored in the database were ignored.


pingsutw (approved):

pingsutw:
Thank you

PR #959, opened 2026-04-10
Add dataproxy client selector

katrogan:
NOT TO BE MERGED UNTIL [BE](https://github.com/flyteorg/flyte/pull/7184) CHANGES ARE IN.

This PR adds a new dependency on the flyte remote client for dataproxy endpoints
1. Use a client conn cache for dataproxy operations, using 'operation' and 'resource' as cache keys
2. On cache misses call SelectCluster to get the respective cluster endpoint and initialize and cache a new client conn

Going forward all new dataproxy calls should use this pattern.

Today this PR updates the only uses of dataproxy in the sdk: UploadInputs and CreateUploadLocation

---
_Testing_

Verified that the CreateRun path (which calls UploadInputs now) works with a local cluster

```
flyte -vvv  --config .flyte/config-oss-local.yaml  run -p flytesnacks -d development examples/basics/hello.py main
...
╭────────────────────────────────────────── Remote Run ───────────────────────────────────────────╮
│ Created Run: rmj7wwft78l69stlj69j                                                               │
│ URL: http://localhost:30080/v2/domain/development/project/flytesnacks/runs/rmj7wwft78l69stlj69j │
╰─────────────────────────────────────────────────────────────────────────────────────────────────╯
```

ref 26-353


src/flyte/remote/_client/controlplane.py
```diff
@@ -232,6 +236,59 @@ def identity_service(self) -> IdentityService:
     def trigger_service(self) -> TriggerService:
         return self._trigger_service
 
+    @property
+    def cluster_service(self) -> ClusterService:
+        return self._cluster_service
+
+    async def get_dataproxy_for_resource(self, operation: int, resource: object) -> DataProxyService:
```
EngHabu: Should this operation be an enum instead?

src/flyte/remote/_client/controlplane.py
```diff
@@ -232,6 +236,59 @@ def identity_service(self) -> IdentityService:
...
+            field_map = {
+                "OrgIdentifier": "org_id",
+                "ProjectIdentifier": "project_id",
+                "TaskIdentifier": "task_id",
+                "ActionIdentifier": "action_id",
+                "ActionAttemptIdentifier": "action_attempt_id",
+            }
+            field_name = field_map.get(type(resource).__name__)
+            if field_name:
+                getattr(req, field_name).CopyFrom(resource)
```
EngHabu: Is this the best way to create the one of 😬, @wild-endeavor @pingsutw ?

src/flyte/remote/_client/controlplane.py
```diff
@@ -232,6 +236,59 @@ def identity_service(self) -> IdentityService:
...
+
+        resp = await self._cluster_service.select_cluster(req)
+        cluster_endpoint = resp.cluster_endpoint
```
EngHabu: Are you going to normalize this here? stripping/adding http/s or dns:/// ?

src/flyte/remote/_client/controlplane.py
```diff
@@ -232,6 +236,59 @@ def identity_service(self) -> IdentityService:
...
+                "ProjectIdentifier": "project_id",
+                "TaskIdentifier": "task_id",
+                "ActionIdentifier": "action_id",
+                "ActionAttemptIdentifier": "action_attempt_id",
+            }
+            field_name = field_map.get(type(resource).__name__)
+            if field_name:
+                getattr(req, field_name).CopyFrom(resource)
+
+        resp = await self._cluster_service.select_cluster(req)
```
EngHabu: Let's make sure we are throwing informative errors here in the infamous "no healthy clusters" error.. we now have a good place to catch that and raise a good error to the user

EngHabu (commented):
LGTM, @pingsutw @wild-endeavor should advise though

EngHabu (dismissed):

src/flyte/remote/_client/controlplane.py
pingsutw: we can do something like
```python
  req = SelectClusterRequest(operation=operation)                  
  if hasattr(resource, "DESCRIPTOR"):                                                                    
      oneof = req.DESCRIPTOR.oneofs_by_name["resource"]  # replace with actual oneof name
      for field in oneof.fields:                                                                         
          if field.message_type is resource.DESCRIPTOR:                                                  
              getattr(req, field.name).CopyFrom(resource)                                                
              break    
```

src/flyte/remote/_client/controlplane.py
katrogan: not necessary! create_session_config already calls [normalize_rpc_endpoint](https://github.com/flyteorg/flyte-sdk/blob/98ab1bd9f9c57a3c62316c28ab036e30f2393f89/src/flyte/remote/_client/auth/_session.py#L34)

src/flyte/remote/_client/controlplane.py
katrogan: done

src/flyte/remote/_client/controlplane.py
katrogan: nice thanks. done

src/flyte/remote/_client/controlplane.py
katrogan: updated

src/flyte/remote/_client/controlplane.py
```diff
@@ -232,6 +340,10 @@ def identity_service(self) -> IdentityService:
     def trigger_service(self) -> TriggerService:
         return self._trigger_service
 
+    @property
+    def cluster_service(self) -> ClusterService:
```
kumare3: Let's not expose this? Do we need it anywhere at all

katrogan: done

src/flyte/remote/_client/controlplane.py
```diff
@@ -161,6 +166,99 @@ def insecure(self) -> bool:
...
+            project_id,
+        )
+        return await client.upload_inputs(request)
+
+    async def _resolve(
+        self,
+        operation: cluster_payload_pb2.SelectClusterRequest.Operation,
+        project_id: identifier_pb2.ProjectIdentifier,
+    ) -> DataProxyService:
+        from flyte._logging import logger
```
kumare3: this is not "co-routine" safe. This will lead to a race condition

katrogan: done, thank you

src/flyte/remote/_client/controlplane.py
```diff
@@ -161,6 +167,117 @@ def insecure(self) -> bool:
...
+            project_id = request.project_id
+        else:
+            raise ValueError("UploadInputsRequest must set either run_id or project_id")
+        client = await self._resolve(
+            cluster_payload_pb2.SelectClusterRequest.Operation.OPERATION_UPLOAD_INPUTS,
+            project_id,
+        )
+        return await client.upload_inputs(request)
+
+    async def _resolve(
```
kumare3: you should be just using alru_cache on this right?

src/flyte/remote/_client/controlplane.py
```diff
@@ -161,6 +167,117 @@ def insecure(self) -> bool:
...
+    ) -> DataProxyService:
+        cache_key = int(operation).to_bytes(4, "little") + project_id.SerializeToString(deterministic=True)
+        # Single-threaded asyncio: between this lookup and the assignment below
+        # nothing else runs. Concurrent callers either see an existing future
+        # to await on, or end up storing the same brand-new future.
+        existing = self._cache.get(cache_key)
+        if existing is not None:
+            return await existing
+
+        loop = asyncio.get_running_loop()
```
kumare3: why do we need this? Why do we need a future? If there is a client can we just not cache it? if you use alru_cache you can simply delete all of this code, just return the cache and let alru_cache handle all of this work for you

kumare3 (approved):

PR #966, opened 2026-04-13
Add extended documentation support to CLI options

ppiegaze:
## Summary

- Adds `DocOption(click.Option)` subclass with a `docs` parameter for rich per-option markdown content
- Extends `_render_command()` in the doc generator to emit per-option subsections (e.g. `#### --copy-style`) after the summary table
- Adds `_docs.py` module with extended documentation constants for `flyte run` (11 options) and `flyte deploy` (9 options)
- `docs` content appears only in `flyte gen docs` output, never in `--help` terminal output

This enables the generated CLI reference (`flyte-cli.md`) to include detailed per-option documentation (usage examples, "how it works" sections, use cases) that was previously maintained as separate hand-written pages in the docs repo.

## Test plan

- [x] `flyte gen docs --type markdown` produces per-option `####` sections after summary tables
- [x] `flyte run --help` and `flyte deploy --help` output unchanged
- [x] All 186 CLI tests pass (`pytest tests/cli/`)
- [ ] Regenerate `flyte-cli.md` in unionai-docs and verify rendered output


ppiegaze:
This is an attempt to add additional docs content to what is generated by `flyte gen docs` which is, in turn, used by the docs build to create the docs page for the Flyte CLI.

The idea is that this adds documentation content that is appropriate for published docs, but not for the CLI help on the command-line. Previous to this PR, the published docs were essentially a concatenation of the command line help.

The design chosen in this PR (by Claude) is to centralize the actual documentation strings in the _docs.py file as string constants. These are then sprinkled into the correct places in the functional CLI code.

I am not sure if this centralization is the right approach. Perhaps the docs content should reside directly beside the function that it describes.

Or, alternatively, maybe this stuff should simply be authored directly in the docs and I should figure out a way of integrating it with the generated CLI docs.

WDYT @kumare3 @wild-endeavor @pingsutw

samhita-alla (approved):

PR #970, opened 2026-04-14
block network access in the sandbox

samhita-alla:
* `block_network` is set to `False` by default in `ContainerTask`. 
* This param is also now available in the codegen plugin.
* Updates sandbox image hash to account for commands and python version.

### DIFF
--- a/src/flyte/extras/_container.py
+++ b/src/flyte/extras/_container.py
@@ -52,9 +52,9 @@ class ContainerTask(TaskTemplate):
     :param output_data_dir: The directory where the output data is stored. This is a string or a Path object.
     :param metadata_format: The format of the output file. This can be "JSON", "YAML", or "PROTO".
     :param local_logs: If True, logs will be printed to the console in the local execution.
-    :param block_network: If True (default), blocks all outbound network access. Locally this
+    :param block_network: If True, blocks all outbound network access. Locally this
         sets Docker ``network_mode=none``. On-cluster it applies the pod template
-        ``sandboxed-pod-template``. Set to False to allow network access.
+        ``sandboxed-pod-template``. Defaults to False.
     """
 
     MetadataFormat = Literal["JSON", "YAML", "PROTO"]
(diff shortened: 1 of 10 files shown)


pingsutw (approved):

src/flyte/extras/_container.py
```diff
@@ -68,8 +71,22 @@ def __init__(
...
     ):
+        if block_network:
+            existing = kwargs.get("pod_template")
+            if existing is None:
+                kwargs["pod_template"] = "sandboxed-pod-template"
+            elif isinstance(existing, str):
+                raise ValueError(
+                    "block_network=True cannot be combined with a string pod_template reference. "
+                    "Use a PodTemplate object instead so the 'sandboxed: true' label can be merged in, "
+                    "or ensure the referenced cluster template already includes that label."
```
kumare3, line 85: this is a weird error message? It should be simple, if you use block_network=True, custom pod_templates cannot be used. how can we merge it, what if they are not complimentary?

samhita-alla: it just requires adding a label, so assumed it shouldn't be too hard to just add a label to an existing pod template?

EngHabu (approved):

src/flyte/extras/_container.py
kumare3: aah ok, if its just a label then works. but why do we need backend then, why not add the label on the client side?

samhita-alla: yeah, @EngHabu suggested we add a pod template in the backend for consistency, but i agree it might be easier to manage in the SDK. wdyt @EngHabu?

EngHabu: Sorry what's the proposal? create a PodTemplate in the client instead that adds the Pod Label? or add Labels as a thing on TaskTemplate?
I thought having it in the back gives us flexibility in customizing it more (maybe we will add other restrictions in the future... etc.). Having a PodTemplate in the Client is okay. But what is the motivation to split this between BE and SDK? hardcoding the label and assuming it matches something in the back feels a bit off to me.

samhita-alla: this is regarding podtemplate. from an oss perspective, do we expect users to create a podtemplate themselves? otherwise, things would fail if it isn’t available in the backend. alternatively, if we provide a default podtemplate in the SDK, we can avoid abrupt failures. also do we expect users to define a networkpolicy or how should that work in oss?

EngHabu: I think we should deploy both by default in the helm chart...

samhita-alla: ^ cc @pingsutw

PR #971, opened 2026-04-14
feat: add disable_keyring config option to skip token caching

pingsutw:
## Summary
- Adds `admin.disableKeyring` YAML config option (and corresponding `FLYTE_ADMIN_DISABLEKEYRING` env var) to disable system keyring use for auth tokens
- Plumbs the flag through `PlatformConfig` → `init_from_config` → `init` → `_initialize_client` → `ClientSet` → `Authenticator` → `KeyringStore.retrieve/store/delete`
- When disabled, `KeyringStore` skips all store/retrieve/delete operations so tokens are never persisted to the system keyring

## Why
Useful for environments where the system keyring is unavailable (headless/CI), misconfigured, or undesirable for security reasons. Users can now opt out of token caching via config or env var without having to rely on side-channel hacks.

## Usage
YAML config:
```yaml
admin:
  endpoint: dns:///your-endpoint
  authType: Pkce
  disableKeyring: true
```

Env var:
```
export FLYTE_ADMIN_DISABLEKEYRING=true
```

## Test plan
- [x] Added unit tests in `tests/flyte/keyring/test_keyring_store.py` covering `store`/`retrieve`/`delete` with `disable=True` and `disable=False`
- [x] Existing keyring tests still pass
- [ ] Manually verify with a real endpoint that no tokens are written to the keyring when `disableKeyring: true` is set

### DIFF
--- a/src/flyte/config/_internal.py
+++ b/src/flyte/config/_internal.py
@@ -8,6 +8,7 @@ class Platform(object):
     CONSOLE_ENDPOINT = ConfigEntry(YamlConfigEntry("console.endpoint"))
     CA_CERT_FILE_PATH = ConfigEntry(YamlConfigEntry("admin.caCertFilePath"))
     HTTP_PROXY_URL = ConfigEntry(YamlConfigEntry("admin.httpProxyURL"))
+    DISABLE_KEYRING = ConfigEntry(YamlConfigEntry("admin.disableKeyring", bool))
 
 
 class Credentials(object):
--- a/src/flyte/_initialize.py
+++ b/src/flyte/_initialize.py
@@ -71,6 +71,7 @@ async def _initialize_client(
     client_credentials_secret: str | None = None,
     rpc_retries: int = 3,
     http_proxy_url: str | None = None,
+    disable_keyring: bool = False,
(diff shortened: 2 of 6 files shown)


src/flyte/config/_internal.py
```diff
@@ -8,6 +8,7 @@ class Platform(object):
     CONSOLE_ENDPOINT = ConfigEntry(YamlConfigEntry("console.endpoint"))
     CA_CERT_FILE_PATH = ConfigEntry(YamlConfigEntry("admin.caCertFilePath"))
     HTTP_PROXY_URL = ConfigEntry(YamlConfigEntry("admin.httpProxyURL"))
+    DISABLE_KEYRING = ConfigEntry(YamlConfigEntry("admin.disableKeyring", bool))
```
pmahindrakar-oss, line 11: This section cant be used for uctl because we wont have this option

pmahindrakar-oss (approved):

mhotan (approved):

src/flyte/config/_internal.py
pingsutw: We should turn off strict mode in uctl

PR #980, opened 2026-04-17
Include other files for fast-deploy in tasks and apps

kumare3:
## Summary

- Adds `include=(…)` on `TaskEnvironment` and `AppEnvironment` for shipping extra files (templates, configs, assets) alongside the auto-discovered code bundle on fast-deploy.
- Unifies bundling across `copy_style` (`loaded_modules` / `all` / `none` / `custom`) — includes are resolved relative to the env's declaring file and unioned into whichever path the builder takes.
- Adds canonical-layout tests under `tests/flyte/code_bundle/testdata/` and user-API tests covering include behavior + tuple hashability regression.

## Example

**TaskEnvironment** — ship an HTML template alongside the Python task:

```python
import flyte
import flyte.report
from pathlib import Path

env = flyte.TaskEnvironment(
    name="html_template_report",
    image=flyte.Image.from_debian_base(python_version=(3, 12)),
    include=("report_template.html",),  # resolved relative to this file
)

@env.task(report=True)
async def generate() -> str:
    template = (Path(__file__).parent / "report_template.html").read_text()
    flyte.report.get_tab("Main").log(template.format(title="Hello"))
    await flyte.report.flush.aio()
    return "ok"
```

**AppEnvironment** — ship a config file + a whole directory:

```python
from flyte.app import AppEnvironment

app_env = AppEnvironment(
    name="my-api",
    image="auto",
    port=8080,
    include=("config.yaml", "assets/"),  # directories expand recursively
)
```

Entries accepted: relative paths (anchored at the env's declaring file), absolute paths, directories (recursive), glob patterns.

## Precedence and combination with `copy_style`

Includes don't override anything — the bundler collects the **union** and dedupes by resolved absolute path.

| `copy_style`      | What's bundled                                                           |
|-------------------|--------------------------------------------------------------------------|
| `loaded_modules`  | imported Python modules under `from_dir` ∪ includes                      |
| `all`             | every non-ignored file under `from_dir` ∪ includes (include is no-op if already picked up) |
| `none`            | includes only (falls back to the relative-paths bundler)                 |
| `custom`          | caller-supplied relative paths ∪ includes                                |

Ordering rules:

1. **Within one env** — entries iterated in declaration order.
2. **Across envs** — iterated in deployment-plan order, deduped by resolved absolute path (first-seen wins).
3. **Against discovery** — if a file is already picked up by `copy_style`, the include is a silent no-op (same file on disk, can't differ).
4. **Final archive** — members are sorted lexicographically; include order is not observable in the tarball.
5. **Containment** — every resolved include must live under `from_dir`; otherwise the bundler raises `"outside the bundle root"`. Point `--root-dir` higher if you need a sibling file.


src/flyte/_code_bundle/_utils.py
```diff
@@ -123,6 +129,39 @@ def ls_files(
     else:
         all_files = list_all_files(source_path, deref_symlinks, ignore_group)
 
+    if additional_files:
+        resolved_source = source_path.resolve()
+        extra_paths: list[str] = []
+        for entry in additional_files:
+            p = pathlib.Path(entry)
+            if p.is_dir():
+                extra_paths.extend(str(child) for child in p.glob("**/*") if child.is_file())
```
pingsutw, line 138: should we do `extra_paths.extend(list_all_files(p, deref_symlinks, ignore_group))`? so it will respect the ignore files (gitignore, statndIgnore)

pingsutw (approved):

src/flyte/_code_bundle/_utils.py
kumare3: hmm good idea, but what if you want one of them. its ok i guess right

kumare3: should include be just raw? dont use it

PR #989, opened 2026-04-21
Add --gpu flag to `flyte start demo`

pingsutw:
## Summary
Adds `--gpu` to `flyte start demo`. When set, the underlying `docker run` is invoked with `--gpus all`, giving the demo container access to host NVIDIA GPUs.

```bash
flyte start demo --gpu --image ghcr.io/flyteorg/flyte-demo:gpu-latest
```

Default off — existing non-GPU users are unaffected.

## Why

The GPU-capable demo image being added in [flyteorg/flyte#7243](https://github.com/flyteorg/flyte/pull/7243) configures k3s, containerd, and the NVIDIA device plugin inside the image, but none of that matters if the host GPUs aren't passed through to the container. `--gpus all` on the `docker run` is the piece the CLI owns.

## Change
- `_start.py` — new `--gpu` click option on the `demo` command.
- `_demo.py` — threads `gpu: bool` through `launch_demo → _launch_demo_rich/plain → _run_step → _run_container`, which appends `--gpus all` when true.

## Test plan
Screenshot 2026-04-21 at 12 38 07 AM

- [x] Installed locally with `uv tool install --from . flyte`; `flyte start demo --help` lists the new flag.
- [x] `flyte start demo --gpu --image flyte-demo:gpu-local` on an A10G host launched the container with `--gpus all`, a Flyte task with `Resources(gpu=1)` ran `torch.cuda.is_available() == True`.
- [x] No regression expected on existing users (flag defaults to off).


src/flyte/cli/_demo.py
```diff
@@ -106,6 +107,8 @@ def _run_container(
         "--volume",
         f"{volume_name}:/var/lib/flyte/storage",
     ]
+    if gpu:
```
kumare3, line 110: i think we should add some unit tests for these? Also how about an alias devbox

pingsutw: > alias devbox

will change it in a separate PR

pingsutw: added tests

kumare3 (approved):

cosmicBboy (approved):

PR #992, opened 2026-04-21
add HF dataset plugin

samhita-alla:
This PR adds a Hugging Face datasets plugin for Flyte. It lets tasks consume Hub datasets directly via `from_hf(...)` and transparently materializes them as `datasets.Dataset` or `datasets.IterableDataset` with optional shared caching through `cache_root`.

```python
from flyteplugins.huggingface.datasets import from_hf

@env.task
async def train(
    ds: datasets.Dataset = from_hf(
        "stanfordnlp/imdb",
        name="plain_text",
        split="train",
        cache_root="s3://my-bucket/flyte-hf-cache",
    ),
) -> int:
    return len(ds)
```

The plugin streams converted parquet shards from the Hub, stores them as sharded parquet artifacts, and reuses cached materializations across runs when available.

You can also return and pass real `datasets.Dataset` objects between tasks:

```python
@env.task
async def create_dataset() -> datasets.Dataset:
    return datasets.Dataset.from_dict(
        {
            "text": ["hello", "world", "flyte"],
            "label": [0, 1, 0],
        }
    )

@env.task
async def filter_positive(ds: datasets.Dataset) -> datasets.Dataset:
    return ds.filter(lambda row: row["label"] == 1)
```

Screenshot 2026-04-24 at 3 24 30 PM


ursucarina (dismissed):

plugins/huggingface/pyproject.toml
```diff
@@ -0,0 +1,84 @@
+[project]
+name = "flyteplugins-huggingface"
+dynamic = ["version"]
+description = "Hugging Face Datasets plugin for flyte"
+readme = "README.md"
+authors = [{ name = "André Ahlert", email = "andre@aex.partners" }, { name = "Samhita Alla", email = "samhita@union.ai" }]
+requires-python = ">=3.10"
+dependencies = [
```
andreahlert, line 8: dropping hf-transfer is fine (HfFileSystem doesn't use it anyway), but the hf_hub_download fallback went with it. was killing the snapshot-download escapehatch intentional?

plugins/huggingface/src/flyteplugins/huggingface/dataset/__init__.py
```diff
@@ -0,0 +1,35 @@
+import functools
+
+from ._source import HFSource, from_hf
+from ._transformers import (
+    HFToHuggingFaceDatasetDecodingHandler,
+    HFToHuggingFaceIterableDatasetDecodingHandler,
+    HuggingFaceDatasetToParquetEncodingHandler,
+    HuggingFaceIterableDatasetToParquetEncodingHandler,
+    ParquetToHuggingFaceDatasetDecodingHandler,
+    ParquetToHuggingFaceIterableDatasetDecodingHandler,
```
andreahlert: triple registration, entry point + both __init__.pys. lru_cache saves it but any of the three can go, right?
```suggestion
```

plugins/huggingface/src/flyteplugins/huggingface/datasets/_io.py
```diff
@@ -0,0 +1,512 @@
...
+            raw = sorted(f for f in found if f.endswith(".parquet"))
+
+        if not raw:
+            return [join_uri_path(uri, f"{0:05}.parquet")]
+
+        if "://" in uri and "://" not in raw[0]:
+            proto = uri.split("://", maxsplit=1)[0] + "://"
+            raw = [f"{proto}{f}" for f in raw]
+        return raw
+    except Exception as e:
```
andreahlert, line 148: bare except hides real errors behind a fake 00000.parquet path. i think we could narrow it and log at warning. wdyt?

plugins/huggingface/src/flyteplugins/huggingface/datasets/_io.py
```diff
@@ -0,0 +1,512 @@
...
+async def ensure_hf_cached(source: HFSource) -> str:
+    """Return the remote path for source, fetching from HF if not cached."""
+    shards = typing.cast(
+        list[HFShard],
+        await run_sync_io("collect HF parquet shards", collect_hf_shards, source),
+    )
+    cache_key = hf_source_cache_key(source, shards)
+    expected_manifest = hf_cache_manifest(source, shards, cache_key)
+
+    if source.cache_root is None:
```
andreahlert, line 461: with cache_root=None the shards/sha256/manifest work runs and gets thrown away, every decode re-downloads to a new local dir. was cache-root meant to be strictly opt-in, or would defaulting to a raw_data remote path make sense?

plugins/huggingface/src/flyteplugins/huggingface/dataset/_io.py
```diff
@@ -0,0 +1,512 @@
...
+
+    default_remote_path = get_hf_cache_path(source, cache_key)
+    logger.info(
+        f"Checking Hugging Face dataset cache for {source.repo} "
+        f"({_source_log_description(source)}) "
+        f"under {source.cache_root}"
+    )
+    registry_record = await read_registry_record(source, cache_key)
+    remote_path = (
+        registry_record.get("artifact_uri", default_remote_path) if registry_record is not None else default_remote_path
```
andreahlert: artifact_uri only matters on hit, the miss branch resets to default_remote_path at 504. migration case in mind, but i really dont know.. maybe can the indirection come out

plugins/huggingface/src/flyteplugins/huggingface/datasets/_source.py
```diff
@@ -0,0 +1,135 @@
...
+        if not uri.startswith("hf://"):
+            raise ValueError(f"Invalid HF URI: {uri}")
+
+        path, _, query = uri[5:].partition("?")
+        repo = path.strip("/")
+        if not repo:
+            raise ValueError(f"Invalid HF URI: {uri}")
+
+        query_params = parse_qs(query)
+        name = query_params.get("name", [None])[0]
```
andreahlert, line 73: any reason to drop _validate_input_name? name/split flow unchecked into join_uri_path at 203, so ?name= ./../etc parses and only fails later with an opque hfs error

plugins/huggingface/src/flyteplugins/huggingface/datasets/_transformers.py
```diff
@@ -0,0 +1,289 @@
+from __future__ import annotations
+
+import pathlib
+import typing
+from pathlib import Path
+
+import datasets
```
andreahlert, line 7: imho back to lazy_module("datasets")? entry-point discovery triggers this and drags pyarrow/numpy/fsspec in for users who never touch the plugin. Is there any other reason i'm not seeing here?

plugins/huggingface/src/flyteplugins/huggingface/dataset/_transformers.py
```diff
@@ -0,0 +1,289 @@
...
+    col_lists = {name: col.to_pylist() for name, col in zip(batch.schema.names, batch.columns)}
+    return [{name: col_lists[name][i] for name in batch.schema.names} for i in range(batch.num_rows)]
+
+
+def _write_iterable_dataset(ds: datasets.IterableDataset, uri: str, filesystem) -> None:
+    file_idx = 0
+    rows_in_shard = 0
+    writer: pq.ParquetWriter | None = None
+
+    for batch in ds.iter(batch_size=10_000):
```
andreahlert: writer leaks if ds.iter() raises mid-stream, _write_dataset above wraps it in try/finally

plugins/huggingface/src/flyteplugins/huggingface/dataset/_transformers.py
```diff
@@ -0,0 +1,289 @@
...
+            )
+
+        filesystem = storage.get_underlying_filesystem(path=uri)
+        logger.info(
+            f"Reading Hugging Face Dataset parquet from "
+            f"{'remote' if storage.is_remote(uri) else 'local'} directory {uri}"
+        )
+
+        parquet_files = await list_parquet_files(uri, filesystem)
+        parquet_files = await _localize_parquet_files(parquet_files)
```
andreahlert: localizing full shards before pyarrow kills projection/row-group pushdown (pq.read_table(..., filesystem=fs, columns=...) range-requests only what it needs). was there a case where fsspec+pyarrow broke down for you? 

this one is not a problem. Just questioning to for learning purposes :)

andreahlert (commented):
a few questions. But after all, it has an overall better DX than my version. tks!

plugins/huggingface/examples/hf_dataset_workflow.py
```diff
@@ -0,0 +1,267 @@
...
+from collections import OrderedDict
+from pathlib import Path
+from typing import Annotated, Any, Callable
+
+import datasets
+import flyte
+from flyte._image import PythonWheels
+from flyte.io import DataFrame
+
+from flyteplugins.huggingface import from_hf
```
cosmicBboy: is there ever a world where we also want to support download models? In that case would be future-proof to rename this `dataset_from_hf` or expose a module `flyteplugins.huggingface.datasets.from_hf` so that we can also do `flyteplugins.huggingface.models.from_hf`

plugins/huggingface/pyproject.toml
samhita-alla: since we're using HFS, we don't need it anymore.

plugins/huggingface/src/flyteplugins/huggingface/dataset/__init__.py
samhita-alla: thanks for the catch! we have init registration as a fallback to support backward compatibility, but i think we can remove it in this case.

plugins/huggingface/src/flyteplugins/huggingface/datasets/_io.py
samhita-alla: opt-in for now as we don't support artifacts in v2 yet. even if we default to a raw_data remote path, it won't persist across runs. so it isn't really helpful. hope that makes sense.

plugins/huggingface/src/flyteplugins/huggingface/dataset/_io.py
samhita-alla: you're right. we don't really need `artifact_uri`. removed it.

plugins/huggingface/src/flyteplugins/huggingface/datasets/_source.py
samhita-alla: added a check.

plugins/huggingface/src/flyteplugins/huggingface/datasets/_transformers.py
samhita-alla: moved imports inside `register_huggingface_dataset_transformers()` to avoid eager imports. think that should do.

plugins/huggingface/src/flyteplugins/huggingface/dataset/_transformers.py
samhita-alla: fixed it.

plugins/huggingface/src/flyteplugins/huggingface/dataset/_transformers.py
samhita-alla: actually, i could use `fsspec` with `pyarrow`. i initially ran into remote storage auth-related errors, so i thought this would make things easier. but you’re right, i could use flyte’s storage abstraction to get the filesystem and read remote parquet files directly.

plugins/huggingface/examples/hf_dataset_workflow.py
samhita-alla: makes sense! it’s now exposed in the module so we avoid the risk of collisions, and each namespace owns its own registration.

pingsutw (approved):

plugins/huggingface/src/flyteplugins/huggingface/dataset/_transformers.py
andreahlert: understood! thank you

PR #1001, opened 2026-04-22
resolve CodeBundleLayer in flyte build

dansola:
## Problem
`flyte build` errors with `root_dir not set for CodeBundleLayer` for any image that uses `.with_code_bundle()`.

## Fix
`build_images()` now resolves `CodeBundleLayer` before building, mirroring what `apply()` already does on the deploy path. The CLI gains `--copy-style` and `--root-dir` (matching `flyte deploy`) so the resolution has the inputs it needs.

## Behavior
- Images without `.with_code_bundle()`: unchanged.
- Images with `.with_code_bundle()`: previously crashed, now build.
- Defaults match `flyte deploy`, so running `flyte build` then `flyte deploy` with no flags produces matching image hashes — deploy reuses the pre-built image instead of rebuilding.

## Why
Enables splitting image builds from deploys in CI, which is otherwise impossible today using .with_code_bundle()`.

### DIFF
--- a/src/flyte/_deploy.py
+++ b/src/flyte/_deploy.py
@@ -615,13 +615,22 @@ async def deploy(
 
 
 @syncify
-async def build_images(envs: Environment) -> ImageCache:
+async def build_images(envs: Environment, copy_style: "CopyFiles" = "loaded_modules") -> ImageCache:
     """
-    Build the images for the given environments.
+    Build the images for the given environment.
     :param envs: Environment to build images for.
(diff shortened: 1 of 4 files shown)


pingsutw (commented):
Thanks, looks good, could we add some tests for it

dansola:
> Thanks, looks good, could we add some tests for it

thanks, added tests @pingsutw!

kumare3 (approved):

PR #1006, opened 2026-04-23
add support for apiKeyEnvVar in config

cosmicBboy:


ursucarina (approved):

src/flyte/_initialize.py
```diff
@@ -409,6 +409,14 @@ async def init_from_api_key(
                 "API key must be provided either as a parameter or via the FLYTE_API_KEY environment variable.",
             )
 
+    api_key = api_key.strip()
+    if not api_key:
+        raise InitializationError(
+            "MissingApiKeyError",
+            "user",
+            "API key must be non-empty after stripping whitespace.",
```
pingsutw, line 417: Could we remove lines 406–409 since they’re duplicated?

pingsutw (commented):
Also, just curious, when would you use apiKeyEnvVar instead of the default FLYTE_API_KEY?

cosmicBboy:
Does the api key env var override the config file?

PR #1019, opened 2026-04-29
Support for RetryStrategy and Timeout types

### LINKED ISSUE #7307: Support for exp backoff in retries and various timeout knobs
Today's controls collapse three different failure questions into one knob:

- **"How long can this attempt run?"** — answered by `max_runtime`.
- **"How long am I willing to wait for this to start?"** — unanswerable. An action stuck in `QUEUED` or `WAITING_FOR_RESOURCES` waits forever (or until a parent timeout fires elsewhere).
- **"What is the total wall-clock budget for this work, including retries?"** — unanswerable. With `retries=5` and `max_runtime=1h`, an action can legally consume 6 hours plus queue time before it gives up.

Retries have the same flattening problem. `retries=N` means "try up to N+1 times back-to-back," with no way to say "wait 30s, then 1m, then 2m before retrying" — which is exactly the right policy for a flaky external API. Authors today either embed manual backoff inside the task body (which loses observability) or accept that retries will hammer a recovering downstream into the ground.

We want to support
```python
@env.task(
    retries=flyte.RetryStrategy(
        count=5,
        backoff=flyte.Backoff(base=timedelta(seconds=30), factor=2.0, cap=timedelta(minutes=5)),
    ),
    timeout=flyte.Timeout(
        max_runtime=timedelta(minutes=10),
        deadline=timedelta(hours=1),
    ),
)
async def bounded_retries():
    ...
```

kumare3:
- This makes it easier to control execution delays, performance and failures like queued times etc

Fixes: https://github.com/flyteorg/flyte/issues/7307

And related to: https://github.com/flyteorg/flyte/pull/7308

Example
Exponential backoff
```python
@env.task(
    retries=flyte.RetryStrategy(
        count=4,
        backoff=flyte.Backoff(
            base=timedelta(seconds=1),
            factor=2.0,
            cap=timedelta(seconds=3),
        ),
    ),
)
async def flaky_call() -> str:
```

Controllable timeouts
```python
@env.task(
    timeout=flyte.Timeout(
        max_runtime=timedelta(minutes=10),
        max_queued_time=timedelta(minutes=5),
        deadline=timedelta(minutes=30),
    ),
)
async def bounded_work() -> str:
```


kumare3:
To merge this PR the backend needs to be updated to handle the new error fields - cc @pingsutw

src/flyte/_internal/runtime/task_serde.py
```diff
@@ -93,6 +93,37 @@ def get_security_context(
...
+def _to_duration(value: timedelta | int | None) -> Optional[duration_pb2.Duration]:
+    """timedelta/int (seconds) -> google.protobuf.Duration; None passes through."""
+    if value is None:
+        return None
+    if isinstance(value, int):
+        value = timedelta(seconds=value)
+    total = value.total_seconds()
+    seconds = int(total)
+    nanos = round((total - seconds) * 1_000_000_000)
+    return duration_pb2.Duration(seconds=seconds, nanos=nanos)
```
pingsutw, line 105: we can use `duration_pb2.Duration().FromTimedelta(value)`

pingsutw (approved):

src/flyte/_internal/runtime/task_serde.py
kumare3: aah good idea let me do that

PR #1024, opened 2026-04-30
[testing] Local multi-processing executor

kumare3:
```flyte run --local-multi ...``` will now run locally across multiple processes.

```flyte.with_runcontext(mode="local-multi").run()``` does the same

### DIFF
--- a/src/flyte/_internal/controllers/__init__.py
+++ b/src/flyte/_internal/controllers/__init__.py
@@ -47,7 +47,7 @@ class TaskCallSequencer:
 if TYPE_CHECKING:
     import concurrent.futures
 
-ControllerType = Literal["local", "remote"]
+ControllerType = Literal["local", "local-multi", "remote"]
 
 R = TypeVar("R")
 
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: src/flyte/_internal/controllers/__init__.py [hunk 1]
```diff
@@ -47,7 +47,7 @@ def clear(self, action_key: str) -> None:
 if TYPE_CHECKING:
     import concurrent.futures
 
-ControllerType = Literal["local", "remote"]
+ControllerType = Literal["local", "local-multi", "remote"]
```
pingsutw, line 50: Would it be better to use multi-process when the user specifies local[4] (i.e., use 4 processes)?
Same as spark local mode. https://github.com/tomaztk/Spark-for-data-engineers/blob/main/SparkEngineers/04_Spark-Architecture-Local-and-cluster-mode.md#spark-architecture---local-and-cluster-mode

kumare3: I dont know if we can do that, as it maybe nested, or are you saying its parallel limit?

ranjit-parva: local[4] would be nice if we could

kumare3: the challenge is deadlock, if local[4] but, we need more to complete then how can we stop

pingsutw: yup, I mean parallel limit

pingsutw: i mean this
```
_DEFAULT_MAX_WORKERS = min(os.cpu_count() or 1, 8)
```

PR #1034, opened 2026-05-06
fix: propagate USER_LOG_LEVEL to remote task and serve environments

AdilFayyaz:
##  Motivation 
                                                                         
  `--user-log-level` affected local initialization only. Remote task containers and app environments fell back to `INFO` because `USER_LOG_LEVEL` was never forwarded in the env setup for `_run_remote` or `_serve_remote`. Additionally, connector code incorrectly imported the public `flyte.logger` instead of the internal one, routing SDK-internal logs through the user-facing logger.

 ## Summary of Changes

  - Add `user_log_level` parameter to `_Runner`, `with_runcontext`, `_Serve`, and `with_servecontext`; emit `USER_LOG_LEVEL` into the remote env alongside `LOG_LEVEL`                                                              
  - Fix `connectors/utils.py` and BigQuery, Databricks, and Snowflake plugin connectors to import from `flyte._logging import logger` instead of from `flyte import logger`, preserving internal log level and `[flyte] prefix`                                                                 
                                                                      
## Test Plan
                               
  - Three new async tests in `tests/flyte/test_serve.py` verify `_serve_remote` sets `USER_LOG_LEVEL` when provided, omits it when None, and that `user_log_level` takes precedence over an explicit value in `env_vars`                                                          
  - Run `uv run pytest tests/flyte/test_serve.py -k user_log_level` to exercise the new cases


src/flyte/_run.py
```diff
@@ -112,6 +112,7 @@ def __init__(
         interruptible: bool | None = None,
         log_level: int | None = None,
         log_format: LogFormat = "console",
+        user_log_level: int | None = None,
```
kumare3, line 115: why have 2? Why not just log_level

kumare3:
why not just log_level. If we need 2, then should we create system_log_level instead of user_log_level

src/flyte/_run.py
AdilFayyaz: The split was intentional. In #1027 we introduced flyte.logger as a user facing logger.
`log_level` controls the internal flyte logger while `user_log_level` controls the user facing one. 

I'd imagine users could want their logs at a different verbosity from what our system level ones are so keeping them separate made more sense to me. What do you think?

wild-endeavor:
i think having two log levels is valid, we want users to be able to see debug level stuff without turning on all sdk internal debug messages right?  but yes we can name the main one log_level, and the internal one system... this will change behavior for people but i think that's okay.  anyone relying on internal logs right now can change, but maybe something to call out in release notes.

plugins/bigquery/src/flyteplugins/bigquery/connector.py
```diff
@@ -4,7 +4,7 @@
 from dataclasses import dataclass
 from typing import Any, Dict, Optional
 
-from flyte import logger
+from flyte._logging import logger
```
pingsutw: Could we add system logger to init file? https://github.com/flyteorg/flyte-sdk/blob/47193cb55799b54d76b1050872ee3a9f6b46e344/src/flyte/__init__.py#L29

So we don't need to import private package in the plugins

pingsutw (approved):

PR #1038, opened 2026-05-07
feat(logging): stamp run/action context via LogRecordFactory

EngHabu:
## Summary

Replaces the per-logger `logging.Filter` approach for context stamping with a global `LogRecordFactory`. The factory runs on every `LogRecord` regardless of which logger created it, so any user-defined logger automatically picks up `run_name`, `action_name`, and `is_flyte_internal` attributes — no per-handler filter wiring required.

This also fixes a latent bug in main: the old `ContextFilter` and `FlyteInternalFilter` mutated `record.msg` in place, so a record passing through multiple handlers could get its `[run][action]` and `[flyte]` prefixes prepended multiple times. Stamping attributes on the record (and rendering them at format time) sidesteps that entirely.

## What changed

- **LogRecordFactory** (`src/flyte/_logging.py`) — stamps the active flyte action context on every record, lazily importing `flyte._context` to avoid import-time cycles.
- **`ContextFormatter`** replaces `ContextFilter` + `FlyteInternalFilter`. The `[flyte]` internal-prefix marker is now a formatter flag (`internal_prefix=True`), so the same record can be formatted by multiple handlers without prefixes compounding.
- **`ContextFormatter(inner=...)`** — optional kwarg to delegate base formatting to an existing `logging.Formatter`. Used to wrap pre-existing root-handler formatters without losing their layout.
- **`initialize_logger(reset_root_logger=False)`** now wraps each existing root-handler formatter with `ContextFormatter(inner=existing)`, preserving main's "third-party log lines through root get `[run][action]`" behavior. Idempotent on repeated init calls. Same temporal limitation as main: handlers added after `flyte.init()` won't be wrapped, but the factory still stamps the attrs so callers can format them themselves.
- **`get_rich_handler(internal_prefix=True)`** — opt-out parameter; the user-logger call site passes `False` so user-emitted lines don't carry the `[flyte]` marker even under rich logging.

## Behavior parity vs main

| | main | this PR |
|---|---|---|
| `[run][action]` on flyte/flyte.user logs | ✅ | ✅ |
| `[run][action]` on third-party logs through root (when handler exists at init) | ✅ (via filter) | ✅ (via formatter wrap) |
| `[run][action]` on third-party logs through root (handler added after init) | ❌ | ❌ — but attrs still on record |
| `[flyte]` marker on flyte internal logs | ✅ | ✅ |
| Compounding prefixes when record passes through multiple handlers | ❌ (latent bug) | ✅ fixed |
| `record.run_name` / `action_name` available everywhere | partial | ✅ universal (factory) |

## Why this matters for users

`flyte.logger` and any child of `flyte.user` continue to work as before. New: anyone using a stdlib logger (`logging.getLogger("myapp")`) outside the flyte namespace also gets `run_name` / `action_name` / `is_flyte_internal` stamped on every record by the factory — they just need a formatter that references them (e.g. `"[%(run_name)s][%(action_name)s] %(message)s"`).

## Example

`examples/basics/reuse_concurrent_logging.py` exercises three logger configs with a reusable container running concurrent tasks, so log output from multiple actions interleaves on a single stderr — each line correctly prefixed by its own `[run][action]`:

1. `flyte.logger` — canonical user-facing logger.
2. `logging.getLogger("flyte.user.myapp")` — child of `flyte.user`, inherits the formatter via propagation.
3. `logging.getLogger("myapp")` — fully independent logger with its own handler that references the factory-stamped attrs in its format string.

## Test plan

- [x] `pytest tests/user_api/test_logging.py tests/flyte/test_logging.py` passes (23 tests)
- [x] `test_user_logger_no_flyte_prefix` updated to check the formatter flag instead of the removed `FlyteInternalFilter`
- [x] `test_user_logger_no_flyte_prefix_after_rich_init` — regression test for rich-handler internal prefix bleeding into user logger
- [x] `test_initialize_logger_wraps_existing_root_handlers` — covers wrap path + idempotency on repeated init
- [ ] Manual: run `examples/basics/reuse_concurrent_logging.py` against a cluster and confirm all three worker variants emit `[run][action]`-prefixed log lines on the shared reused container


src/flyte/_logging.py
```diff
@@ -1,15 +1,73 @@
...
+    _LOG_CONTEXT_VARS.pop(name, None)
+
+
+_orig_record_factory = logging.getLogRecordFactory()
+
+
+def _flyte_record_factory(*args: Any, **kwargs: Any) -> logging.LogRecord:
+    record = _orig_record_factory(*args, **kwargs)
+
+    for attr, var in _LOG_CONTEXT_VARS.items():
```
AdilFayyaz: When we're iterating over `_LOG_CONTEXT_VARS.items()` with concurrent usage, can't this run into a race condition where one thread tried to update and another tries to read. Python wouldn't let you iterate over a dictionary that has changed its size. 
This may not be too noticeable since registration is done before the task even starts, but if someone were to dynamically register mid run this could be an issue.

AdilFayyaz:
```suggestion
    for attr, var in list(_LOG_CONTEXT_VARS.items()):
```

kumare3: good catch

kumare3: but, it happens on startup only right?

kumare3: I guess not, but this is python and so it will work fine :( GIL, but we should make it cleaner, have a lock

AdilFayyaz: Lock would be cleaner yes, or you could just snapshot by casting it as a list()

EngHabu: I got rid of that code altogether. feature creep.. adding custom context vars is out of scope :D

EngHabu: ditto

AdilFayyaz (approved):

PR #1054, opened 2026-05-11
refactor: remove repair_union_prev_checkpoint_uri SDK workaround

AdilFayyaz:
## Motivation

The backend bug that caused prev-checkpoint URIs to point to the current attempt directory instead of the prior one has been fixed (https://github.com/unionai/cloud/pull/15143). The SDK-side URI rewriting was always a temporary hack and can now be removed.
  
 ## Summary
  - Removed `repair_union_prev_checkpoint_uri` and the `_REMOTE_CHECKPOINT_SCHEMES` constant it depended on
  - Removed the call site in `Checkpoint.__init__` that patched remote prev-URIs on construction
  - Dropped the now-unused `re` import
  - Cleaned up the module docstring section describing the workaround

## Test Plan
  - Existing checkpoint tests cover the load/save paths — no behaviour changed for correct URIs
  - Verify end-to-end on a retried task: the prev-checkpoint URI provided by the executor should now be
  correct without SDK intervention

### DIFF
--- a/src/flyte/_checkpoint.py
+++ b/src/flyte/_checkpoint.py
@@ -14,12 +14,6 @@ open / open_sync in ``"wb"`` mode) so explicit `file://` checkpoint URIs work wi
 that downloads the previous attempt's **blob** into a temp workspace and uploads a new blob to the
 task output prefix (including tarball encoding when you save a directory).
 
-- **Previous-checkpoint URI repair:** For remote prev URIs (e.g. `s3://`), if the path uses the **current**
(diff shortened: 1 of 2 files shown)


cosmicBboy (dismissed):
🔥 thanks for fixing this @AdilFayyaz !

cosmicBboy:
looks like `repair_union_prev_checkpoint_uri` tests need to be deleted

cosmicBboy (approved):

PR #1058, opened 2026-05-12
fix: fall back to ASCII spinner when stdout can't encode unicode

EngHabu:
## Summary
- Rich's default \"dots\" spinner uses braille pattern characters (e.g. `'⠙'`). On legacy Windows consoles where `sys.stdout.encoding` is cp1252, writing those characters crashes the CLI with `UnicodeEncodeError` mid-render — Sentry has observed this taking down `flyte deploy` before any real work happens.
- Add a `safe_spinner()` helper in `flyte.cli._common` that probes `sys.stdout.encoding` and falls back to Rich's ASCII-only `\"line\"` spinner when the requested spinner's characters can't be encoded. Apply it in `cli_status` and at the direct `console.status(...)` callsites that explicitly request `\"dots\"`.

## Sentry issues
Fixes [FLYTE-SDK-A](https://unionai.sentry.io/issues/7476377068/) (`UnicodeEncodeError: 'charmap' codec can't encode character '⠙'`)

## Test plan
- [x] New `tests/flyte/cli/test_safe_spinner.py` covers utf-8 (keeps \"dots\"), cp1252 / ascii / missing encoding (falls back to \"line\")
- [x] `make fmt` clean

### DIFF
--- a/src/flyte/cli/_abort.py
+++ b/src/flyte/cli/_abort.py
@@ -24,7 +24,7 @@ def run(cfg: common.CLIConfig, run_name: str, reason: str, project: str | None =
     r = remote.Run.get(name=run_name)
     if r:
         console = common.get_console()
-        with console.status(f"Aborting run '{run_name}'...", spinner="dots"):
+        with console.status(f"Aborting run '{run_name}'...", spinner=common.safe_spinner("dots")):
             r.abort(reason=reason)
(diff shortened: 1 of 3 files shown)


EngHabu (commented):
LGTM

pingsutw (approved):
LGTM

PR #1072, opened 2026-05-15
fix(config): escape bracket in read_file_if_exists debug log

jeevb:
The debug log embedded the file path inside square brackets. Under rich>=15 the markup parser strictly rejects `[/path]` as a closing tag with no matching opener (RichHandler is configured with markup=True in flyte._logging), so any debug-level CLI run that read a config file crashed with MarkupError. Escape the opening `\[` so Rich renders the brackets literally while non-rich handlers stay readable.

Add a regression test that installs a RichHandler against the flyte logger and asserts read_file_if_exists logs cleanly with the brackets preserved in the rendered output.


AdilFayyaz (approved):
LGTM

PR #1073, opened 2026-05-15
fix: raise actionable error when module file is outside root_dir

EngHabu:
## Summary
- `load_python_modules` derives the dotted module name from a `.py` file path via `Path.resolve().relative_to(root_dir)`. If the user runs `flyte deploy path/to/file.py` on a file that lives **outside** the configured project root (commonly: src-layout projects where the SDK inferred `src/` but the file is elsewhere in the project), `pathlib` raises `ValueError: '...' is not in the subpath of '...'` and the raw stack lands in Sentry as an unhandled SDK crash.
- Wrap the `relative_to` call in a `_relative_to_root` helper that translates the `ValueError` into a `click.ClickException`. The user gets an actionable message pointing at the most common cause (forgot `--root-dir`), and the exception short-circuits the Sentry filter for click errors so we stop logging this as an SDK crash.

## Sentry issues
Fixes [FLYTE-SDK-2T](https://unionai.sentry.io/issues/7481668541/) (`ValueError: '/Users/.../examples/basics/multi_status.py' is not in the subpath of '/Users/.../flyte-sdk/src'`)

## Test plan
- [x] New regression test `test_load_python_modules_file_outside_root_raises_click_exception` reproduces the Sentry crash shape and asserts a `click.ClickException` with the actionable message is raised
- [x] Existing `tests/flyte/utils/test_module_loader.py` suite still passes
- [x] `make fmt` clean

### DIFF
--- a/src/flyte/_utils/module_loader.py
+++ b/src/flyte/_utils/module_loader.py
@@ -5,10 +5,33 @@ from pathlib import Path
 from types import ModuleType
 from typing import List, Tuple
 
+import click
+
 import flyte.errors
 from flyte._code_bundle._ignore import GitIgnore, IgnoreGroup, StandardIgnore
 from flyte._constants import FLYTE_SYS_PATH
 from flyte._logging import logger
+
+
+def _relative_to_root(path: Path, root_dir: Path) -> Path:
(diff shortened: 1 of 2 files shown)


EngHabu (commented):
LGTM

pingsutw (approved):

PR #1074, opened 2026-05-15
fix: show distinct sys.modules keys in duplicate-env error message

EngHabu:
## Summary
- When the same file gets loaded twice under different import paths (the classic src/ layout dual-import scenario), `_check_duplicate_env` told the user the env was *"imported twice under different module names ('multi_status' and 'multi_status')"* — the two displayed names were identical, defeating the entire purpose of the message.
- Root cause: `_find_env_module` returned the module object, and the error formatter read `module.__name__`. But when the SDK's path-based loader creates a module via `importlib.util.spec_from_file_location(file_path.stem, ...)`, both module objects end up with the *same* `__name__` (the file stem), even though they sit at different `sys.modules` keys.
- Switch `_find_env_module` to iterate `sys.modules.items()` and return the `(key, module)` tuple. The error message now uses the sys.modules keys — the distinguishing import paths the user can actually act on.

## Sentry issues
Fixes [FLYTE-SDK-2S](https://unionai.sentry.io/issues/7481668212/) (`ValueError: Environment 'multi_status_demo' ... imported twice under different module names ('multi_status' and 'multi_status')`)

## Test plan
- [x] New regression test `test_check_duplicate_env_dual_import_shows_distinct_sys_modules_keys` constructs two modules that intentionally share `__name__` but sit at different sys.modules keys (the exact shape from the Sentry crash) and asserts the error message shows both distinct keys
- [x] All existing `_check_duplicate_env` / dual-import tests still pass
- [x] `make fmt` clean

### DIFF
--- a/src/flyte/_deploy.py
+++ b/src/flyte/_deploy.py
@@ -533,23 +533,31 @@ async def apply(deployment_plan: DeploymentPlan, copy_style: CopyFiles, dryrun: 
 
 
 def _find_env_module(env: Environment):
-    """Scan sys.modules to find the module that contains this env as a top-level variable."""
-    for module in list(sys.modules.values()):
+    """Scan sys.modules to find the (sys.modules key, module) that contains this env as a top-level variable.
+
+    Iterates ``sys.modules.items()`` rather than ``.values()`` so callers can show the *import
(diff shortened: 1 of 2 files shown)


EngHabu:
LGTM

AdilFayyaz (approved):

PR #1075, opened 2026-05-15
fix: surface missing src path as ImageBuildError in copy_files_to_context

EngHabu:
## Summary
- `copy_files_to_context` is called for every image layer that needs a file or directory inside the Docker build context (requirements files, pyproject/uv.lock, source folders, etc.). It used to delegate the existence check to `shutil.copy` / `shutil.copy2` and let `FileNotFoundError` bubble all the way up. When the user pointed a layer at a path that didn't exist, the raw traceback ended up in Sentry as an unhandled SDK crash even though it's purely a configuration mistake on the user side.

Two changes:
1. Guard the entrypoint with `src.exists()` and raise `ImageBuildError` with an actionable message naming the missing path and the most likely culprits (`with_requirements`, `with_source_folder`, `with_pyproject`). `ImageBuildError` is already filtered from Sentry via `_is_user_error` (#1052), so we stop logging this as an SDK crash.
2. Inside the directory branch, defensively catch `FileNotFoundError` from `shutil.copy2` so that a file that disappears between enumeration (`pm.walk`) and the copy (e.g. a transient venv entry, a broken symlink, or a file removed mid-build) only generates a warning rather than aborting the entire image build.

## Sentry issues
Fixes [FLYTE-SDK-2X](https://unionai.sentry.io/issues/7481752765/) (`FileNotFoundError: [Errno 2] No such file or directory: '/Users/.../flyte-sdk/.venv/lib/python3.13/dist'` from `shutil.copyfile` inside `copy_files_to_context`)

## Test plan
- [x] New regression test `test_copy_files_to_context_missing_src_raises_image_build_error` covers the missing-src branch
- [x] New regression test `test_copy_files_to_context_skips_files_that_vanish_mid_walk` covers the vanish-mid-walk branch
- [x] Existing `tests/flyte/imagebuild/test_utils.py` suite still passes
- [x] `make fmt` clean

### DIFF
--- a/src/flyte/_internal/imagebuild/utils.py
+++ b/src/flyte/_internal/imagebuild/utils.py
@@ -8,6 +8,7 @@ from typing import List, Optional
 from flyte._code_bundle._ignore import STANDARD_IGNORE_PATTERNS
 from flyte._image import DockerIgnore, Image
 from flyte._logging import logger
+from flyte.errors import ImageBuildError
 
 
 def copy_files_to_context(src: Path, context_path: Path, ignore_patterns: list[str] = STANDARD_IGNORE_PATTERNS) -> Path:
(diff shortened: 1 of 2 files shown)


EngHabu:
LGTM

pingsutw (approved):

PR #1078, opened 2026-05-16
fix: walk exception cause chain in Sentry user-error filter

EngHabu:
## Summary

`_upload_single_file` (and other call sites) wrap the underlying `ConnectError` / `AuthenticationError` / `RuntimeError("SelectCluster failed ...")` in a `RuntimeSystemError("Failed to get signed url ...")`. The outer `RuntimeSystemError` is not in `_is_user_error`'s set, so even though the chained cause is a user-facing problem (expired refresh token, expired device code, IDP rejection, transport timeout), the wrapped exception lands in Sentry as an unhandled SDK crash.

Two changes:

1. **`_iter_cause_chain`** — walk `__cause__` / `__context__` with a `seen` set and a depth cap of 16 (cycle-safe, won't spin even if a downstream bug wires a cycle).
2. **`_is_user_error`** — classify against the cause chain rather than just the outermost exception, and add `AuthenticationError` / `AccessTokenNotFoundError` (`flyte.remote._client.auth.errors`) to the user-error set. This also makes the existing `DeploymentError` / `ImageBuildError` / `InitializationError` / click filters robust to wrapping further up the stack.

## Sentry issues

These all share the same outer title (`RuntimeSystemError: Failed to get signed url for ...`) but the chained cause varies — this PR filters them all out at the Sentry boundary:

- [FLYTE-SDK-2A](https://unionai.sentry.io/issues/7475986722/) — cause: `AuthenticationError`/`RuntimeError(invalid_grant ... refresh token expired)`
- [FLYTE-SDK-2P](https://unionai.sentry.io/issues/7479057871/) — cause: `AuthenticationError(device code has expired)`
- [FLYTE-SDK-29](https://unionai.sentry.io/issues/7475758083/) — cause: `TimeoutError` / `ConnectError(Request timed out)`
- [FLYTE-SDK-31](https://unionai.sentry.io/issues/7483511296/) — cause: `ConnectTimeout`

`FLYTE-SDK-2H` (cause: server-side validation `project_id.organization length must be at least 1`) is **not** classified as a user error by this change — it's a config gap on the user's side but the SDK should produce a clearer error before hitting the server. Tracked separately.

## Test plan

- [x] New tests in `tests/flyte/test_sentry.py`:
  - wrapped `AuthenticationError` 3-deep chain (the exact shape from FLYTE-SDK-2P) is filtered
  - wrapped `AccessTokenNotFoundError` is filtered
  - wrapped `DeploymentError` (via `__cause__`) is filtered
  - unrelated `RuntimeError` still goes to Sentry (no regression on real crashes)
  - cycle in `__cause__` chain doesn't hang `_iter_cause_chain`
- [x] All 13 sentry tests pass
- [x] `make fmt` clean
- [x] Commit signed (`-s`)

### DIFF
--- a/src/flyte/_sentry.py
+++ b/src/flyte/_sentry.py
@@ -60,30 +60,57 @@ def init() -> None:
         logger.debug("Failed to initialize Sentry", exc_info=True)
 
 
+def _iter_cause_chain(exc: BaseException):
+    """Walk __cause__ / __context__ chain so wrapping doesn't hide the real type.
+
+    Bounded depth - exception chains in the wild stay shallow (3-5 deep), but a
+    bug elsewhere could create a cycle and we don't want this to spin.
+    """
+    seen: set[int] = set()
+    cur: BaseException | None = exc
+    depth = 0
+    while cur is not None and id(cur) not in seen and depth < 16:
+        yield cur
+        seen.add(id(cur))
+        nxt = cur.__cause__ or cur.__context__
+        cur = nxt
+        depth += 1
+
+
 def _is_user_error(exc: BaseException) -> bool:
     """Errors raised intentionally as user-facing messages — not crash reports."""
     try:
         import click
 
-        if isinstance(exc, (click.Abort, click.exceptions.Exit, click.ClickException)):
-            return True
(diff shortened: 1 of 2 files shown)


EngHabu (commented):
LGTM

pingsutw (approved):

PR #1080, opened 2026-05-18
fix(controller): guard against None responses in informer watch loop

EngHabu:
When gRPC deserialization fails, grpc._common._transform() logs the exception and returns None instead of raising.  In the informer's watch loop this caused an AttributeError on resp.control_message, which crashed the loop and triggered a retry/backoff cycle.  During that backoff window the trace-action completion event was never received, so wait_for_action_completion() blocked forever.

Two fixes:
1. _informer.py: skip None responses with a warning so the stream stays alive and can receive the real completion update on the next message.
2. _core.py: add a 60-second timeout on wait_for_action_completion() for trace actions only (task actions may run for hours).  On timeout we fire the completion event ourselves so the caller is unblocked even if the server update was genuinely missed.  The timeout is configurable via _F_TRACE_COMPLETION_TIMEOUT.

### DIFF
--- a/src/flyte/_internal/controllers/remote/_core.py
+++ b/src/flyte/_internal/controllers/remote/_core.py
@@ -335,9 +335,23 @@ class Controller:
             )
 
         logger.debug(f"{threading.current_thread().name} Waiting for completion of {action.name}")
-        # Wait for completion
+        # Wait for completion.  For trace actions apply a timeout so a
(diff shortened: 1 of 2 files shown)


EngHabu (commented):
LGTM

pingsutw (approved):

PR #1092, opened 2026-05-20
feat: add clustered task pod entrypoint module

AdilFayyaz:
## Motivation

 JobSet-based distributed training pods need a bootstrapper that runs before torchrun — deriving master address from Kubernetes-injected env vars, waiting for DNS, and handing off to torchrun with the correct node rank and rendezvous ID. This is essentially a Python module that ships in the SDK alongside a0.

## Summary

  - Adds `flyte/distributed/_entrypoint.py` — runs as PID 1 via `python -m flyte.distributed._entrypoint`, derives `MASTER_ADDR` from `JOBSET_NAME + POD_NAMESPACE`, busy-waits for DNS (5-min timeout), then `os.execvp("torchrun", [...])`
  - `RDZV_ID` is rotated on each JobSet restart (`JOBSET_NAME-JOBSET_RESTART_ATTEMPT`) so torchrun does a fresh rendezvous rather than reconnecting to a stale one
  - Adds `flyte/distributed/__init__.py` as a package stub — public API (`ClusteredTaskEnvironment` etc.) added in PR 5. This is PR 2 of the plan.
  - 7 unit tests covering happy path, DNS timeout, missing env vars, RDZV_ID rotation, NODE_RANK mapping, and MASTER_PORT default (29500)

## Test Plan

  - `cd flyte-sdk && .venv/bin/pytest tests/flyte/distributed/test_entrypoint.py -v` — all 7 tests pass
  - `python -c "import flyte.distributed._entrypoint"` — module is importable with no side effects


src/flyte/distributed/_entrypoint.py
```diff
@@ -0,0 +1,100 @@
...
+    "NNODES",
+    "NPROC_PER_NODE",
+    "RDZV_BACKEND",
+]
+
+
+def _require(var: str) -> str:
+    val = os.environ.get(var)
+    if not val:
+        print(f"[clustered-entrypoint] ERROR: required env var {var!r} is not set", flush=True)
```
pingsutw: should we use logger here

src/flyte/distributed/_entrypoint.py
```diff
@@ -0,0 +1,100 @@
...
+def _master_addr(jobset_name: str, namespace: str) -> str:
+    return f"{jobset_name}-workers-0-0.{jobset_name}.{namespace}.svc.cluster.local"
+
+
+def _wait_for_dns(hostname: str, timeout: float = _DNS_TIMEOUT_SEC, interval: float = _DNS_RETRY_INTERVAL_SEC) -> None:
+    deadline = time.monotonic() + timeout
+    while True:
+        try:
+            socket.getaddrinfo(hostname, None)
+            print(f"[clustered-entrypoint] DNS resolved: {hostname}", flush=True)
```
pingsutw: ditto

pingsutw (approved):

PR #675, opened 2026-02-16
PR into #488

wild-endeavor:
* Bump flyteidl2 version to latest and get rid of the manually copied code
* Fix the pem issue where we had to manually download the cert from amazon before creating the channel (with tonic connections you have to call `with_native_roots()` to pick up root certs).
* Fix maturin setting so it'll build.
* Pull in some changes that have been made to the main controller. (#607, #521, #621)
* Add in some CI as a separate PR - https://github.com/flyteorg/flyte-sdk/pull/676.  But we need to play with this experience a bit more first.  Need to go through the whole experience and see how things feel.

* More changes described in https://github.com/flyteorg/flyte-sdk/pull/675#issuecomment-4388680320

PR #1093, opened 2026-05-21
[Controller] Support trace in rust controller

machichima:
## Related PR

Follow up on https://github.com/flyteorg/flyte-sdk/pull/675#issuecomment-4388680320 to support traces in rust controller

## Description

Support traces in rust controller. 

Tested on dogfood to ensure things work: https://dogfood-gcp.cloud-staging.union.ai/v2/domain/development/project/flytesnacks/runs/urqdqb7g56vf89g4dkrk

Also test with hello_v2 example to ensure normal task path does not break: https://dogfood-gcp.cloud-staging.union.ai/v2/domain/development/project/flytesnacks/runs/urdr86c7pmtngs5zw8bs

### DIFF
--- a/rs_controller/src/action.rs
+++ b/rs_controller/src/action.rs
@@ -297,7 +297,7 @@ impl Action {
             realized_outputs_uri: Some(outputs_uri),
             phase: ActionPhase::Succeeded.into(),
             err: None,
-            started: true,
+            started: false,
             retries: 0,
             client_err: None,
             cache_key: None,
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -3,8 +3,6 @@ name: Publish
 on:
   release:
     types: [published]
-  push:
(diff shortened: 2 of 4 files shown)


rs_controller/src/action.rs
```diff
@@ -297,7 +297,7 @@ impl Action {
             realized_outputs_uri: Some(outputs_uri),
             phase: ActionPhase::Succeeded.into(),
             err: None,
-            started: true,
+            started: false,
```
pingsutw, line 300: why do we set it to false?

pingsutw (approved):

rs_controller/src/action.rs
machichima: If we set `started` to true here, it will never be enqueued

https://github.com/flyteorg/flyte-sdk/blob/eca8122e271240be7ba745a92b00fcdb7ef4295b/rs_controller/src/core.rs#L798-L813

PR #1105, opened 2026-05-22
Enable Auth Metadata Local Caching in Flyte-SDK

edwardsun-union:
### - Description
Based on the decision make in eng slack channel.  We have decided to enable auth metadata local cache capability in order to reduce the unnecessary `GetPublicClientConfig` API calls currently existing when executing common tasks such as `flyte run` workflow.   As part of the decision, the cache will be stored within the hidden `.flyte` directory under home based on the OS.  

### - Testing
before change:
1.  In local env,  create a new config. Example:
`flyte create config --endpoint dns:///dogfood.cloud-staging.union.ai --project edward-test --domain staging --image-builder remote`
2. executing flyte task using the create config.  Example test command below:
`flyte -vvv -c ./.flyte/config.yaml run ./basics/types/int_collection.py main --numbers '[1,2,3,4,5,10,20,30]'`
For client configuration refer to dogfood staging config v2
3. `GetPublicClientConfig` call occurs for every task run

after change:
1.  repeat the same test steps.
2.  cache file contains auth metadata is generated and stored with all required auth metadata    
3.  Repeat the same flyte task run
4.  observed that GetPublicClientConfig API call is not made from Grafana (To differentiate traffic from local env, temporarily bump up connectrpc minor version to such as `connectrpc/0.10.1`)


src/flyte/cli/_create.py
```diff
@@ -318,10 +318,18 @@ def config(
     if endpoint:
         endpoint = sanitize_endpoint(endpoint)
         admin["endpoint"] = endpoint
-    if insecure:
+    if image_builder == "remote":
+        admin["authType"] = common.sanitize_auth_type(auth_type) if auth_type else "Pkce"
+        admin["clientId"] = "dogfood-uctl"
```
pmahindrakar-oss: This seems incorrect to hardcode

src/flyte/config/_internal.py
```diff
@@ -9,6 +9,10 @@ class Platform(object):
     CA_CERT_FILE_PATH = ConfigEntry(YamlConfigEntry("admin.caCertFilePath"))
     HTTP_PROXY_URL = ConfigEntry(YamlConfigEntry("admin.httpProxyURL"))
     DISABLE_KEYRING = ConfigEntry(YamlConfigEntry("admin.disableKeyring", bool))
+    AUTHORIZATION_HEADER = ConfigEntry(YamlConfigEntry("admin.authorizationHeader"))
+    REDIRECT_URI = ConfigEntry(YamlConfigEntry("admin.redirectUri"))
+    REDIRECT_URL = ConfigEntry(YamlConfigEntry("admin.redirectURL"))
```
pmahindrakar-oss: Do we need both here ?

src/flyte/remote/_client/auth/_authenticators/client_credentials.py
```diff
@@ -51,14 +51,18 @@ async def _do_refresh_credentials(self) -> Credentials:
         """
         cfg = await self._resolve_config()
 
+        if cfg.token_endpoint is None:
+            raise ValueError("token_endpoint is required for client credentials authentication")
```
pmahindrakar-oss: Wont this break existing clients who haven't generated there new configs. Ideally we want to keep it backward compat

src/flyte/cli/_create.py
edwardsun-union: Good caught, Fixed this

src/flyte/config/_internal.py
edwardsun-union: @pmahindrakar-oss removed

src/flyte/remote/_client/auth/_authenticators/client_credentials.py
edwardsun-union: I addressed this by keeping ClientConfig untouched and pack my changes into LocalClientConfigOverrides, this way is cleaner and ensure backward compatibility for existing client

src/flyte/config/_config.py
```diff
@@ -108,6 +113,12 @@ def auto(cls, config_file: typing.Optional[typing.Union[str, ConfigFile]] = None
 
         kwargs = set_if_exists(kwargs, "http_proxy_url", _internal.Platform.HTTP_PROXY_URL.read(config_file))
         kwargs = set_if_exists(kwargs, "disable_keyring", _internal.Platform.DISABLE_KEYRING.read(config_file))
+        kwargs = set_if_exists(
+            kwargs, "authorization_header", _internal.Platform.AUTHORIZATION_HEADER.read(config_file)
+        )
+        redirect_uri = _internal.Platform.REDIRECT_URI.read(config_file)
```
pmahindrakar-oss: this var is not used other that in L120. Can use similar pattern to L121

pyproject.toml
```diff
@@ -110,6 +110,7 @@ dev = [
     "mcp>=1.26.0",
     "starlette",
     "uvicorn",
+    "debugpy>=1.8.20",
```
pmahindrakar-oss, line 113: Do we need this dependency

src/flyte/_initialize.py
```diff
@@ -357,6 +390,8 @@ async def init_from_config(
         proxy_command=cfg.platform.proxy_command,
         client_id=cfg.platform.client_id,
         client_credentials_secret=cfg.platform.client_credentials_secret,
+        local_client_config_overrides=local_client_config_overrides,
+        audience=audience,
```
pmahindrakar-oss, line 402: Any reason to keep audience different from local_client_config_overrides

src/flyte/remote/_client/auth/_client_config.py
```diff
@@ -39,6 +39,21 @@ def with_override(self, other: "ClientConfig") -> "ClientConfig":
...
+    """
+
+    redirect_uri: typing.Optional[str] = None
+    client_id: typing.Optional[str] = None
+    scopes: typing.Optional[typing.List[str]] = None
+    header_key: typing.Optional[str] = None
+    audience: typing.Optional[str] = None
+
+    def has_required_public_client_fields(self) -> bool:
+        return bool(self.client_id and self.redirect_uri and self.header_key and self.scopes)
```
pmahindrakar-oss, line 64: are we missing audience

src/flyte/cli/_create.py
```diff
@@ -318,14 +318,20 @@ def config(
...
-    if insecure:
-        admin["insecure"] = insecure
-    if auth_type:
-        admin["authType"] = common.sanitize_auth_type(auth_type)
 
     if not org and endpoint:
         org = org_from_endpoint(endpoint)
 
+    if not org:
+        raise click.BadParameter("--endpoint or --org must be provided")
```
pmahindrakar-oss: Check if this could break for local-persistence mode. 
```
--local-persistence enables a local SQLite-backed record of local runs. The data comes from the SDK process itself while it executes locally: task/
  run metadata, inputs/outputs references, status, timestamps, etc. It is written to local storage, then tools like the local TUI can read it back.
```

src/flyte/cli/_create.py
```diff
@@ -318,14 +318,20 @@ def config(
...
+
+    admin["authType"] = common.sanitize_auth_type(auth_type) if auth_type else "Pkce"
+    admin["clientId"] = f"{org}-uctl"
+    admin["insecure"] = insecure
+    admin["authorizationHeader"] = "flyte-authorization"
+    admin["redirectUri"] = "http://localhost:53593/callback"
+    admin["scopes"] = ["all"]
```
pmahindrakar-oss: We want to make sure these defaults are ok to be used in selfhosted environments. 
cc : @mhotan .

pmahindrakar-oss: I think as an escape hatch , can we allow these to be passed, for eg in cases of selfhosted or where there are deviations in these defaults. An explicit args would help here.

pyproject.toml
edwardsun-union: @pmahindrakar-oss yes, it is needed to enable breakpoints and it is dev dependency only

src/flyte/cli/_create.py
pmahindrakar-oss: Thinking about this more, we can populate this from GetPublicClientConfig call so we remove these unknowns. We only call this API on create config call and once that data exists in cli's config, it doesn't need to refetch it

src/flyte/remote/_client/auth/_client_config.py
edwardsun-union: audience is not a required metadata for auth to succeed,  I verified this in staging

src/flyte/cli/_create.py
```diff
@@ -308,24 +313,49 @@ def config(
...
-            return
-
     admin: Dict[str, Any] = {}
     if endpoint:
         endpoint = sanitize_endpoint(endpoint)
         admin["endpoint"] = endpoint
-    if insecure:
-        admin["insecure"] = insecure
-    if auth_type:
-        admin["authType"] = common.sanitize_auth_type(auth_type)
```
pmahindrakar-oss, line 324: Any reason we are removing these

src/flyte/remote/_client/auth/_public_client_cache.py
```diff
@@ -0,0 +1,151 @@
...
+        )
+
+
+def get_public_client_auth_metadata_cache_root(cache_root: Path | None = None) -> Path:
+    return cache_root or Path.home() / ".flyte" / ".cache"
+
+
+def get_public_client_auth_metadata_cache_path(org: str, domain: str, cache_root: Path | None = None) -> Path:
+    resolved_root = get_public_client_auth_metadata_cache_root(cache_root)
+    return resolved_root / f"{org}-{domain}.yaml"
```
pmahindrakar-oss: org.yaml should be sufficient as org is guaranteed to be unique in an environment

src/flyte/cli/_create.py
```diff
@@ -308,24 +313,49 @@ def config(
...
 
     if not org and endpoint:
         org = org_from_endpoint(endpoint)
 
+    if endpoint and not org:
+        raise click.BadParameter("--endpoint or --org must be provided")
+
+    if endpoint and not domain:
+        raise click.BadParameter("--domain must be provided")
+
```
pmahindrakar-oss: Why do we need domain parameter

pyproject.toml
pmahindrakar-oss, line 113: Is this needed

src/flyte/cli/_create.py
edwardsun-union: fixed

src/flyte/remote/_client/auth/_public_client_cache.py
edwardsun-union: addressed this to use endpoint based on offline discussion in slack

src/flyte/cli/_create.py
edwardsun-union: fixed

pyproject.toml
edwardsun-union: this is for enabling breakpoint to help debugging, added as dev dependency only, should be no harm

src/flyte/cli/_create.py
edwardsun-union: Tested this should be handled the same as prior change, also added test for it

src/flyte/remote/_client/auth/_public_client_cache.py
```diff
@@ -0,0 +1,156 @@
...
+        insecure: bool = False,
+    ) -> "CachedPublicClientAuthMetadata":
+        return cls(
+            authType=auth_type,
+            clientId=public_client_config.client_id,
+            insecure=insecure,
+            authorizationHeader=public_client_config.authorization_metadata_key,
+            redirectUri=public_client_config.redirect_uri,
+            scopes=list(public_client_config.scopes),
+            audience=public_client_config.audience,
```
edwardsun-union, line 58: @pmahindrakar-oss the audience is already refreshed from PublicClientConfig Call here

PR #1106, opened 2026-05-22
error handling

wild-endeavor:


src/flyte/_internal/runtime/convert.py
```diff
@@ -337,6 +349,7 @@ def convert_from_native_to_error(err: BaseException) -> Error:
                 code=err.code,
                 message=str(err),
                 worker=err.worker,
+                recoverability=execution_pb2.ContainerError.RECOVERABLE,
```
kumare3, line 352: is that what we want to do? should it be recoverable by default? and non recoverable by option

PR #1108, opened 2026-05-23
feat(_task): allow secret and env_vars overrides on reusable tasks

EngHabu:
## Motivation

`task.override(secrets=...)` and `task.override(env_vars=...)` previously raised when the task's environment is reusable:

```
ValueError: Cannot override secrets when reusable is set. ...
ValueError: Cannot override env when reusable is set. ...
```

The guard's intent — keep one actor pool per env — is already enforced *downstream* by the actor pool key. `extract_unique_id_and_image` (in `_internal/runtime/reuse.py`) folds both the task's `security_context` **and** the serialized container (which carries the env vars) into the pool key. So a secret or env-var override deterministically maps to a **different** actor pool, never a polluted shared one.

## Concrete use case

We're building a multi-user "Claude Code in Flyte" service where each user owns a per-identifier credentials secret (`<identifier>-claude-session`) and per-session env. The launcher does:

```python
pinned = claude_session.override(
    secrets=flyte.Secret(key=secret_key, mount=SECRET_MOUNT),
    env_vars={"SESSION_ID": session_id},
)
return await pinned(...)
```

`claude_session` stays reusable so concurrent sessions for the *same* user share an actor pod (and its FUSE-mounted volume at `/root`); cross-user/session isolation comes from the secret + env override routing to a distinct pool. The old guard blocked this outright, and `reusable='off'` would cost us the pool sharing we explicitly want.

## What changed

Drop the `secrets` and `env_vars` checks from the `reusable` guard in `Task.override()`. Both feed the actor pool key, so overriding either routes to a distinct pool instead of mutating a shared one.

**Resources stay locked.** A reusable actor runs with its environment's pod spec, so a per-task resource override wouldn't take effect on the shared actor — `override(resources=...)` on a reusable task still raises; use `reusable='off'` to change resources.

## Test plan

- [x] `test_override_secrets_with_reuse` — secret override on a reusable env is allowed.
- [x] `test_override_env_vars_with_reuse` — env-var override on a reusable env is allowed.
- [x] `test_oomer_override_with_reuse_incorrect` — resources override on a reusable env still raises.

### DIFF
--- a/src/flyte/_internal/runtime/task_serde.py
+++ b/src/flyte/_internal/runtime/task_serde.py
@@ -282,8 +282,10 @@ def get_proto_task(
             raise flyte.errors.RuntimeUserError(
                 "BadConfig", f"Expected ReusePolicy, got {type(task.reusable)} for task {task.name}"
             )
-        env_name = None
-        if task.parent_env is not None:
+        # Prefer the override-derived name (set when env_vars/secrets are overridden
+        # on a reusable task) so each distinct combination gets its own actor pool.
+        env_name = task._actor_pool_env_name
+        if env_name is None and task.parent_env is not None:
             env = task.parent_env()
             if env is not None:
                 env_name = env.name
--- a/src/flyte/_task.py
+++ b/src/flyte/_task.py
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
 import asyncio
+import hashlib
+import json
 import weakref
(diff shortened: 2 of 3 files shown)


kumare3:
@EngHabu we should not do this, this will result in many actor pools and weird errors that are hard to debug. I am not in favor of this change. Its hard for users to know that this will impact reusable environments. I think its better to explicitly create new environments

PR #1110, opened 2026-05-25
examples(genai): add Concentrate AI gateway example

AaronSiddiky:
## Summary

Adds `examples/genai/concentrate_agent.py`, a near-identical mirror of the existing [`examples/genai/openai_agent.py`](https://github.com/flyteorg/flyte-sdk/blob/main/examples/genai/openai_agent.py) that routes every LLM call through the [Concentrate AI](https://concentrate.ai) gateway.

The diff vs. `openai_agent.py` is intentionally tight:

- Override `base_url` on `AsyncOpenAI` to `https://api.concentrate.ai/v1`
- Swap the Flyte `Secret` from `OPENAI_API_KEY` to `CONCENTRATE_API_KEY`
- Pin a Concentrate model slug (`"gpt-5.4"`; users can swap in `"auto"` for least-cost routing or `"openai/gpt-5.4"` to provider-pin)
- Disable the `openai-agents` default tracing exporter (it ships traces to OpenAI's tracing endpoint, which rejects Concentrate keys)

Tool calling, agent orchestration, and Flyte task wrapping are unchanged because Concentrate exposes the same `/v1/responses` surface that `openai-agents` uses by default. Governance, failover, spend visibility, and guardrails are configured per API key in the Concentrate dashboard, so no extra code is required in the example to demo them; once traffic flows through the gateway it picks up the policy.

## Context

Coordinated with @samhita-alla (see thread on the Union/Concentrate test accounts) — Samhita suggested contributing an example here as the starting point for the partnership and offered to amplify once merged. Happy to pair with a blog post on our end too.

## Test plan

- [x] `ruff format` + `ruff check` pass
- [x] Verified Concentrate `/v1/responses` works end-to-end with `gpt-5.4` via `AsyncOpenAI(base_url=...)` outside Flyte
- [x] Verified the full `openai-agents` Runner path (`set_default_openai_client` + `set_tracing_disabled` + `Agent(model="gpt-5.4", tools=[...])`) executes a tool-call loop end-to-end against Concentrate
- [ ] Reviewer can run the full Flyte workflow against a cluster with a `concentrate_api_key` secret

Made with [Cursor](https://cursor.com)

### DIFF
--- a/examples/genai/concentrate_agent.py
+++ b/examples/genai/concentrate_agent.py
@@ -1,0 +1,140 @@
+# /// script
+# requires-python = "==3.13"
+# dependencies = [
+#     "flyte>=2.0.0",
+#     "flyteplugins-openai>=2.0.0",
+#     "openai-agents>=0.11.1",
+# ]
+# ///
+
+
+import asyncio
+import os
+from typing import Optional
+
+from agents import Agent, Runner, set_default_openai_client, set_tracing_disabled
+from flyteplugins.openai.agents import function_tool
+from openai import AsyncOpenAI
+
+import flyte
+
+CONCENTRATE_BASE_URL = "https://api.concentrate.ai/v1"
+# Concentrate routes 120+ models. Use a specific slug like "gpt-5.4" for
+# deterministic behavior, "openai/gpt-5.4" to provider-pin, or "auto" to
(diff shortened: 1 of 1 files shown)


AaronSiddiky:
Closing — opened prematurely, will reopen after internal review.

PR #1111, opened 2026-05-25
fix(remote-builder): translate PythonWheels → CopyConfig + Commands

EngHabu:
Bottom of the volume-examples stack. Translates `PythonWheels` image layers into `CopyConfig` + `Commands` for the server-side image builder, since the `flyteidl2` `PythonWheels` proto lacked `package_name` at the time.

⚠️ Superseded higher in the stack: `haytham/volume-examples` reverts this to the native `python_wheels` proto now that `package_name` exists.

---
_Stack (bottom→top):_
1. `haytham/fix-remote-builder-pythonwheels` → `main`
2. `haytham/fix-reusable-secret-override`
3. `haytham/output-name-in-context`
4. `haytham/volume-examples`


EngHabu:
Closing: obsolete. `main` already emits the native `python_wheels` layer with `package_name` (landed in #1101), so this branch's CopyConfig+Commands translation is a regression rather than a fix. The volume-examples stack is being rebased off `main` to drop it.

PR #1112, opened 2026-05-25
feat(convert): expose current output name to type transformers

EngHabu:
## What

Adds a contextvar that exposes the **name of the output slot** currently being serialized (e.g. `o0`) to type transformers, plus a `current_output_name()` accessor.

`convert_from_native_to_outputs` sets the contextvar around each `TypeEngine.to_literal` call (and clears it in a `finally`), so a transformer can attribute a returned value to the specific output it's being serialized as — without threading the name through `to_literal`'s signature, which every transformer would otherwise have to adopt.

- `None` on the input path and outside output conversion; readers treat absence as "unknown".
- Scoped per-conversion via `set`/`reset`, so it never leaks across outputs.

## Why

The `flyteplugins-union` Volume lineage work needs to record which task output a `Volume` was returned as (so `flyte volume explore` can deep-link to a specific output and stamp `produced_by.output_name`). That information only exists at output-conversion time. This is the minimal, generic hook that surfaces it; the plugin reads it defensively and degrades to `None` against an SDK without this change.

## Scope

~24 lines in `convert.py`, behind the existing output loop. No behavior change for existing transformers.

### DIFF
--- a/src/flyte/_internal/runtime/convert.py
+++ b/src/flyte/_internal/runtime/convert.py
@@ -2,11 +2,12 @@ from __future__ import annotations
 
 import asyncio
 import base64
+import contextvars
 import hashlib
 import inspect
 from dataclasses import dataclass
 from types import NoneType
-from typing import Any, Dict, List, Tuple, Union, get_args
+from typing import Any, Dict, List, Optional, Tuple, Union, get_args
 
 from flyteidl2.core import execution_pb2, interface_pb2, literals_pb2
 from flyteidl2.task import common_pb2, task_definition_pb2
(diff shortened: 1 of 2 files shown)


src/flyte/_internal/runtime/convert.py
```diff
@@ -17,6 +18,23 @@
 from flyte.models import ActionID, NativeInterface, TaskContext
 from flyte.types import TypeEngine, TypeTransformerFailedError
 
+# Name of the output slot currently being serialized (e.g. "o0"), scoped to
+# a single ``TypeEngine.to_literal`` call by ``convert_from_native_to_outputs``.
+# Lets a TypeTransformer attribute a value to the output it's being returned
+# as — without threading the name through ``to_literal``'s signature (which
+# every transformer would have to adopt). ``None`` outside output conversion
+# (e.g. on the input path), so readers must treat absence as "unknown".
+_output_name_var: contextvars.ContextVar[Optional[str]] = contextvars.ContextVar(
```
kumare3, line 27: what is this? how do we use it? where are the tests?

EngHabu:
I wanted to capture what name of the output that is producing the Volume to add it to the lineage...
Added tests that show both how to use it (just call current_output_name() within the to_literal() call)... I can move it to an argument to the interface I guess? not sure if you have a preference...

kumare3 (approved):

PR #1113, opened 2026-05-25
examples(volumes): typed RO/RW lifecycle + enable_fuse_mount + reusable env overrides

EngHabu:
Top of the stack — the flyte-sdk support + examples for the `flyteplugins-union` Volume type.

## SDK

- **`enable_fuse_mount` flag** on `Environment` / `TaskEnvironment`: at serialize time, injects `CAP_SYS_ADMIN` + a `/dev/fuse` host device into the pod spec, so in-process FUSE mounts work without a hand-written PodTemplate (`_environment.py`, `_pod.py`, `_internal/runtime/task_serde.py`, `app/_runtime/app_serde.py`).
- **`env_vars` overrides on reusable tasks** (`_task.py`, `tests/user_api/test_override.py`).

## Examples (`examples/volumes/`)

- Typed `ROVolume`/`RWVolume` lifecycle across a multi-task lineage (`init → append → branch → read_all`), a cold-fork demo, and a metadata-store benchmark (sqlite / badger / redis).
- A plain image + the locally-built `flyteplugins-union` wheel (via `with_local_flyteplugins_union`) is all that's needed; the default `sqlite` store is forkable, so no `metadata_store_type=` pin.

Validated end-to-end on demo: full `init → append → branch → append → read_all` lineage SUCCEEDED.

---
_Stack (bottom → top):_
1. `haytham/fix-reusable-secret-override` → `main` (#1108)
2. `haytham/output-name-in-context` (#1112)
3. `haytham/volume-examples` (#1113, this)


examples/volumes/volume_cold_fork.py
```diff
@@ -0,0 +1,106 @@
+# /// script
+# requires-python = "==3.12"
+# dependencies = [
+#    "kubernetes",
+#    "flyte",
+#    "flyteplugins-union",
```
kumare3: we should probably put kubernetes as a dependency of this, its weird to add kuberenetes otherwise or we should raise an error when you import volume to add k8s

examples/volumes/volume_example.py
```diff
@@ -0,0 +1,147 @@
...
+
+@env.task(cache="auto")
+async def init_volume(volume_name: str) -> ROVolume:
+    logger.info("init_volume: declaring fresh volume name=%s", volume_name)
+    # Volume.new() returns a writable RWVolume backed by the default sqlite
+    # store — daemon-less and fork-capable, so the later `branch`/`append`
+    # forks Just Work with no extra image deps.
+    vol = Volume.new(name=volume_name)
+    logger.info("init_volume: bucket resolved to %s", vol.bucket)
+    await vol.mount()
```
kumare3: can we pass the path where we want to mount to this?

examples/volumes/volume_example.py
```diff
@@ -0,0 +1,147 @@
...
+    Path("/workspace/hello.txt").write_text("hello from volume\n")
+    # finalize() drains writeback, unmounts, and publishes an immutable
+    # ROVolume. (Returning the mounted RWVolume directly would do the same
+    # via the type transformer.)
+    sealed = await vol.finalize(message="initial write")
```
kumare3, line 86: how about volume.commit? or persist?

examples/volumes/volume_cold_fork.py
EngHabu: Removed... This was needed when I had the pod template in here... it's already in flyteplugins-union

examples/volumes/volume_example.py
EngHabu: it takes `mount_path`... added to the example

kumare3:
The core of this shipped differently in #1205 — unprivileged `allow_fuse()` via the FUSE device plugin rather than the privileged hostPath path here. This is also stacked on closed #1108. Closing; reopen a fresh PR if any of the example/lifecycle work is still wanted.

PR #1116, opened 2026-05-26
add support for Union types with Field annotation in pydantic types

cosmicBboy:
This pull request adds robust support for Pydantic v2 discriminated (tagged) unions—schemas using `oneOf` and a top-level `discriminator`—to Flyte's type engine. It fixes previous issues where such schemas would cause errors, and ensures that union types are correctly roundtripped, validated, and reconstructed. The changes include enhancements to schema parsing, dynamic class generation, and comprehensive regression tests.

**Support for Pydantic v2 discriminated unions:**

* Updated the type engine to recognize and properly handle JSON schemas with `oneOf` and a top-level `discriminator` (as emitted by Pydantic v2 for `Annotated[Union[...], Field(discriminator=...)]`), ensuring correct type inference and runtime dispatch.
* Added the `_DiscriminatedUnion` descriptor to capture discriminator metadata and variant mappings, enabling correct dict-to-object conversion during dataclass instantiation.
* Improved handling of optional and list fields containing discriminated unions, so that `Optional[...]` and `List[...]` of such types are supported seamlessly.

**Testing and documentation:**

* Added a comprehensive regression test suite covering schema generation, type inference, correct dispatch on discriminator values, roundtripping through the type engine, and validation using dynamically generated Pydantic models.
* Introduced a new example (`examples/basics/types/pydantic_union_types.py`) demonstrating end-to-end usage of Pydantic v2 discriminated unions as Flyte task inputs.

These changes ensure that Flyte tasks can now reliably accept and process Pydantic v2 discriminated unions, addressing previous limitations and improving compatibility with modern Pydantic codebases.

### DIFF
--- a/src/flyte/types/_type_engine.py
+++ b/src/flyte/types/_type_engine.py
@@ -1194,13 +1194,46 @@ class _DiscriminatedUnion:
     """
 
     discriminator_property: typing.Optional[str]
-    mapping: typing.Mapping[str, type]
+    mapping: typing.Mapping[typing.Any, type]
     variants: typing.Tuple[type, ...]
 
 
+def _normalize_discriminator_value(value: typing.Any) -> typing.Any:
+    """Normalize a discriminator value for mapping lookup.
+
+    Pydantic v2 emits the schema-level ``discriminator.mapping`` keys as JSON
+    primitives (strings/ints/bools), but at runtime the corresponding model
+    field value can be an ``Enum`` member (e.g. when the discriminator field
+    is typed as a non-``str`` ``Enum``). Unwrap such values to their underlying
+    primitive so the lookup keys match.
+    """
+    if isinstance(value, enum.Enum):
+        return value.value
+    return value
+
+
+def _select_unambiguous_variant(variants: typing.Sequence[type], value: dict[str, Any]) -> type | None:
+    """Return the single variant whose dataclass fields accept ``value`` keys.
+
+    Used as a safe fallback when a ``oneOf`` schema lacks a usable discriminator.
+    Returns ``None`` if zero or more than one variant matches so the caller can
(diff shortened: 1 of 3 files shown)


src/flyte/types/_type_engine.py
```diff
@@ -2349,11 +2482,32 @@ def convert_mashumaro_json_schema_to_python_class(schema: dict, schema_name: typ
...
+                if isinstance(descriptor, _DiscriminatedUnion):
+                    target_cls = None
+                    disc_property = descriptor.discriminator_property
+                    if disc_property is not None:
+                        disc_value = value.get(disc_property)
+                        if disc_value is not None:
+                            target_cls = descriptor.mapping.get(disc_value)
+                    if target_cls is None:
+                        # Best effort: try each variant constructor until one succeeds. This is the
+                        # fallback path when no discriminator is specified or it does not match.
```
AdilFayyaz: If the mapping lookup fails and we fall through here, what stops us from picking the wrong variant? e.g. if two models have the same fields, the first one tried would always win silently.

src/flyte/types/_type_engine.py
```diff
@@ -1210,18 +1235,64 @@ def _append_schema_field(
...
+
+    for variant in variants:
+        if not isinstance(variant, dict):
+            variant_types.append(_get_element_type(variant, schema))
+            continue
+        if variant.get("$ref"):
+            ref_name = variant["$ref"].split("/")[-1]
+            if ref_name in defs:
+                ref_schema = defs[ref_name].copy()
+                if ref_schema.get("enum"):
```
AdilFayyaz, line 1304: what if the discriminator field is an enum? Would the mapping still work?

AdilFayyaz (approved):

PR #1117, opened 2026-05-27
Update edit settings to remove unused string list setting 

afbrock:
This change also improves string map handling.
The label setting changed types from string list to map, and was the only instance of a string list.  This change removes the unused type.  It also improves how maps display entries from different scopes.  Each entry is displayed on its own line with a comment as to its source scope.

### DIFF
--- a/src/flyte/cli/_edit.py
+++ b/src/flyte/cli/_edit.py
@@ -118,11 +118,7 @@ def settings(cfg: common.CLIConfig, project: str | None, domain: str | None, fro
 
     console = common.Console()
 
-    try:
-        settings = remote.Settings.get_settings_for_edit(project=project, domain=domain)
-    except Exception as e:
-        console.print(f"[red]Error fetching settings:[/red] {e}")
-        raise click.Abort
+    settings = remote.Settings.get_settings_for_edit(project=project, domain=domain)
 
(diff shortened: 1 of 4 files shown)


src/flyte/cli/_edit.py
```diff
@@ -118,11 +118,7 @@ def settings(cfg: common.CLIConfig, project: str | None, domain: str | None, fro
 
     console = common.Console()
 
-    try:
-        settings = remote.Settings.get_settings_for_edit(project=project, domain=domain)
-    except Exception as e:
-        console.print(f"[red]Error fetching settings:[/red] {e}")
-        raise click.Abort
+    settings = remote.Settings.get_settings_for_edit(project=project, domain=domain)
```
kumare3, line 121: why remove the Abort?

afbrock: When dogfood was misconfigured i actually got an exception while trying to test, and this re-throwing of an error hides the exception.  The exception gets displayed with less noise if there isnt a re-throw. 
what it says is that during the handling of the exception, another error was thrown - so you get two stacks, and it isnt clear that they have anything to do with each other.

kumare3:
you will have to signoff on dco and also make fmt

kumare3 (approved):

PR #1128, opened 2026-05-29
Fixes recursive traces

kumare3:


SVilgelm (dismissed):

src/flyte/_internal/controllers/remote/_controller.py
```diff
@@ -401,7 +401,7 @@ async def get_action_outputs(
                     org=current_action_id.org,
                 ),
             ),
-            current_action_id.name,
+            (tctx.parent_action or current_action_id).name,
```
wild-endeavor: why or?

wild-endeavor: is there a legitmate use case for a trace where the parent context is not set?

src/flyte/models.py
```diff
@@ -219,6 +219,10 @@ class TaskContext:
     :param run_start_time: UTC datetime at which the parent run was triggered. Populated by the backend via the
       ``{{.runStartTime}}`` template; defaults to ``datetime.now(timezone.utc)`` when not supplied so local runs
       always have a value.
+    :param parent_action: The action ID of the running container. Set once at task-execution start and preserved
```
wild-endeavor: can we add to this comment please to state that this is not the parent of the running action.  this is the running action itself.  Maybe

```
The action ID of the currently running action. Set at task-execution start even for all Task actions including the root action and preserved across `@trace` action replacements (which swap `action` per trace scope). Used as `parent_action_name` when submitting trace records, so trace bookkeeping nests under the real running action — not enclosing traces in the case of nested trace actions.
```

src/flyte/models.py
```diff
@@ -241,6 +245,7 @@ class TaskContext:
     #: Type transformers should omit non-essential side effects (e.g. duplicate HTML tabs) when set.
     in_driver_literal_conversion: bool = False
     run_start_time: Optional[datetime] = field(default_factory=lambda: datetime.now(timezone.utc))
+    parent_action: ActionID | None = None
```
wild-endeavor: Why None?  it's set to the current action right?

src/flyte/models.py
kumare3: ya i was going to call it original action?

src/flyte/models.py
kumare3: I guess i have set it so everywhere, sure, will set it, i think i will rename it to original_action?

src/flyte/_internal/controllers/remote/_controller.py
kumare3: nope

pingsutw (approved):

wild-endeavor (approved):

PR #1134, opened 2026-05-31
fix: wrap KeyError at user-module import as ModuleLoadError

EngHabu:
## Summary

When a user module does \`os.environ[\"GCP_ADR\"]\` at top level and the env var is missing, the resulting \`KeyError\` escapes the module loader and bubbles into Sentry as if it were an SDK crash (FLYTE-SDK-3D — same user produces sibling \`TypeError\` (FLYTE-SDK-3B) and \`NameError\` (FLYTE-SDK-3C) which are already wrapped).

The module loader's two \`except\` clauses already catch \`(ImportError, SyntaxError, NameError, AttributeError, TypeError, ValueError)\` at user-code import time and wrap them as \`ModuleLoadError\`. This change adds \`KeyError\` to both clauses so env-var-missing failures at the top of user modules surface as \`ModuleLoadError\` — a \`RuntimeUserError\` subclass already filtered by \`flyte/_sentry.py:_is_user_error\`.

The list is intentionally not broadened further. \`KeyError\` is consistent with \`NameError\` (also a common top-level user-config failure) and tight enough to not mask runtime SDK bugs.

## Closes

- [FLYTE-SDK-3D](https://unionai.sentry.io/issues/FLYTE-SDK-3D/) — \`KeyError: 'GCP_ADR'\` from \`os in __getitem__\` at the top of a user task module.

\`fixes FLYTE-SDK-3D\`

## Test plan

- [x] Parametrized \`test_load_python_modules_single_file_wraps_user_code_errors\` now includes \`KeyError\` and asserts \`ModuleLoadError\` is raised.
- [x] New \`test_load_python_modules_single_file_wraps_missing_env_var_keyerror\` reproduces the FLYTE-SDK-3D shape (\`os.environ[\"NEVER_SET_GCP_ADR_XYZ\"]\` at module top level) and asserts \`ModuleLoadError\` wraps it with the file path and the missing key in the message.
- [x] All 12 tests in \`tests/flyte/utils/test_module_loader.py\` pass.
- [x] \`make fmt\` clean.

### DIFF
--- a/src/flyte/_utils/module_loader.py
+++ b/src/flyte/_utils/module_loader.py
@@ -58,7 +58,7 @@ def load_python_modules(
             loaded_modules.append(imported_module)
         except flyte.errors.ModuleLoadError as e:
             failed_paths.append((path, str(e)))
-        except (ImportError, SyntaxError, NameError, AttributeError, TypeError, ValueError) as e:
+        except (ImportError, SyntaxError, NameError, AttributeError, TypeError, ValueError, KeyError) as e:
             raise flyte.errors.ModuleLoadError(f"Failed to load {path}: {type(e).__name__}: {e}") from e
 
     elif path.is_dir():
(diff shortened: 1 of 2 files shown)


EngHabu (commented):
LGTM

AdilFayyaz (approved):

PR #1138, opened 2026-06-01
feat: add ClusteredTaskEnvironment for distributed multi-node training

AdilFayyaz:
## Motivation

  Flyte had no first-class way to run distributed multi-node training jobs. This adds a ClusteredTaskEnvironment
  that emits a Kubernetes JobSet and launches workers via torchrun, exposing the distributed runtime context (rank,
  world size, master address, etc.) to user tasks.
  
  ## Summary

  - Add flyte.distributed public API: ClusteredTaskEnvironment, TorchRun, and ClusterFailurePolicy, with validation
  (replica/proc counts, GPU-vs-nproc_per_node, interconnect) and to_custom_dict() that serializes to the
  ClusteredTaskSpec proto.
  - task_serde detects a clustered parent env and rewrites the proto task: sets type clustered-task (v1), swaps the
  container command to the flyte.distributed._entrypoint wrapper, and attaches the clustered custom config — for
  both plain-container and pod-template paths.
  - _entrypoint exports derived NODE_RANK/MASTER_ADDR so torchrun children inherit them, and adds --no-python so
  torchrun execs the a0 console script directly.
  - Expose distributed fields on TaskContext (rank, local_rank, world_size, node_rank, master_addr, master_port,
  restart_attempt, …), all None on non-clustered tasks.
  - In io.py, only rank-0 uploads outputs/errors so other ranks don't race to clobber error.pb.
  - Bump flyteidl2 2.0.17 → 2.0.18 (root + rs_controller) for the new clustered_pb2 types.

  Test Plan

  - uv run pytest tests/flyte/distributed/test_env.py tests/flyte/distributed/test_serde.py
  - Validation: confirm ClusteredTaskEnvironment raises on replicas < 1, nproc_per_node < 1, gpu < nproc_per_node,
  unsupported runtime, and invalid interconnect; non-tcp interconnect warns.
  - Serde: confirm a task under a clustered env produces type clustered-task, command python -m 
  flyte.distributed._entrypoint, and a populated custom block (both container and pod-template paths).
  - End-to-end: run examples/distributed/ddp_train.py and verify the JobSet launches, ranks join rendezvous, and
  only rank-0 writes outputs.


src/flyte/distributed/_environment.py
```diff
@@ -0,0 +1,138 @@
+from __future__ import annotations
+
+import warnings
```
kumare3: lets make this `flyte.clustered` instead of distributed. flyte is already distributed

src/flyte/distributed/_environment.py
```diff
@@ -0,0 +1,138 @@
...
+    :param runtime: Launcher configuration. Phase 1 supports only TorchRun().
+    :param interconnect: Network fabric. Non-TCP options require RDMA-capable images (Phase 3).
+    :param failure_policy: JobSet-level restart and eviction policy.
+    :param ttl_seconds_after_finished: Seconds to retain the JobSet after completion.
+    """
+
+    replicas: int
+    nproc_per_node: int
+    runtime: Runtime = field(default_factory=TorchRun)
+    interconnect: Literal["tcp", "efa", "infiniband", "roce"] = "tcp"
```
kumare3: remove the interconnects that dont work, you can keep one `tcp`

src/flyte/distributed/_entrypoint.py
kumare3, line 131: lets check if torchrun exists, if not lets raise a nice error (it will be printed to stderr and flyte will catch it). The error should be `please install torchrun`

kumare3:
can you test https://github.com/axolotl-ai-cloud/axolotl 
or
https://github.com/huggingface/accelerate

src/flyte/clustered/_environment.py
```diff
@@ -0,0 +1,138 @@
...
+
+    max_restarts: int = 0
+    restart_on_host_maintenance: bool = False
+
+
+_INTERCONNECT_VALUES = ("tcp", "efa", "infiniband", "roce")
+
+
+@dataclass(kw_only=True)
+class ClusteredTaskEnvironment(TaskEnvironment):
```
cosmicBboy, line 48: What do you think about `DistributedTaskEnvironment`?

Seems to reflect what this task env is all about (and in fact, this is what the subpackage `flyte.distributed` is already called 🙂)

AdilFayyaz: Yeahh but was talking to Ketan about this and we plan on changing this to `flyte.clustered` instead since Flyte by default is already distributed, and this can be a Clustered Task which is what we are essentially doing.

src/flyte/_internal/runtime/task_serde.py
```diff
@@ -212,6 +236,17 @@ def get_proto_task(
 
     custom = task.custom_config(serialize_context)
 
+    # --- clustered task: rewrite type + custom (command was rewritten above) ---
+    _task_type = task.task_type
```
pingsutw: Should we override the `task` func in the `ClusteredTaskEnvironment`, so that the task will return the correct type, custom, command here and we don't need to change anything task_serde.py. 
https://github.com/flyteorg/flyte-sdk/blob/9b87f13384a842ed548c047718a1f93109e8aa51/src/flyte/_task_environment.py#L262-L273

pingsutw (dismissed):

src/flyte/_internal/runtime/task_serde.py
AdilFayyaz: That makes sense, ill fix this

src/flyte/_internal/runtime/io.py
```diff
@@ -59,6 +61,9 @@ async def upload_outputs(outputs: Outputs, output_path: str, max_bytes: int = -1
     :param output_path: The path to upload the output file.
     :param max_bytes: Maximum number of bytes to write to the output file. Default is -1, which means no limit.
     """
+    # In clustered tasks, only rank-0 owns the output; all other ranks skip upload.
+    if os.environ.get("RANK", "0") != "0":
```
pingsutw: I think it's safer to have another gate here since user may have `RANK` env var in their regular python task, like
```python
if is_clusterd or os.environ.get("RANK", "0") != "0":
```
or we rewrite [load_and_run_task](https://github.com/flyteorg/flyte-sdk/blob/0cec0d833b0ff25ee7c760d3cb2829761793608e/src/flyte/_internal/runtime/entrypoints.py#L210) for jobset - load_and_run_jobset, so that we can decide what to download/upload in it

src/flyte/_internal/runtime/io.py
```diff
@@ -77,6 +82,11 @@ async def upload_error(err: execution_pb2.ExecutionError, output_prefix: str, re
     :param output_prefix: The output prefix of the remote uri.
     :param recoverable: If False, sets ContainerError.kind to NON_RECOVERABLE so the engine skips retries.
     """
+    error_uri = error_path(output_prefix)
+    # In clustered tasks, only rank-0 owns the error file; other ranks skip the write
+    # so they don't race to clobber error.pb. (RANK is unset on non-clustered tasks.)
+    if os.environ.get("RANK", "0") != "0":
```
pingsutw: similar

src/flyte/_bin/runtime.py
```diff
@@ -27,41 +27,81 @@
...
+def main(ctx: click.Context, **params):
+    _run_action(ctx, controller_enabled=True, **params)
+
+
+@_pass_through.command("clustered")
+@_action_options
+@click.pass_context
+def clustered_main(ctx: click.Context, **params):
+    """Entrypoint for clustered (JobSet-based) distributed training pods.
+
```
pingsutw: nit, could we move it to another file? _bin/clusterd.py

src/flyte/_internal/runtime/io.py
AdilFayyaz: oh yes nice catch

src/flyte/_bin/runtime.py
AdilFayyaz: yeah

src/flyte/_bin/clustered.py
```diff
@@ -0,0 +1,137 @@
...
+    "NNODES",
+    "NPROC_PER_NODE",
+    "RDZV_BACKEND",
+)
+
+
+@click.command("clustered")
+@_action_options
+@click.pass_context
+def clustered_main(ctx: click.Context, **params):
```
pingsutw: nit
```suggestion
def main(ctx: click.Context, **params):
```

pyproject.toml
```diff
@@ -33,6 +33,7 @@ dependencies = [
 [project.scripts]
 flyte = "flyte.cli.main:main"
 a0 = "flyte._bin.runtime:main"
+clustered = "flyte._bin.clustered:clustered_main"
```
pingsutw: nit: 
```suggestion
clustered = "flyte._bin.clustered:main"
```

pingsutw (approved):

PR #1139, opened 2026-06-01
Add 'flyte delete local-cache' command to clear the local cache directory

pingsutw:
## Why

The Flyte SDK keeps a local cache directory at `~/.flyte/local-cache` (a SQLite
DB used for image lookups, bundle uploads, run history, and task caching). When
stale entries cause incorrect behavior, there was no first-class way to clear it
short of manually deleting the directory.

## What

Adds a `flyte delete local-cache` CLI command that removes the **entire**
`~/.flyte/local-cache` directory.

```console
$ flyte delete local-cache
Cleared local cache directory: /Users/me/.flyte/local-cache
```

- New `LocalDB._get_cache_dir()` — resolves the local-cache directory (parent of
  the cache DB), and `_get_db_path()` now derives from it.
- New `LocalDB.purge()` — closes all connections, resets state, and `rmtree`s the
  local-cache directory. Returns the targeted path; re-initialization recreates
  the DB cleanly on next use.
- New `local-cache` subcommand under the existing `flyte delete` group.

## Tests

- `tests/persistence/test_purge.py` — `LocalDB.purge()` removes the whole dir,
  is a no-op when missing, and the DB is recreated on re-init.
- `tests/cli/test_delete.py` — CLI command invokes `purge` and reports the path.

All tests pass; `ruff format`/`ruff check` clean.


src/flyte/_internal/imagebuild/image_builder.py
```diff
@@ -151,6 +151,18 @@ def _write_image_cache(repository: str, tag: str, arch: Tuple[str, ...], image_u
         logger.debug(f"Failed to write image cache: {e}")
 
 
+def clear_image_cache() -> int:
```
kumare3: why here? Shall we do it in persistence?

pingsutw: updated it

src/flyte/_persistence/_db.py
```diff
@@ -220,3 +226,20 @@ def close_sync():
...
+        """Close all connections and delete the entire local-cache directory.
+
+        Returns the directory that was targeted for removal (whether or not it existed).
+        """
+        cache_dir = LocalDB._get_cache_dir()
+        with LocalDB._lock:
+            if LocalDB._conn_sync:
+                LocalDB._conn_sync.close()
+                LocalDB._conn_sync = None
+            LocalDB._conn = None
```
AdilFayyaz, line 241: Should we be closing `_conn` similar to `_conn_sync` L239

pingsutw: `_conn` is the async `aiosqlite` connection, which can only be closed via `await _conn.close()`. `purge()` is synchronous (it backs the sync `flyte delete local-cache` Click command), so we can't await here — same reason `close_sync()` doesn't close `_conn` either.

In this code path `_conn` is never even opened: the async connection is only created by `initialize()`/`_initialize_async()`, which the sync CLI never calls, so `_conn` is always `None` here and we just drop the reference. The `shutil.rmtree` removes the on-disk DB regardless.

(We could wrap it in `asyncio.run(_conn.close())`, but that would only ever be dead code on this path, and it would make `purge()` raise `RuntimeError` if it's ever called from within a running event loop — so I'd rather keep it sync-safe.)

AdilFayyaz (approved):

PR #1140, opened 2026-06-01
Add labels to create run and list runs

afbrock:
This commit adds --label to create run command, can be used multiple times and takes a key=value pair.
example:  flyte run --label team=ml --label env=prod my_workflow.py main

Also adds label matching to list runs:
example:  flyte get run --with-label team=ml
example:  flyte get run --with-label-key team


kumare3:
looks good, unit tests

afbrock:
> looks good, unit tests

unit tests will fail until https://github.com/flyteorg/flyte/pull/7430 merges

PR #1142, opened 2026-06-02
fix: handle None record name in flyte log record factory

AdilFayyaz:
## Motivation
  
  _flyte_record_factory crashed with AttributeError when record.name was None, which happens when logging.makeLogRecord (used by structlog's stdlib bridge) invokes the global record factory without a name.
  
 ## Summary

  - Coalesce record.name to an empty string before computing is_flyte_internal, preventing the NoneType.startswith crash.
  - A None-named record now safely resolves to is_flyte_internal = False.
  - Added test_record_factory_with_none_name to reproduce the original crash and guard against
  regressions.

  ## Test Plan
  
  - Run the logging tests: uv run pytest tests/flyte/test_logging.py -v
  - Verify both test_record_factory_with_none_name and test_logging pass (2 passed).
  - The new test directly invokes the factory with name=None and asserts it does not raise and returns is_flyte_internal is False.

### DIFF
--- a/src/flyte/_logging.py
+++ b/src/flyte/_logging.py
@@ -33,9 +33,7 @@ def _flyte_record_factory(*args: Any, **kwargs: Any) -> logging.LogRecord:
         record.action_name = None
 
     name = record.name or ""
-    record.is_flyte_internal = name == "flyte" or (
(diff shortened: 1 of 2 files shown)


pingsutw (approved):
LGTM, thanks!

PR #1150, opened 2026-06-03
[Devbox] Skip k8s context switch and kubeconfig update when kubectl not installed

popojk:
## Goal
Skip k8s context switch and kubeconfig update when kubectl not installed instead of blocking the devbox start up process

## tests
Unit tests

### DIFF
--- a/src/flyte/cli/_devbox.py
+++ b/src/flyte/cli/_devbox.py
@@ -36,12 +36,9 @@ def _ensure_docker_available() -> None:
         )
 
 
-def _ensure_kubectl_available() -> None:
-    if shutil.which("kubectl") is None:
-        raise click.ClickException(
-            "kubectl is not installed or not on PATH. Install kubectl "
-            "(https://kubernetes.io/docs/tasks/tools/) and try again."
-        )
+def _ensure_kubectl_available() -> bool:
+    """Return True if kubectl is installed and on PATH, False otherwise."""
+    return shutil.which("kubectl") is not None
 
 
 def _run_docker(cmd: list[str], failure_message: str) -> subprocess.CompletedProcess:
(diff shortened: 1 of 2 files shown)


src/flyte/cli/_devbox.py
```diff
@@ -36,12 +36,9 @@ def _ensure_docker_available() -> None:
...
 
-def _ensure_kubectl_available() -> None:
-    if shutil.which("kubectl") is None:
-        raise click.ClickException(
-            "kubectl is not installed or not on PATH. Install kubectl "
-            "(https://kubernetes.io/docs/tasks/tools/) and try again."
-        )
+def _ensure_kubectl_available() -> bool:
+    """Return True if kubectl is installed and on PATH, False otherwise."""
+    return shutil.which("kubectl") is not None
```
popojk, line 41: We don't raise error here, so that the the k8s context switch and kubeconfig update can be optional.

src/flyte/cli/_devbox.py
```diff
@@ -201,7 +204,13 @@ def _flatten_kubeconfig(default_kubeconfig: Path, kubeconfig_path: Path) -> subp
 def _merge_kubeconfig(kubeconfig_path: Path, container_name: str) -> None:
     import tempfile
 
-    _ensure_kubectl_available()
+    if not _ensure_kubectl_available():
+        click.echo(
```
pingsutw: The output looks strange, can we try console.print() or something else

Image

src/flyte/cli/_devbox.py
```diff
@@ -36,12 +36,9 @@ def _ensure_docker_available() -> None:
         )
 
 
-def _ensure_kubectl_available() -> None:
-    if shutil.which("kubectl") is None:
-        raise click.ClickException(
-            "kubectl is not installed or not on PATH. Install kubectl "
-            "(https://kubernetes.io/docs/tasks/tools/) and try again."
-        )
+def _ensure_kubectl_available() -> bool:
```
pingsutw: probably rename it to _is_kubectl_installed

src/flyte/cli/_devbox.py
popojk: I just changed the error message into red warning letters. Does that looks good to you?
Screenshot 2026-06-04 at 8 33 03 AM

src/flyte/cli/_devbox.py
popojk: Fixed!

pingsutw (approved):

PR #1151, opened 2026-06-03
fix(extras/shell): optional flags, binary outputs, and debug work on remote

k1sauce:
Building the flyteplugins-bio test suite against a real devbox surfaced three shell-task bugs: 

- optional File/Dir flags were emitted for unset inputs (the backend writes a "null" sentinel file, defeating the `[ -e ]` check); 
- File/Dir outputs were read as UTF-8, corrupting binary outputs like gzip; 
- and debug=True output never reached remote logs. Fixes all three, and adds the bio wrappers and single-run e2e tests (cat/gunzip/untar + test_all) that exercised them.

TODO:
Glob / Dir outputs don't round-trip on remote yet. A wrapper with a multi-file output (currently untar) declares a multipart Dir correctly — the serialized task interface is blob { dimensionality: MULTIPART } — but the backend's copilot uploads the directory as a single blob, so the consumer's Dir transformer rejects it (Expected multipart, received 0). Single-File output tools work fine on remote; untar only runs locally until the backend supports multipart raw-container outputs. test_all includes it, so the full suite is local-only for now.

### DIFF
--- a/plugins/bio/README.md
+++ b/plugins/bio/README.md
@@ -14,13 +14,13 @@ each shipped via its official image. Built on `flyte.extras.shell`.
 Each submodule exposes:
 
 - One or more callable shell tasks (`bedtools_intersect`, `samtools_view`, ...)
-- A module-level `env` (a `flyte.TaskEnvironment`) that pipelines add to their
-  `depends_on`. Adding the env causes the deploy pipeline to register the task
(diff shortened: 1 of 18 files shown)


samhita-alla:
closing this PR as we now have https://github.com/flyteorg/flyte-sdk/pull/1158

PR #1163, opened 2026-06-04
set ImageIdentifier.repository for custom registries

dansola:
## Set `ImageIdentifier.repository` for custom registries

When checking whether an image already exists, tell the backend which repo to look in.

**Why:** for images built with `clone(registry=...)`, the existence check (`GetImage`) was
implicitly looking in the cluster's default repo, never finding the image, so it rebuilt every
run. Now we pass the custom registry/repo explicitly via the new `ImageIdentifier.repository`
field so the lookup hits. (SE-775)

Default-repo images send an empty `repository`, so nothing changes for them. Pulled the
detection into a small `_custom_repository()` helper with a unit test.

**Depends on** the `flyteidl2` release that includes `ImageIdentifier.repository` — needs a
`flyteidl2` version bump here once it's out.

### DIFF
--- a/src/flyte/_internal/imagebuild/remote_builder.py
+++ b/src/flyte/_internal/imagebuild/remote_builder.py
@@ -57,6 +57,15 @@ IMAGE_TASK_PROJECT = "system"
 IMAGE_TASK_DOMAIN = os.environ.get("FLYTE_IMAGEBUILDER_TASK_DOMAIN", "production")
 
 
+def _custom_repository(repository: str) -> str:
+    # SE-775: for a custom registry override (clone(registry=...)) the image is pushed to a custom
+    # repo; tell GetImage to look there instead of the cluster's default repo. Empty for the default.
+    first_segment = repository.split("/", maxsplit=1)[0]
+    if "." in first_segment and not repository.startswith(f"{_BASE_REGISTRY}/"):
+        return repository
+    return ""
+
+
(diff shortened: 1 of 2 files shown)


dansola:
https://github.com/unionai/cloud/pull/16338
https://github.com/flyteorg/flyte/pull/7474

src/flyte/_internal/imagebuild/remote_builder.py
```diff
@@ -57,6 +57,15 @@
 IMAGE_TASK_DOMAIN = os.environ.get("FLYTE_IMAGEBUILDER_TASK_DOMAIN", "production")
 
 
+def _custom_repository(repository: str) -> str:
```
pingsutw, line 60: nit: could we rename it to _get_registry

src/flyte/_internal/imagebuild/remote_builder.py
```diff
@@ -57,6 +57,15 @@
...
+    # SE-775: for a custom registry override (clone(registry=...)) the image is pushed to a custom
+    # repo; tell GetImage to look there instead of the cluster's default repo. Empty for the default.
+    first_segment = repository.split("/", maxsplit=1)[0]
```
pingsutw, line 63:
```suggestion
    registry = repository.split("/", maxsplit=1)[0]
```

PR #1167, opened 2026-06-05
fix: wrap docker buildx ls/create failures as ImageBuildError

EngHabu:
## Problem

`DockerImageBuilder._ensure_buildx_builder` runs `docker buildx ls` and `docker buildx create --name flytex ...` with `check=True`. When either command fails, the resulting `subprocess.CalledProcessError` bubbled out raw and was reported to Sentry as an SDK crash — **FLYTE-SDK-4R** (`CalledProcessError: Command '['docker', 'buildx', 'create', ...]' returned non-zero exit status 1`).

These failures are local Docker-environment problems (daemon not running, BuildKit driver unavailable, or a leftover/locked `flytex` builder), not SDK bugs.

## Fix

- Wrap the `docker buildx ls` and `docker buildx create` calls in `ImageBuildError` (a `RuntimeUserError`, which `_sentry.py` already filters out) with an actionable message that includes the captured `stderr` and points the user at `docker buildx rm flytex` or the remote image builder (`image_builder='remote'`).
- If `create` fails because the builder *already exists* (e.g. a concurrent build created it between our `ls` check and the `create`), reuse it instead of failing.

This mirrors the existing wrapping already done for `docker buildx version` and the dockerfile build step.

## Tests

Added three tests covering: create failure → `ImageBuildError`, `ls` failure → `ImageBuildError`, and `already exists` → reuse (no raise). All buildx tests pass.

fixes FLYTE-SDK-4R

---
**Update:** also covers **FLYTE-SDK-5J** (`docker buildx ls returned non-zero exit status 125`), which is the same `docker buildx ls` `CalledProcessError` path this PR now wraps in `ImageBuildError`.

### DIFF
--- a/src/flyte/_internal/imagebuild/docker_builder.py
+++ b/src/flyte/_internal/imagebuild/docker_builder.py
@@ -784,7 +784,6 @@ class DockerImageBuilder(ImageBuilder):
         except subprocess.CalledProcessError:
             raise ImageBuildError("Docker buildx is not available. Make sure BuildKit is installed and enabled.")
 
-        # List builders
         try:
             result = await run_sync_with_loop(
                 subprocess.run, ["docker", "buildx", "ls"], capture_output=True, text=True, check=True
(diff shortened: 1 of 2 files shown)


EngHabu (commented):
LGTM

pingsutw (approved):

PR #1094, opened 2026-05-21
fix: surface user-module load errors as ModuleLoadError + filter via Sentry user-error set

EngHabu:
## Summary

`flyte deploy <dir>` and `flyte deploy <file>` both invoke
`flyte._utils.module_loader.load_python_modules`. When a user `.py` file
fails at import time with `ModuleNotFoundError` (missing third-party
dep), `SyntaxError`, `NameError`, `KeyError`-via-bad-import, `TypeError`,
etc., the bare exception bubbled past the Sentry `capture_errors`
decorator in `flyte/cli/_deploy.py` and landed in Sentry as an SDK
crash — even though the code that failed was in the user's workflow,
not in flyte.

## Changes

1. **`flyte._utils.module_loader.load_python_modules`** — catch
   `ImportError` / `SyntaxError` / `NameError` / `AttributeError` /
   `TypeError` / `ValueError` raised during `importlib.import_module`:
   - **Single-file path**: wrap in `flyte.errors.ModuleLoadError`
     (a `RuntimeUserError`), preserving the original traceback via
     `from e`, so the CLI surfaces an actionable message and the
     exception type itself signals "user error".
   - **Directory path**: record the failure in `failed_paths` so deploy
     either warns or aborts via `--ignore-load-errors`, matching how
     pre-existing `ModuleLoadError`s were already handled.

2. **`flyte._sentry._is_user_error`** — classify any `RuntimeUserError`
   subclass as a user error. Previously hard-coded to
   `(DeploymentError, ImageBuildError, InitializationError)`, missing
   the other ~20 `RuntimeUserError` subclasses (`ModuleLoadError`,
   `OOMError`, `TaskTimeoutError`, `RuntimeDataValidationError`,
   `CodeBundleError`, `RemoteTaskNotFoundError`, etc.). Filtering on
   the parent class is the right primitive — these errors are
   intentionally raised as user-facing messages, not crash reports.

## Closes

All match the same `module_loader` → `<frozen importlib._bootstrap>` →
user-module shape; this PR filters them out at the Sentry boundary:

- fixes https://github.com/flyteorg/flyte-sdk/issues/FLYTE-SDK-3T (`fold_to_ascii`)
- fixes https://github.com/flyteorg/flyte-sdk/issues/FLYTE-SDK-3R (`langchain_community`)
- fixes https://github.com/flyteorg/flyte-sdk/issues/FLYTE-SDK-3Q (`requests`)
[...]

PR #1168, opened 2026-06-05
fix: wrap any user-module import exception as ModuleLoadError

EngHabu:
## Problem

User modules can raise arbitrary third-party exceptions at **import time** that don't inherit from `RuntimeError` / `ImportError` / etc. The clearest example: `google.api_core.exceptions.RetryError` (parent: `Exception`) is raised when a user's module-top-level code calls Google Secret Manager and gcloud credentials have expired. The SDK's import loop only catches a narrow tuple — `(ImportError, SyntaxError, NameError, AttributeError, TypeError, ValueError, KeyError, RuntimeError)` — so anything else escapes and lands in Sentry as if it were an SDK bug.

PRs #1094 / #1134 / #1146 widened that tuple incrementally. Time to finish the job: anything raised inside `importlib.import_module(<user-module>)` is, by definition, an error in the user's code or its third-party imports.

## Fix

Widen both `except` clauses in `src/flyte/_utils/module_loader.py` (file-path branch and directory branch) from the explicit tuple to plain `except Exception`. Wrap as `flyte.errors.ModuleLoadError` (a `RuntimeUserError` already filtered by `flyte/_sentry.py:_is_user_error`) so deploy surfaces an actionable error and Sentry stops receiving user import-time crashes.

## Closes

- [FLYTE-SDK-39](https://unionai.sentry.io/issues/FLYTE-SDK-39/) — `google.api_core.exceptions.RetryError: Timeout of 60.0s exceeded` raised from the user's `gsm_secrets.set_secrets_as_dict_gcp_flyte` during user-module import (expired gcloud creds — `gcloud auth application-default login`).

`fixes FLYTE-SDK-39`

## Tests

- Added a test that a user module raising a non-RuntimeError `Exception` subclass at import time is captured as a load failure (or re-raised as `ModuleLoadError` for the file-path branch).

### DIFF
--- a/src/flyte/_utils/module_loader.py
+++ b/src/flyte/_utils/module_loader.py
@@ -58,16 +58,13 @@ def load_python_modules(
             loaded_modules.append(imported_module)
         except flyte.errors.ModuleLoadError as e:
             failed_paths.append((path, str(e)))
-        except (
-            ImportError,
-            SyntaxError,
-            NameError,
-            AttributeError,
-            TypeError,
-            ValueError,
-            KeyError,
-            RuntimeError,
-        ) as e:
+        except Exception as e:
+            # Anything raised inside ``importlib.import_module(<user-module>)`` is by definition an
+            # error in the user's code or one of its third-party imports (e.g.
+            # ``google.api_core.exceptions.RetryError`` from expired gcloud creds at module top-level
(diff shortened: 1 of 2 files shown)


EngHabu (commented):
LGTM.. not sure how you feel about the blanket `except Exception`

AdilFayyaz (approved):

PR #1173, opened 2026-06-06
(bugfix) Overrides not applied to pods

kumare3:
## Problem

`TaskDetails.override(resources=...)` applied the container resource override **only when the task template had a `container` field**. Pod-template tasks (multi-container environments — e.g. the code-execution sandbox) carry the primary container's resources inside the `k8s_pod.pod_spec` Struct and have **no** `template.container`, so the override was silently skipped. Only `extended_resources` (the GPU accelerator *type* → nodeSelector/toleration) was set.

Result: a per-run `override(resources=Resources(gpu="L4:1"))` produced a pod with the accelerator nodeSelector but **no `nvidia.com/gpu` request**, so it scheduled without a GPU. cpu/memory overrides were dropped the same way.

This was found via the unionai-sandbox GPU notebook override path (follow-up to #1170, which fixed the accelerator-type half of the same `override`).

## Fix

When the template uses `k8s_pod`, apply the resource override to the **primary container inside the pod_spec Struct**, mirroring the build path (`task_serde._get_k8s_pod`): convert the override to k8s resource maps via `_sanitize_resource_name` and replace the primary container's `resources` (full replacement, matching `template.container.resources.CopyFrom`). The primary container is located by `_PRIMARY_CONTAINER_NAME_FIELD` (with a single-container fallback); other containers are left untouched.

## Tests

- `test_gpu_override_on_pod_template_sets_container_request` — GPU override populates the primary container's `requests` (cpu/memory/gpu) and the accelerator type; sidecar untouched.
- `test_override_does_not_mutate_original_pod_template` — override returns a new `TaskDetails` without mutating the source.
- Existing container-path override tests still pass (67 passed in `test_task.py` + `test_pod_template.py`).

### DIFF
--- a/src/flyte/remote/_task.py
+++ b/src/flyte/remote/_task.py
@@ -19,6 +19,7 @@ from flyte._context import internal_ctx
 from flyte._initialize import ensure_client, get_client, get_init_config
 from flyte._internal.runtime.resources_serde import get_proto_extended_resources, get_proto_resources
 from flyte._internal.runtime.task_serde import (
+    _sanitize_resource_name,
     get_proto_max_runtime,
     get_proto_retry_strategy,
     get_proto_timeout_strategy,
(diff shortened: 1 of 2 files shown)


src/flyte/remote/_task.py
```diff
@@ -19,18 +19,65 @@
...
 from flyte._internal.runtime.resources_serde import get_proto_extended_resources, get_proto_resources
 from flyte._internal.runtime.task_serde import (
+    _sanitize_resource_name,
     get_proto_max_runtime,
     get_proto_retry_strategy,
     get_proto_timeout_strategy,
     get_security_context,
 )
 from flyte._logging import logger
+from flyte._pod import _PRIMARY_CONTAINER_NAME_FIELD
```
kumare3: Import this lazily

cosmicBboy (approved):

ursucarina (approved):

PR #1175, opened 2026-06-08
fix(storage): auto-size multipart parts so large uploads don't exceed the 10k-part limit

kumare3:
## Problem

A user uploading a large `.fastq.gz` to GCS (Flyte v2) hit:

```
InvalidArgument: Part number, 10002, must be between 1 and 10000
```

**Root cause — not a GCS or obstore bug.** `flyte.storage.put()` delegated to fsspec → obstore-fsspec `_put_file`, which calls `store.put_async(...)` **without a `chunk_size`**, leaving it pinned to obstore's **5 MiB** default. GCS and S3 both **hard-cap multipart uploads at 10,000 parts**, so 5 MiB × 10,000 = a silent **~48.8 GiB (~52 GB) upload ceiling**. Any file above that fails at `partNumber=10001+`. (The reported "30 GB" file is larger than estimated — part 10002 ≈ a ~52 GB object at 5 MiB parts.)

The upload path exposed **no** way to raise the part size (downloads already have `FLYTE_IO_CHUNK_SIZE` / `MAX_CONCURRENCY` via `ObstoreParallelReader`; uploads had nothing).

## Fix

Auto-scale the multipart part size from the file size so the part count always stays under the limit — **fully automatic, no user-facing knob**:

- `_compute_upload_chunk_size(size)` → `max(5 MiB, ceil(size / 9000))`.
- `_put_obstore_bypass()` calls `obstore.put_async` directly with the computed `chunk_size` (streams from a `pathlib.Path`, no full read into memory); handles single files and recursive directory uploads.
- `put()` routes obstore-protocol uploads (`s3`/`gs`/`abfs`/`abfss`) through the bypass, guarded identically to the existing download bypass; everything else falls back to fsspec unchanged.

Mirrors the existing download-side `_get_obstore_bypass` / `ObstoreParallelReader` pattern.

## Behavior / perf

- Files ≤ ~45 GiB: identical 5 MiB parts → **no change**.
- Files > ~45 GiB: larger parts → **fewer** PUTs (these used to fail outright). A 1 TB file uses ~117 MiB parts and succeeds.
- Only added work per file is one `os.path.getsize` stat — negligible next to the upload.
- Recursive directory uploads keep the same concurrency (`BATCH_SIZE`) and per-file path as before.

## Verification

- Confirmed the recursive bypass produces an **identical destination key layout** to fsspec's `_put` (contents directly under the prefix, no basename nesting) for both the async obstore backend and `LocalFileSystem` — so `DataFrame`/`Dir` round-trips are unaffected.
- New unit tests: chunk-size math (small → 5 MiB floor; 60 GiB → parts < 10,000), put routing through the bypass with auto-sized `chunk_size`, and recursive multi-file layout.
- `tests/internal/storage/` — **58 passed** (1 deselected = the LocalStack `@pytest.mark.sandbox` test, needs a running cluster). Lint + format clean.

### DIFF
--- a/src/flyte/storage/_storage.py
+++ b/src/flyte/storage/_storage.py
@@ -32,6 +32,10 @@ BATCH_SIZE = int(os.getenv("FLYTE_IO_BATCH_SIZE", str(32)))
 # part size imposes a hard file-size ceiling (obstore's 5 MiB default => ~48.8 GiB). We instead
 # auto-scale the part size with the file size so uploads never run out of part numbers, while small
 # files keep small parts. Nothing here is user-tunable -- it just works.
+#
+# The floor and concurrency mirror obstore's own put/put_async defaults (chunk_size=5242880,
+# max_concurrency=12), defined here:
+# https://github.com/developmentseed/obstore/blob/c6279ce19f73eef89321179d55930d0dd7ba6a62/obstore/python/obstore/store.py#L500-L501
 _UPLOAD_CHUNK_FLOOR = 5 * 2**20  # 5 MiB -- obstore's own default; floor for small/multipart files
 _UPLOAD_MAX_CONCURRENCY = 12  # obstore's own default
 _MAX_SAFE_PARTS = 9000  # margin below the cloud 10,000-part hard limit
--- a/src/flyte/storage/_storage.py
+++ b/src/flyte/storage/_storage.py
@@ -28,6 +28,27 @@ if typing.TYPE_CHECKING:
 _OBSTORE_SUPPORTED_PROTOCOLS = ["s3", "gs", "abfs", "abfss"]
 BATCH_SIZE = int(os.getenv("FLYTE_IO_BATCH_SIZE", str(32)))
 
+# Multipart upload sizing. GCS and S3 both hard-cap a multipart upload at 10,000 parts, so a fixed
+# part size imposes a hard file-size ceiling (obstore's 5 MiB default => ~48.8 GiB). We instead
+# auto-scale the part size with the file size so uploads never run out of part numbers, while small
+# files keep small parts. Nothing here is user-tunable -- it just works.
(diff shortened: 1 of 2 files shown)


src/flyte/storage/_storage.py
```diff
@@ -28,6 +28,27 @@
 _OBSTORE_SUPPORTED_PROTOCOLS = ["s3", "gs", "abfs", "abfss"]
 BATCH_SIZE = int(os.getenv("FLYTE_IO_BATCH_SIZE", str(32)))
 
+# Multipart upload sizing. GCS and S3 both hard-cap a multipart upload at 10,000 parts, so a fixed
+# part size imposes a hard file-size ceiling (obstore's 5 MiB default => ~48.8 GiB). We instead
+# auto-scale the part size with the file size so uploads never run out of part numbers, while small
+# files keep small parts. Nothing here is user-tunable -- it just works.
+_UPLOAD_CHUNK_FLOOR = 5 * 2**20  # 5 MiB -- obstore's own default; floor for small/multipart files
```
EngHabu, line 39: Should we add a max chunk size too? it 5TB...

EngHabu (dismissed):

src/flyte/storage/_storage.py
```diff
@@ -28,6 +28,27 @@
...
+_UPLOAD_MAX_CONCURRENCY = 12  # obstore's own default
```
machichima, line 40: nit: could we add link to obstore showing where the default value is defined for future reference?

https://github.com/developmentseed/obstore/blob/c6279ce19f73eef89321179d55930d0dd7ba6a62/obstore/python/obstore/store.py#L500-L501

src/flyte/storage/_storage.py
kumare3: i think we should let gcs etc fail, as this may change

machichima (approved):

PR #1178, opened 2026-06-08
fix(logging): include log level in console log lines

pingsutw:
## Problem

Log lines emitted by reusable container tasks don't show the log level (`WARNING`, `DEBUG`, ...).

Console log records were set up with `fmt="%(message)s"`, so the severity was never rendered. This wasn't caught because the **rich** handler renders the level in its own column — but a reusable container task runs **in-cluster**, where `get_rich_handler()` returns `None` (`ctx.is_in_cluster()`). The plain `StreamHandler` path is used instead, and that path dropped the level entirely.

Before:
```
user warning line
[flyte] internal debug line
```

After:
```
WARNING user warning line
[flyte] DEBUG internal debug line
```

Regression introduced alongside #1038 (`ContextFilter`/`FlyteInternalFilter` → `ContextFormatter`).

## Fix

- Add a shared `DEFAULT_CONSOLE_FORMAT = "%(levelname)s %(message)s"` constant.
- Point every console `ContextFormatter` call site (flyte internal, user, root, and the import-time default loggers) plus the `ContextFormatter` default at it.
- Leave the **rich** handler and the **JSON** formatter untouched — both already emit the level, so this avoids double-printing it.

The level renders after the context markers, before the message:
```
[r1][a0] [flyte] WARNING hello     # internal, with run/action context
INFO user-line                     # user logger, no context
```

## Test plan

- [x] Added 4 unit tests in `tests/user_api/test_logging.py`: constant includes the level, formatter renders the level, marker ordering (`[run][action] [flyte] LEVEL message`), and initialized console handlers emit the level.
- [x] `pytest tests/user_api/test_logging.py tests/flyte/test_logging.py` → 29 passed.
- [x] `ruff check` → clean.

### DIFF
--- a/src/flyte/_internal/runtime/taskrunner.py
+++ b/src/flyte/_internal/runtime/taskrunner.py
@@ -240,6 +240,4 @@ async def extract_download_run_upload(
         logger.info(f"Task {task.name} completed successfully, no outputs")
         return
     await upload_outputs(outputs, output_path) if output_path else None
-    logger.info(
-        f"Task {task.name} completed successfully, uploaded outputs to {output_path} in {time.time() - t:.2f}s"
-    )
+    logger.info(f"Task {task.name} completed successfully, uploaded outputs to {output_path} in {time.time() - t:.2f}s")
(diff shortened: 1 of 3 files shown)


AdilFayyaz (approved):
LGTM

PR #1181, opened 2026-06-08
Fix shell task

k1sauce:
- optional File/Dir flags were emitted for unset inputs (the backend writes a "null" sentinel file, defeating the [ -e ] check);
- File/Dir outputs were read as UTF-8, corrupting binary outputs like gzip;
- and debug=True output never reached remote logs. Fixes all three.


k1sauce:
Also removes bio plugin. After more internal discussion I decided to pull out bio to its own python package. The main reason is testing a large container surface requires running flyte workflows. One of the key draws to doing bio with flyte is the ability to do large scale testing. Also community contributions will be better scoped to a standalone package.

kumare3 (approved):

PR #1182, opened 2026-06-09
init_in_cluster: fall back to init_from_config when config file is present

mhotan:
## Context

Running v2 actions today requires a two-phase deploy: install the dataplane chart, **then** run `flyte create secret EAGER_API_KEY ...` before any action can execute. That side-channel step breaks GitOps for selfhosted users.

`_UNION_EAGER_API_KEY` is also base64-encoded but **not encrypted** on the task pod manifest — readable via `pods/get`, in audit logs, and via `echo \$_UNION_EAGER_API_KEY` inside the user container.

## Change

When no `api_key=` is passed, `init_in_cluster()` checks `UCTL_CONFIG` and `FLYTECTL_CONFIG` env vars in that order. If either is set, it delegates to `init_from_config.aio()` for the SDK side effects (logger, type transformers, sys.path, `_init_config` globals) and returns controller-shaped kwargs derived from the same `PlatformConfig` so `runtime._run_action` can spread them into `create_remote_controller(**kwargs)` — no env var on the pod manifest.

`_platform_to_client_kwargs(pc: PlatformConfig) -> dict` is the single source of truth for translating PlatformConfig fields into the kwargs accepted by `_initialize_client` and `create_remote_controller`. `init_from_config` spreads its output into `init.aio(...)` instead of threading each field by hand. A new PlatformConfig field that needs to reach the HTTP/auth stack is added in one place.

`ca_cert_file_path` takes precedence over `insecure_skip_verify` in the helper. `_resolve_tls_ca_cert` otherwise prefers the bootstrap-from-server path, which fetches only what the leaf presents — nginx typically serves the leaf alone, so the bootstrapped chain has no path back to the trust anchor and pyqwest reports `UnknownIssuer`. The chart-managed CA bundle, when mounted, is unambiguously the correct trust anchor.

### What this PR deliberately does NOT do

- Does **not** walk `resolve_config_path()`'s full precedence chain in `init_in_cluster`. That walk includes a `git rev-parse` subprocess, which is wasted work in a task pod and shows up in `Popen`-mocking tests as an extra call. The chart contract is "the pod's env var points at the mounted file"; honoring just that env var is enough.
- Does **not** add a hardcoded `/etc/flyte/config.yaml` fallback to `resolve_config_path()`. The chart sets `FLYTECTL_CONFIG` explicitly; no fallback default is needed.
- Does **not** introduce a new SDK-level config-loading entry point. `init_from_config()` is the existing public surface; this PR adds one new call site (inside `init_in_cluster`) and refactors `init_from_config`'s internal field threading to use the helper.

A typo in `FLYTECTL_CONFIG` raises `InitializationError("ConfigFileNotFoundError")` rather than silently routing to the legacy api-key branch — keeps misconfiguration loud.

Explicit `api_key=` and the legacy env-var path are unchanged when no env var is set.

## Caveats

Solves the two-phase deploy. **Reduces — does not eliminate — API-key leakage**: the mounted `client_secret` file is still readable from inside the user container. _Tentative:_ a follow-up `union-auth-proxy` sidecar (Phase 2, not implemented, subject to review) is one candidate for closing that remaining gap. Alternatives (capability-only narrowing, short-lived token exchange) are still open.

## Test plan

- [x] `FLYTECTL_CONFIG` set → delegates and returns controller-shaped kwargs
- [x] `UCTL_CONFIG` takes precedence over `FLYTECTL_CONFIG`
- [x] Neither env var set → legacy api-key/env-var branch runs
- [x] Env var points at missing path → loud `InitializationError`, no silent fallback
- [x] Explicit `api_key=` → no fallback even when env var is set
- [x] Helper round-trips every chart-set `PlatformConfig` field
- [x] Helper picks `ca_cert_file_path` over `insecure_skip_verify` when both set
- [x] End-to-end: v2 eager run on Union selfmanaged dataplane (mike-apple-aws), file-mounted creds, internal-CA TLS, no env var, first-attempt success — run [`ucbzldwjldgtq9fwxchs`](https://mike-andromeda-aws.sh-us-west-2.union.ai/v2/domain/development/project/union-health-monitoring/runs/ucbzldwjldgtq9fwxchs), 4/4 actions Success.

### DIFF
--- a/src/flyte/_initialize.py
+++ b/src/flyte/_initialize.py
@@ -17,6 +17,7 @@ from ._logging import LogFormat, initialize_logger, logger
 if TYPE_CHECKING:
     from flyte._internal.imagebuild import ImageBuildEngine
     from flyte.config import Config
+    from flyte.config._config import PlatformConfig
     from flyte.remote._client.auth import AuthType, ClientConfig
     from flyte.remote._client.controlplane import ClientSet
     from flyte.storage import Storage
@@ -54,6 +55,54 @@ class _InitConfig(CommonInit):
 # Global singleton to store initialization configuration
 _init_config: _InitConfig | None = None
 _init_lock = threading.RLock()  # Reentrant lock for thread safety
+
+
+def _platform_to_client_kwargs(pc: "PlatformConfig") -> dict[str, typing.Any]:
+    """Translate PlatformConfig fields into kwargs accepted by both
+    _initialize_client (via init) and create_remote_controller. Single
+    source of truth — extending the HTTP/auth stack with a new
+    PlatformConfig field means editing this one function.
+
+    Callers may need to layer on caller-specific kwargs that are NOT
+    derived from PlatformConfig (api_key, headless, in-cluster env-var
+    overrides); those stay at the call site.
+
+    ca_cert_file_path takes precedence over insecure_skip_verify when
+    both are present. _resolve_tls_ca_cert otherwise prefers the
+    bootstrap path, which fetches only what the server's leaf presents
+    and trips "UnknownIssuer" on chains nginx serves without
+    intermediates.
+    """
+    kw: dict[str, typing.Any] = {}
+    if pc.endpoint:
+        kw["endpoint"] = pc.endpoint
+    if pc.insecure:
+        kw["insecure"] = True
+    if pc.ca_cert_file_path:
+        kw["ca_cert_file_path"] = pc.ca_cert_file_path
+    elif pc.insecure_skip_verify:
+        kw["insecure_skip_verify"] = True
+    if pc.client_id:
+        kw["client_id"] = pc.client_id
+    if pc.client_credentials_secret:
+        kw["client_credentials_secret"] = pc.client_credentials_secret
+    if pc.auth_mode:
+        kw["auth_type"] = pc.auth_mode
+    if pc.command:
+        kw["command"] = pc.command
+    if pc.proxy_command:
+        kw["proxy_command"] = pc.proxy_command
+    if pc.http_proxy_url:
+        kw["http_proxy_url"] = pc.http_proxy_url
+    if pc.rpc_retries:
+        kw["rpc_retries"] = pc.rpc_retries
+    # NOTE: disable_keyring is intentionally NOT emitted here. The helper
+    # output flows both into _initialize_client (via init) AND into
+    # create_remote_controller (via init_in_cluster). The controller's
+    # constructor doesn't accept disable_keyring, so emitting it would
+    # raise TypeError on the in-cluster path. init_from_config threads
+    # disable_keyring through to init separately, alongside this helper.
+    return kw
 
 
 async def _initialize_client(
@@ -283,6 +332,7 @@ async def init_from_config(
     log_level: int | None = None,
     log_format: LogFormat = "console",
     user_log_level: int | None = None,
+    org: str | None = None,
     project: str | None = None,
     domain: str | None = None,
     storage: Storage | None = None,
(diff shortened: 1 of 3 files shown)


kumare3:
### Code review

Found 1 issue:

1. The new config-file fallback drops the `org` parameter. The legacy branch resolves `org = org or os.getenv("_U_ORG_NAME")` and passes it to `init.aio(org=org, ...)`, but the fallback calls `init_from_config.aio(...)` with only `project` and `domain` — `init_from_config` has no `org` parameter and derives org solely from `cfg.task.org`. Since `runtime.py` calls `init_in_cluster(org=org, ...)` with org from `_U_ORG_NAME`, an org passed by the runtime (or explicitly by a caller) is silently ignored whenever a config file is mounted, and the config file's org (often unset) wins instead.

https://github.com/flyteorg/flyte-sdk/blob/f8ca1e74668e2346cd82d4811efad633eb5cc7a5/src/flyte/_initialize.py#L530-L536

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

src/flyte/_initialize.py
```diff
@@ -56,6 +57,50 @@ def replace(self, **kwargs) -> _InitConfig:
...
+    if pc.command:
+        kw["command"] = pc.command
+    if pc.proxy_command:
+        kw["proxy_command"] = pc.proxy_command
+    if pc.http_proxy_url:
+        kw["http_proxy_url"] = pc.http_proxy_url
+    if pc.rpc_retries:
+        kw["rpc_retries"] = pc.rpc_retries
+    if pc.disable_keyring:
+        kw["disable_keyring"] = pc.disable_keyring
```
kumare3: this is not a safe keyword arg for controller, you should probably drop this

src/flyte/_initialize.py
```diff
@@ -472,6 +508,44 @@ async def init_in_cluster(
...
+        cfg_path_str = os.getenv(UCTL_CONFIG_ENV_VAR) or os.getenv(FLYTECTL_CONFIG_ENV_VAR)
+        if cfg_path_str:
+            # Existence is intentionally NOT pre-checked: a typo in the
+            # env var should surface as the FileNotFoundError that
+            # init_from_config raises when it tries to open the path,
+            # not silently fall back to the legacy api-key branch.
+            logger.info(f"init_in_cluster: delegating to init_from_config({cfg_path_str})")
+            # init_from_config is @syncify-decorated; call its async form so we don't
+            # block the syncify thread.
+            await init_from_config.aio(
```
kumare3, line 542: can you add org here

mhotan:
Thanks @kumare3 — both addressed in `7aec70fe1`:

**1. `org` propagation through the fallback** (line 531 + the summary comment):
- Added `org: str | None = None` to `init_from_config`'s public signature, forwarded as `org=org or cfg.task.org` to `init.aio(...)`. The runtime-passed org now wins over the config file's `cfg.task.org`, matching the legacy branch's semantics.
- `init_in_cluster`'s fallback threads `org=org or os.getenv(ORG_NAME)` into `init_from_config.aio(...)` so a runtime-provided org (or `_U_ORG_NAME`) survives the fallback.
- Two new regression tests: `test_org_propagates_from_env_to_init_from_config` and `test_explicit_org_kwarg_wins_over_env_in_fallback`.

**2. `disable_keyring` dropped from `_platform_to_client_kwargs`** (line 100):
- The helper's output is spread into both `_initialize_client` (via `init`) AND `create_remote_controller` (via `init_in_cluster`). The controller's constructor doesn't accept `disable_keyring` — emitting it would have raised TypeError on every in-cluster task pod.
- `init_from_config` now threads `disable_keyring=cfg.platform.disable_keyring` to `init.aio(...)` separately, outside the helper, so the laptop/CLI path still honors the user's setting.
- New `test_platform_kwargs_helper_omits_disable_keyring` drift catcher asserts the helper output doesn't include the key.

Net diff vs. previous: +64/-3.

All tests pass (52 init-related, mypy clean, `test_main_command_execution` green).

kumare3 (approved):

PR #1184, opened 2026-06-10
fix: clear error for missing .dockerignore in with_dockerignore

EngHabu:
## Problem

`Image.with_dockerignore(path)` stored the path as-is, and both build paths copied it into the build context with a bare `shutil.copy(layer.path, context_path)`. When the file wasn't there at build time — typically running `flyte deploy` from outside the project root — that raised a raw, uninformative `FileNotFoundError` which leaked to Sentry as if it were an SDK bug.

Sentry: **FLYTE-SDK-4Z** (`FileNotFoundError: [Errno 2] No such file or directory: '.dockerignore'`, observed on 2.4.4).

## Fix

Validate at **build time only** — in `remote_builder._get_layers_proto` and `docker_builder.DockerIgnoreHandler.handle`. Those are the only two places the file genuinely has to exist, and both raise a clear **`ImageBuildError`** (a `RuntimeUserError`, so it's shown to the user and filtered from Sentry) instead of the raw `FileNotFoundError`.

`with_dockerignore()` itself is unchanged from `main`.

### Why no check at definition time

An earlier revision validated inside `with_dockerignore()`. Per [@pingsutw's review](https://github.com/flyteorg/flyte-sdk/pull/1184#discussion_r3390319883), that was wrong: image definitions are module-level code that also executes **inside the container at task runtime**, where the developer's `.dockerignore` is absent — so a definition-time check would raise spuriously. `DockerIgnore.update_hash` already falls back to hashing the path string when the file is missing, confirming "may not exist" was the intended contract.

That revision also resolved the path to absolute. Dropped as well: it wasn't load-bearing. The Sentry stack shows the copy failing in `_get_layers_proto`, which runs client-side in the same process that defined the image, so there's no cwd mismatch to correct.

## Tests

- `tests/flyte/test_image.py` — a missing `.dockerignore` must **not** raise at definition time, and `img.uri` must still compute (locks in the contract above).
- `tests/flyte/imagebuild/test_docker_builder.py` — the build-time handler raises `ImageBuildError` for a missing file; happy-path copy still works.

`tests/flyte/test_image.py` + `tests/flyte/imagebuild/` → 184 passed. (`test_image_build_engine.py::test_persistent_cache_write_and_read` fails identically on `main` — pre-existing test-isolation issue, unrelated.)

fixes FLYTE-SDK-4Z

### DIFF
--- a/src/flyte/_image.py
+++ b/src/flyte/_image.py
@@ -1278,7 +1278,17 @@ class Image:
         return self.clone(addl_layer=CodeBundleLayer(copy_style=copy_style, dst=dst))
 
     def with_dockerignore(self, path: Path) -> Image:
-        new_image = self.clone(addl_layer=DockerIgnore(path=str(path)))
+        import flyte.errors
+
+        # Resolve to an absolute path so the file can still be located at build time, which may run
+        # from a different working directory than where the image was defined (e.g. the remote builder).
+        resolved = Path(path).expanduser()
+        if not resolved.is_file():
+            raise flyte.errors.ImageBuildError(
+                f"The .dockerignore file specified via with_dockerignore() was not found at '{path}'. "
+                f"Ensure the path points to an existing file."
+            )
+        new_image = self.clone(addl_layer=DockerIgnore(path=str(resolved.resolve())))
         return new_image
 
     def with_uv_project(
--- a/src/flyte/_image.py
+++ b/src/flyte/_image.py
@@ -1278,17 +1278,12 @@ class Image:
         return self.clone(addl_layer=CodeBundleLayer(copy_style=copy_style, dst=dst))
 
     def with_dockerignore(self, path: Path) -> Image:
-        import flyte.errors
-
-        # Resolve to an absolute path so the file can still be located at build time, which may run
-        # from a different working directory than where the image was defined (e.g. the remote builder).
-        resolved = Path(path).expanduser()
-        if not resolved.is_file():
-            raise flyte.errors.ImageBuildError(
-                f"The .dockerignore file specified via with_dockerignore() was not found at '{path}'. "
-                f"Ensure the path points to an existing file."
-            )
-        new_image = self.clone(addl_layer=DockerIgnore(path=str(resolved.resolve())))
+        # Deliberately no existence check here: image definitions are module-level code that also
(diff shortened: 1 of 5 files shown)


src/flyte/_image.py
```diff
@@ -1180,7 +1180,17 @@ def with_code_bundle(
...
+        import flyte.errors
+
+        # Resolve to an absolute path so the file can still be located at build time, which may run
+        # from a different working directory than where the image was defined (e.g. the remote builder).
+        resolved = Path(path).expanduser()
+        if not resolved.is_file():
+            raise flyte.errors.ImageBuildError(
+                f"The .dockerignore file specified via with_dockerignore() was not found at '{path}'. "
+                f"Ensure the path points to an existing file."
+            )
```
pingsutw: I think we don't need the check here since the `dockerignore` file may not exist in the image at runtime

EngHabu (commented):
LGTM

src/flyte/_image.py
EngHabu: Good catch — you're right, and I've removed it.

Image definitions are module-level code that also executes inside the container at task runtime, where the developer's `.dockerignore` is absent, so a definition-time check would raise spuriously. `DockerIgnore.update_hash` already falls back to hashing the path string when the file is missing, which confirms "may not exist" was the intended contract all along.

`with_dockerignore()` is now byte-identical to `main` — I dropped the absolute-path resolution too, since it was not load-bearing (the Sentry event shows the copy failing in `_get_layers_proto`, which runs client-side in the same process as the image definition, so there is no cwd mismatch to correct; the user had simply run `flyte deploy` from outside their project root).

The FLYTE-SDK-4Z crash is still fixed by the two **build-time** guards, in `remote_builder._get_layers_proto` and `DockerIgnoreHandler.handle` — the only places the file genuinely has to exist. The raw `FileNotFoundError: '.dockerignore'` from `shutil.copy` becomes a clear `ImageBuildError` (a `RuntimeUserError`, so it is shown to the user and filtered out of Sentry).

Tests updated to lock in the new contract: definition time must **not** raise for a missing file (and `img.uri` must still compute), while the build-time handler must raise `ImageBuildError`. Added a happy-path copy test as well.

pingsutw (approved):

PR #1194, opened 2026-06-11
fix: raise InitializationError for invalid auth mode

EngHabu:
## Problem

An unrecognized auth mode in the creds config raised a bare `ValueError` from the authenticator factory (`_authenticators/factory.py`). During `SelectCluster` it gets wrapped as `RuntimeError("SelectCluster failed...")` → `RuntimeSystemError` and leaked to Sentry as an SDK crash with no actionable signal — even though it's a user-config mistake.

Sentry: **FLYTE-SDK-50** — `RuntimeSystemError: ... Invalid auth mode [None] specified. Please update the creds config to use a valid value`

## Fix

Raise a typed `InitializationError` (a user-facing error already in the Sentry filter's allowlist) instead of a bare `ValueError`. The Sentry user-error filter walks the cause chain and now skips it, and the user still gets the clear 'update the creds config' message.

## Tests

- `test_get_async_authenticator_invalid_mode_raises_initialization_error` — factory raises `InitializationError` for `None`/bogus modes.
- `test_capture_exception_skips_invalid_auth_mode_wrapped_as_system_error` — the wrapped `RuntimeSystemError` chain is filtered from Sentry.

fixes FLYTE-SDK-50

### DIFF
--- a/src/flyte/remote/_client/auth/_authenticators/factory.py
+++ b/src/flyte/remote/_client/auth/_authenticators/factory.py
@@ -163,8 +163,16 @@ def get_async_authenticator(
 
             return PassthroughAuthenticator(endpoint=endpoint, **kwargs)
         case _:
-            raise ValueError(
-                f"Invalid auth mode [{auth_type}] specified. Please update the creds config to use a valid value"
+            # An unrecognized auth mode is a user-config mistake, not an SDK crash.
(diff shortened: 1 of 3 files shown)


EngHabu (commented):
LGTM

pingsutw (approved):

PR #1195, opened 2026-06-11
fix: skip files that vanish before being added to code bundle

EngHabu:
## Problem

The code bundle file list (`ls`) is computed by walking the source tree, but a file can disappear between listing and `tar.add` stat-ing it — most commonly a transient lock file (e.g. `.codegraph/codegraph.lock`). This raised a raw `FileNotFoundError` in `tarfile.gettarinfo` that aborted the entire deploy.

Sentry: **FLYTE-SDK-52** — `FileNotFoundError: [Errno 2] No such file or directory: '.../.codegraph/codegraph.lock'` (culprit `tarfile in gettarinfo`, via `_packaging.create_bundle`).

## Fix

Wrap `tar.add` in a `try/except FileNotFoundError` that logs a warning and skips the vanished file. A file disappearing mid-bundle is an environment race, not an SDK bug, and shouldn't abort the whole deploy — the rest of the bundle is still produced.

## Tests

- `test_create_bundle_skips_files_that_vanish_before_add` — a file present in the list but missing on disk is skipped; the bundle is still produced with the remaining files.

fixes FLYTE-SDK-52

### DIFF
--- a/src/flyte/_code_bundle/_packaging.py
+++ b/src/flyte/_code_bundle/_packaging.py
@@ -164,12 +164,21 @@ def create_bundle(
                 # entry shouldn't abort the entire bundle.
                 logger.warning(f"Skipping {ws_file}: resolves outside source root {abs_source}")
                 continue
-            tar.add(
-                os.path.join(source, ws_file),
-                recursive=False,
-                arcname=rel_path,
-                filter=tar_strip_file_attributes,
-            )
+            try:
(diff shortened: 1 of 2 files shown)


EngHabu (commented):
LGTM.. I'm guessing some of these maybe symlinked or something?

pingsutw (approved):

PR #1198, opened 2026-06-12
Create run with TaskId for lazy task

popojk:
## Goal
While creating runs with remote task(lazy task), we should create run request with TaskID instead of TaskSpec. Thus, the user only needs execution permission to do such an operation. 

p.s. While executing ephemeral runs(create runs with TaskSpec), the user must process write permission because the BE will create TaskSpec in the DB.

## Tests
Add log to see if we are creating run with TaskID or TaskSpec:
Screenshot 2026-06-12 at 10 31 02 AM

Test with this script to run remote task:
```
import flyte

if __name__ == "__main__":
    flyte.init_from_config()
    task = flyte.remote.Task.get("named_fanout.fanout", auto_version="latest")
    result = flyte.run(task, n1=2, n2=3, n3=4)
    print(f"Result: {result}")
```
The log shows we create run with TaskID
Screenshot 2026-06-12 at 10 41 14 AM

Execute an ephemeral run, the log show we create run with TaskSpec:
Screenshot 2026-06-12 at 10 43 04 AM

Test if task override works by overriding the task Short_name:
```
import flyte

if __name__ == "__main__":
    flyte.init_from_config()
    task = flyte.remote.Task.get("named_fanout.fanout", auto_version="latest").override(short_name="overridden_my_task")
    result = flyte.run(task, n1=2, n2=3, n3=4)
    print(f"Result: {result}")
```

The name is overridden:
Screenshot 2026-06-15 at 11 14 41 AM

## Rollout plan
The backend permission check changes is till processing. It is ok to merge this PR first as current BE authz logic will not block create run request with TaskID.

### DIFF
--- a/src/flyte/_run.py
+++ b/src/flyte/_run.py
@@ -182,12 +182,14 @@ class _Runner:
         domain = self._domain or cfg.domain
 
         task: TaskTemplate[P, R, F] | TaskDetails
+        task_id = None
         if isinstance(obj, (LazyEntity, TaskDetails)):
             if isinstance(obj, LazyEntity):
                 task = await obj.fetch.aio()
             else:
                 task = obj
             task_spec = task.pb2.spec
+            task_id = task.pb2.task_id
             inputs = await convert_from_native_to_inputs(
                 task.interface, *args, custom_context=self._custom_context, **kwargs
             )
@@ -352,14 +354,17 @@ class _Runner:
                 )
             # Fill in task id inside the task template if it's not provided.
             # Maybe this should be done here, or the backend.
-            if task_spec.task_template.id.project == "":
-                task_spec.task_template.id.project = project or ""
-            if task_spec.task_template.id.domain == "":
-                task_spec.task_template.id.domain = domain or ""
-            if task_spec.task_template.id.org == "":
-                task_spec.task_template.id.org = cfg.org or ""
-            if task_spec.task_template.id.version == "":
-                task_spec.task_template.id.version = version
+            # Only needed for locally-defined tasks; a fetched task (task_id set) already
+            # carries a fully-populated id and is sent by reference, not as a spec.
+            if task_id is None:
+                if task_spec.task_template.id.project == "":
+                    task_spec.task_template.id.project = project or ""
+                if task_spec.task_template.id.domain == "":
+                    task_spec.task_template.id.domain = domain or ""
+                if task_spec.task_template.id.org == "":
+                    task_spec.task_template.id.org = cfg.org or ""
+                if task_spec.task_template.id.version == "":
+                    task_spec.task_template.id.version = version
 
             kv_pairs: List[literals_pb2.KeyValuePair] = []
             for k, v in env.items():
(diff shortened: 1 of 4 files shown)


src/flyte/_run.py
```diff
@@ -352,14 +354,17 @@ async def _run_remote(self, obj: TaskTemplate[P, R, F] | LazyEntity, *args: P.ar
...
-            if task_spec.task_template.id.project == "":
-                task_spec.task_template.id.project = project or ""
-            if task_spec.task_template.id.domain == "":
-                task_spec.task_template.id.domain = domain or ""
-            if task_spec.task_template.id.org == "":
-                task_spec.task_template.id.org = cfg.org or ""
-            if task_spec.task_template.id.version == "":
-                task_spec.task_template.id.version = version
+            # Only needed for locally-defined tasks; a fetched task (task_id set) already
+            # carries a fully-populated id and is sent by reference, not as a spec.
```
pingsutw: does it work with task override?

popojk: Thanks for noticing. Fixed in [c0ab563](https://github.com/flyteorg/flyte-sdk/pull/1198/commits/c0ab5639af951b25828d0b96e02cae8c45432539). I also tested it and put the result in PR description.

src/flyte/remote/_task.py
```diff
@@ -165,6 +165,7 @@ class TaskDetails(ToJSONMixin):
     pb2: task_definition_pb2.TaskDetails
     max_inline_io_bytes: int = 10 * 1024 * 1024  # 10 MB
     overriden_queue: Optional[str] = None
+    overridden: bool = False  # Flag if the TaskDetails was overridden
```
popojk, line 168: Add this var in TaskDetails to flag if the fetched TaskDetail is overridden. If it is overridden, we should still send full TaskSpec to the backend while creating runs.

kumare3:
so does `flyte run remote-task` not work already?

popojk:
> so does `flyte run remote-task` not work already?

@kumare3 Right now `the flyte run remote-task` will send the full TaskSpec to the BE while creating runs. However, we could just send TaskID to the BE as the task is already deployed, and that is the point of this PR. We made RBAC changes in the BE so ephemeral runs(create runs with TaskSpec) will require Deploy_Inventory permission because the BE will create TaskSpec data in the DB for ephemeral runs.

src/flyte/_run.py
```diff
@@ -399,46 +409,53 @@ def _to_cache_lookup_scope(scope: CacheLookupScope | None = None) -> run_pb2.Cac
...
-                            ),
-                            notification_rule_name=notification_rule_name,
-                            notification_rules=notification_rules,
                         ),
+                        notification_rule_name=notification_rule_name,
+                        notification_rules=notification_rules,
                     ),
                 )
+                # Reference an already-registered task by id; otherwise send the full spec.
+                # The user should have Action_ACTION_REGISTER_FLYTE_INVENTORY permission while sending the full spec.
```
pingsutw: nit: could we remove it since it‘s closed source thing
```suggestion
```

popojk: Sure, fixed!

pingsutw (approved):

PR #1204, opened 2026-06-12
fix: clustered (JobSet) restart — non-zero exit on failure + terminal-only error.pb

AdilFayyaz:
## Motivation

Clustered (JobSet) tasks restart the **whole pod set** up to `failure_policy.max_restarts` times *inside a single Flyte attempt*. Two things were broken in that flow:

1. **`error.pb` was written on every failed attempt** to the same output path. A later successful restart left the stale `error.pb` behind, so the run could be misreported as FAILED despite succeeding. (The original version of this PR worked around it by *deleting* the stale file on success — fragile, and it left a window where the error could be read mid-restart.)
2. **A task failure never produced a non-zero pod exit.** The runtime's contract is "write `error.pb`, exit 0" (correct for normal tasks, where the backend reads `error.pb`). But the JobSet controller keys its gang-restart off the **pod exit code**, not `error.pb`. So a clustered worker that hit an exception exited 0 → torchrun exited 0 → the Job "succeeded" → **the JobSet never restarted**, even with `max_restarts > 0`.

## Summary

Replaces the delete-on-success workaround with two complementary fixes:

**1. Write `error.pb` only on the terminal JobSet attempt** (`io.py`)
- `upload_error` skips the write on transient restarts and only writes once the restart budget is exhausted (`JOBSET_RESTART_ATTEMPT >= JOBSET_MAX_RESTARTS`).
- New helpers `_get_clustered_max_restarts()` / `_is_terminal_clustered_attempt()`. Safe fallback: if the budget env is absent (older operator) or the task isn't clustered, it writes as before — errors are never silently hidden.
- Removed `_clear_stale_clustered_error_if_needed` / `_delete_path` (no stale file is ever written now, so there's nothing to delete).

**2. Clustered workers exit non-zero on failure** (`taskrunner.py`)
- After the (now gated) `error.pb` handling, a clustered worker (`TORCHELASTIC_RUN_ID` present) calls `sys.exit(1)` so torchrun → Job → JobSet detects the failure and triggers the gang restart. Normal tasks keep the exit-0 contract.
- Added `_is_clustered_worker()` (refactored out of `_is_nonzero_rank_clustered_worker`).

Together: the non-zero exit makes the JobSet climb through restart attempts; the terminal-only `error.pb` write keeps those transient attempts from poisoning the eventual success.

## Test Plan

- **Unit** — `pytest tests/flyte/internal/runtime/test_taskrunner.py tests/flyte/clustered/`:
  - `upload_error` gating: transient attempt → no write; terminal attempt → write; non-clustered → write; missing-budget-env fallback → write; `max_restarts=0` → write; non-zero rank → no write; `upload_outputs` no longer deletes.
  - exit behavior: clustered failure → `SystemExit(1)`; non-clustered failure → no exit; clustered success → no exit.
- **End-to-end** — `examples/clustered/ddp_train_restart.py` (`max_restarts=1`, fails on attempt 0): with the new wheel **and** the operator deployed, attempt 0 exits non-zero → JobSet restarts → attempt 1 trains → run **SUCCEEDED** with no stale `error.pb`.


src/flyte/_internal/runtime/io.py
```diff
@@ -34,6 +36,66 @@ def _is_nonzero_rank_clustered_worker() -> bool:
...
+        return None
+
+
+async def _delete_path(path: str) -> None:
+    fs = storage.get_underlying_filesystem(path=path)
+    if isinstance(fs, AsyncFileSystem):
+        if hasattr(fs, "_rm_file"):
+            await fs._rm_file(path)  # pylint: disable=W0212
+            return
+        if hasattr(fs, "_rm"):
```
pingsutw: do we need `rm`? from my claude:

```
  _delete_path should follow the same shape. I verified the methods exist on the backends you actually hit:
  - AsyncFileSystem._rm_file is defined on the base, and obstore's FsspecStore implements _rm_file directly (S3/GCS/ABFS).
  - sync LocalFileSystem has rm_file.

  So the whole thing collapses to:

  async def _delete_path(path: str) -> None:
      fs = storage.get_underlying_filesystem(path=path)
      if isinstance(fs, AsyncFileSystem):
          await fs._rm_file(path)  # pylint: disable=W0212
      else:
          fs.rm_file(path)
```

src/flyte/_internal/runtime/io.py
```diff
@@ -34,6 +36,66 @@ def _is_nonzero_rank_clustered_worker() -> bool:
...
+    raise NotImplementedError(f"Filesystem {type(fs).__name__} does not support deleting paths")
+
+
+async def _clear_stale_clustered_error_if_needed(output_path: str) -> None:
+    restart_attempt = _get_clustered_restart_attempt()
+    if restart_attempt is None or restart_attempt <= 0:
+        return
+
+    stale_error_uri = error_path(output_path)
+    if not await storage.exists(stale_error_uri):
```
pingsutw: Why do we need it? Don't we write output.pb and error.pb to different folder (<raw_output>/0, <raw_output>/1) for different attempts?

AdilFayyaz: Because a Jobset restart attempt is a k8s Jobsets restart within the same attempt. This is different from flyte attempts because they go to different output folders. For Jobsets, the crashed errors.pb and successful restarts error.pb go in the same folder. 
But are you suggesting we write per restart outputs to a jobsets suffixed restart attempt subfolder instead of deleting? Will have to scope that out and would require plugin changes.

src/flyte/_internal/runtime/io.py
AdilFayyaz: this is cleaner I agree

src/flyte/_internal/runtime/io.py
pingsutw: If we use Jobsets restart, do we see the attempt (1/2/3/4) on UI

pingsutw: if we delete the stale error.pb, how do see the error on UI for the previous attempt

AdilFayyaz: Nope you wont see attempt 1/2/3/4 because this is a Jobset level restart not a Flyte one. And regarding the deletion, we only clear error.pb when the run is about to write to outputs.pb (on success - i.e. on recovery). If the run fails it goes through a different path are returns before deletion so the real error is seen on the UI.

pingsutw: cc @EngHabu @kumare3 do you have any suggestions? I'm not sure if we need to use jobset restart policy or not.

kumare3: What is the benefit of jobset restart policy.
My preference is to use one - flyte restart system for better UX and observability

pingsutw (approved):

PR #1205, opened 2026-06-13
feat(pod): unprivileged allow_fuse() via FUSE device plugin + volume examples

EngHabu:
## What

Two things, together:

1. **`PodTemplate.allow_fuse()` now grants unprivileged FUSE via a device plugin by default** — `smarter-devices/fuse` resource request + `CAP_SYS_ADMIN`, no privileged container, no `/dev/fuse` hostPath. `allow_fuse(privileged=True)` is kept as a legacy escape hatch (hostPath + privileged) for clusters without a device plugin. The two paths are cleanly separated (`_apply_fuse_device_plugin` / `_apply_fuse_privileged`).
2. **JuiceFS `Volume` examples** (`examples/volumes/`) updated to use `pod_template=flyte.PodTemplate().allow_fuse()` instead of the old privileged/`enable_fuse_mount` approach.

## Why

The old `allow_fuse(privileged=False)` path used a `/dev/fuse` hostPath, which doesn't actually work: a hostPath surfaces the device node but the container's **devices cgroup** still denies `open()` with `EPERM`. Only `privileged` (or a device plugin) grants access. The device-plugin path is the real unprivileged solution — requesting `smarter-devices/fuse` makes kubelet inject `/dev/fuse` into the devices-cgroup allowlist; `CAP_SYS_ADMIN` covers `mount(2)`. Composes with `allow_nested_sandboxing()`.

The cluster must run a FUSE device plugin advertising `smarter-devices/fuse` — the Union dataplane chart ships an opt-in `fuseDevicePlugin` DaemonSet (unionai/helm-charts#443, unionai/cloud#16489).

## Validated end-to-end

On a real Union dataplane (dogfood, device plugin deployed): an unprivileged task pod built from `PodTemplate().allow_fuse()` (CAP_SYS_ADMIN + `smarter-devices/fuse`, not privileged, no hostPath) opened `/dev/fuse` and mounted an S3-backed **flyteplugins-union `Volume`** with a read/write round-trip. `CapEff=a82425fb` (SYS_ADMIN only).

## Tests

`TestAllowFuse` (device-plugin default) + `TestAllowFusePrivileged` (legacy hostPath escape hatch) + updated sandboxing-composition tests — 54 passed.

### DIFF
--- a/src/flyte/_pod.py
+++ b/src/flyte/_pod.py
@@ -12,10 +12,17 @@ if TYPE_CHECKING:
 _PRIMARY_CONTAINER_NAME_FIELD = "primary_container_name"
 _PRIMARY_CONTAINER_DEFAULT_NAME = "primary"
 
-# Name used for the hostPath volume and its volumeMount added by
-# ``PodTemplate.allow_fuse()``.
+# Name/path for the hostPath volume + volumeMount used by the legacy
+# ``allow_fuse(privileged=True)`` escape hatch.
 _FUSE_DEVICE_VOLUME_NAME = "fuse-device"
 _FUSE_DEVICE_PATH = "/dev/fuse"
+
+# Extended resource advertised by a FUSE device plugin (smarter-device-manager /
+# fuse-device-plugin DaemonSet). Requesting it makes kubelet inject /dev/fuse
+# into the container's devices-cgroup allowlist — granting an UNPRIVILEGED
+# container access to the device, which a hostPath cannot do. This is what the
+# default (device-plugin) path of ``PodTemplate.allow_fuse()`` uses.
+_FUSE_DEVICE_RESOURCE = "smarter-devices/fuse"
 
 # Every capability helper stamps ``flyte.org/capability-<name>: "true"`` on the
 # resulting template so the requested grant is auditable on the pod itself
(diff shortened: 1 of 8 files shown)


src/flyte/_pod.py
```diff
@@ -248,40 +274,42 @@ def _set_apparmor_unconfined(pt: PodTemplate, caller: str) -> None:
...
+        # Default device-plugin path: request smarter-devices/fuse so kubelet
+        # injects /dev/fuse into the devices cgroup; CAP_SYS_ADMIN for mount(2).
+        # No privileged, no hostPath. Composes with allow_nested_sandboxing()
+        # (SYS_ADMIN + allowPrivilegeEscalation=false is permitted).
+        _add_fuse_device_resource(primary)
+        _add_sys_admin(primary)
+        _stamp_capability(pt, "fuse")
+        return pt
+
+    # Legacy escape hatch (privileged=True): /dev/fuse hostPath + privileged, for
```
kumare3: is this weird that this will fallback on randomly, i think we should cleanly separate 2 methods.
if privileged
     _fallback_to_legacy
else:
   enable_fuse_device

EngHabu: Good call — split into two clearly separated helpers in `de82835c`: `_apply_fuse_device_plugin(primary)` (default, unprivileged) and `_apply_fuse_privileged(pt, primary)` (legacy hostPath + privileged escape hatch). `_apply_fuse` just dispatches on the flag; the shared CAP_SYS_ADMIN grant + capability annotation stamp run once after. No inline fallback.

kumare3 (approved):

PR #1206, opened 2026-06-13
examples(volumes): use PodTemplate().allow_fuse() instead of enable_fuse_mount

EngHabu:
Switches the volume examples to `pod_template=flyte.PodTemplate().allow_fuse()`. With flyteorg/flyte-sdk#1205, `allow_fuse()` grants unprivileged FUSE via the device plugin (smarter-devices/fuse request + CAP_SYS_ADMIN, no privileged/hostPath), so the examples no longer need the `enable_fuse_mount` flag. Base is the volume-examples branch; rebase onto main once #1205 lands.

### DIFF
--- a/examples/volumes/bench.py
+++ b/examples/volumes/bench.py
@@ -63,8 +63,11 @@ image = with_high_throughput_volume_deps(base)
 
 env = flyte.TaskEnvironment(
     name="vol-bench",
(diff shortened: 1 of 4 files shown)


EngHabu:
Folding these examples into #1205 so the allow_fuse() change and the examples that use it ship as a single PR. Closing in favor of #1205.

PR #1213, opened 2026-06-16
[Image Build] Enable setting platform in `from_base` and switch to `flyte` user only if exists

machichima:
## Description

This PR fixes:

1. Enable setting `platform` when using `from_base`, otherwise only `linux/amd` image is built and cannot run on devbox in macOS
2. Only switch to `USER flyte` only if being created, to prevent `no users found` error

image

### DIFF
--- a/examples/plugins/spark_dataframe_example.py
+++ b/examples/plugins/spark_dataframe_example.py
@@ -9,8 +9,14 @@ import flyte
 from flyte.io import File
 
 image = (
-    flyte.Image.from_base("apache/spark-py:v3.4.0", platform=("linux/amd64", "linux/arm64"))
-    .clone(name="spark", python_version=(3, 10), registry="ghcr.io/flyteorg", extendable=True)
+    flyte.Image.from_base("apache/spark-py:v3.4.0")
+    .clone(
+        name="spark",
+        python_version=(3, 10),
+        registry="ghcr.io/flyteorg",
+        extendable=True,
+        platform=("linux/amd64", "linux/arm64"),
+    )
     .with_pip_packages("flyteplugins-spark")
     .with_pip_packages("pandas", "pyarrow")
 )
--- a/examples/plugins/spark_dataframe_example.py
+++ b/examples/plugins/spark_dataframe_example.py
@@ -9,7 +9,7 @@ import flyte
 from flyte.io import File
 
 image = (
-    flyte.Image.from_base("apache/spark-py:v3.4.0")
+    flyte.Image.from_base("apache/spark-py:v3.4.0", platform=("linux/amd64", "linux/arm64"))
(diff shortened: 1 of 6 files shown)


src/flyte/_image.py
```diff
@@ -736,7 +738,11 @@ def from_debian_base(
         return base_image
 
     @classmethod
-    def from_base(cls, image_uri: str) -> Image:
+    def from_base(
+        cls,
+        image_uri: str,
+        platform: Union[Architecture, Tuple[Architecture, ...], None] = None,
```
pingsutw: Could we just set the platform in `clone()`?

examples/plugins/spark_dataframe_example.py
```diff
@@ -9,7 +9,7 @@
 from flyte.io import File
 
 image = (
-    flyte.Image.from_base("apache/spark-py:v3.4.0")
+    flyte.Image.from_base("apache/spark-py:v3.4.0", platform=("linux/amd64", "linux/arm64"))
```
pingsutw: Could we move platform to clone()?

src/flyte/_internal/imagebuild/docker_builder.py
```diff
@@ -768,6 +783,13 @@ async def _build_image(self, image: Image, *, push: bool = True, dry_run: bool =
             for layer in image._layers:
                 dockerfile = await _process_layer(layer, tmp_path, dockerfile, docker_ignore_patterns)
 
+            # Only switch to the `flyte` user for images that actually created it (the
+            # `from_debian_base` path adds the create-flyte-user command layer). Doing this
+            # unconditionally breaks `from_base`/`from_dockerfile` images whose base never
+            # created that user — containerd then fails with "no users found".
```
pingsutw: nit: could we remove the comments since the function name is already self-explanatory

src/flyte/_image.py
machichima: Fixed in https://github.com/flyteorg/flyte-sdk/pull/1213/commits/7df870fc62c718669717f50476184da9c2f6ea03

examples/plugins/spark_dataframe_example.py
machichima: Fixed in https://github.com/flyteorg/flyte-sdk/pull/1213/commits/7df870fc62c718669717f50476184da9c2f6ea03

src/flyte/_internal/imagebuild/docker_builder.py
machichima: Fixed in https://github.com/flyteorg/flyte-sdk/pull/1213/commits/af475da8cc85738f997b2c38c03333a0308d9c62

pingsutw (approved):

PR #1218, opened 2026-06-17
fix(imagebuild): install local wheel by path so dev pre-release isn't dropped

pingsutw:
## Why

`with_local_v2()` (used when building a dev image from a locally-modified `flyte`) stopped picking up the local build. The image would install the local wheel and then immediately replace it with a stable release from PyPI:

```
#18 [build] Install wheel flyte ...
 - flyte==2.5.1
 + flyte==2.5.2.dev1+g512b21d95.d20260617   # local wheel installed
#19 [build] Install dependencies for flyte ...
 - flyte==2.5.2.dev1+g512b21d95.d20260617
 + flyte==2.5.1                              # ...then silently swapped back
```

### Root cause

`PythonWheelHandler.handle` emitted **two** `uv pip install` commands per `PythonWheels` layer:

1. Install the wheel only: `uv pip install --find-links /dist --no-deps --no-index --reinstall <pkg>`
2. Install dependencies: `uv pip install --find-links /dist <pkg>`

Once step 2 drops `--no-index`, uv sees the package from two sources — the local `/dist` wheel (a `.dev` **pre-release**) and PyPI (a **stable** release). Per PEP 440, uv/pip ignore pre-releases when a stable version satisfies an unconstrained requirement, and no `--pre` is passed (`PipOption.pre` defaults to `False`). So step 2 resolves `<pkg>` to the stable PyPI version, uninstalls the local dev wheel, and reinstalls the stable one. `--find-links /dist` doesn't help: the local wheel is *available* but still loses as a pre-release.

This only surfaces once a competing **stable** release exists on PyPI, which is why it "used to work" and then regressed.

## What

Install the wheel **directly by file path** instead of by package name:

```
uv pip install <args> /dist/<package_name>-*.whl
```

Pointing uv at the file removes the resolution contest for the package itself (no pre-release exclusion), while its dependencies still resolve from the index. This also collapses the two commands into one.

- `package_name` is already the underscore-normalized distribution name (`_image.py` builds it via `plugin.replace("-", "_")`), so `/dist/<package_name>-*.whl` matches this package's wheel. The trailing dash anchors it (`flyte-*` does not match `flyteplugins_*`).
- The glob is expanded by the build shell (the `RUN` is emitted in shell form).
- Multiple-version ambiguity can't arise from the normal build: `make dist` runs `clean` (`rm -rf dist/`) and `make dist-plugins` runs `clean-plugins` (`rm -f dist/flyteplugins_*.whl`) first, so `dist/` holds at most one wheel per package.

## Tests

Added two unit tests in `tests/flyte/imagebuild/test_docker_builder.py`:

- `test_python_wheel_handler_installs_local_wheel_by_path` — asserts a single `uv pip install`, that it installs `/dist/flyte-*.whl`, that none of `--no-deps/--no-index/--reinstall/--find-links` remain, and that the glob doesn't match a sibling `flyteplugins_*` wheel.
- `test_python_wheel_handler_glob_matches_normalized_plugin_name` — asserts the glob matches an underscore-normalized plugin name.

Full `test_docker_builder.py` suite passes (37 passed, 1 skipped); `ruff` clean.

### Verification

Reproduced the swap and confirmed the fix locally with the real `flyte` dev wheel:

| command | result |
|---|---|
| `uv pip install --find-links ./dist flyte` (old, fresh resolve) | installs **stable** PyPI build, drops local dev wheel |
| `uv pip install ./dist/flyte-*.whl` (new) | installs **local dev wheel**, all deps resolved from PyPI |

### DIFF
--- a/src/flyte/_internal/imagebuild/docker_builder.py
+++ b/src/flyte/_internal/imagebuild/docker_builder.py
@@ -245,32 +245,18 @@ class PythonWheelHandler:
         pip_install_args = layer.get_pip_install_args()
         secret_mounts = _get_secret_mounts_layer(layer.secret_mounts)
 
-        # First install: Install the wheel without dependencies using --no-deps
-        pip_install_args_no_deps = [
-            *pip_install_args,
-            *[
-                "--find-links",
-                "/dist",
-                "--no-deps",
-                "--no-index",
-                "--reinstall",
-                layer.package_name,
-            ],
-        ]
-
-        delta1 = UV_WHEEL_INSTALL_COMMAND_TEMPLATE.substitute(
-            PIP_INSTALL_ARGS=" ".join(pip_install_args_no_deps), SECRET_MOUNT=secret_mounts
+        # Install the local wheel directly by its file path (the glob is expanded by the build
+        # shell). Pointing uv at the file means there is no resolution contest for the package
+        # itself, so a locally-built pre-release wheel (e.g. ``2.5.2.dev1+g...``) is never swapped
+        # for a stable release on PyPI by PEP 440 pre-release exclusion -- which is what happened
+        # when the package name was passed unpinned. The wheel's dependencies still resolve from
+        # the index. ``package_name`` is already the underscore-normalized distribution name, so
+        # ``/dist/<package_name>-*.whl`` matches only this package's wheel (the trailing dash keeps
+        # ``flyte-*`` from matching ``flyteplugins_*``).
+        install_args = [*pip_install_args, f"/dist/{layer.package_name}-*.whl"]
+        dockerfile += UV_WHEEL_INSTALL_COMMAND_TEMPLATE.substitute(
+            PIP_INSTALL_ARGS=" ".join(install_args), SECRET_MOUNT=secret_mounts
         )
-        dockerfile += delta1
-
-        # Second install: Install dependencies from PyPI. Keep /dist as a findlink so the package
-        # itself resolves to the local wheel even when it isn't published to PyPI (e.g. a local
-        # plugin wheel); its dependencies still come from the index.
-        pip_install_args_deps = [*pip_install_args, "--find-links", "/dist", layer.package_name]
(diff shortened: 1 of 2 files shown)


pingsutw:
Revised the approach. The first version installed the wheel by direct file path (`/dist/<pkg>-*.whl`), which works for pure-Python packages (flyte + plugins, one `py3-none-any` wheel) but **breaks for `flyte_controller_base`** (the Rust controller). `rs_controller`'s `build-wheels` target produces per-arch wheels (`build-wheels-arm64` + `build-wheels-amd64` + local) into one dist, so the glob expands to several wheels of the same version and uv rejects them:

```
error: Requirements contain conflicting URLs for package `flyte_controller_base`:
- .../flyte_controller_base-<v>-...-manylinux_2_17_aarch64.whl
- .../flyte_controller_base-<v>-...-manylinux_2_17_x86_64.whl
```

New approach keeps `--find-links /dist` (so uv selects the wheel matching the **build platform**) and pins the **exact version** read from the wheel filename:

```
uv pip install <args> --find-links /dist <package_name>==<version>
```

This covers all four requirements in one command:

| requirement | mechanism |
|---|---|
| pick platform-compatible wheel (multi-arch Rust controller) | `--find-links /dist` |
| local `.dev` pre-release is eligible | exact `==` opts into pre-release (PEP 440) |
| local build wins over a stable PyPI release | exact `==<local version>` pins to it |
| dependencies from PyPI | index enabled |

Verified locally that `--find-links` over a dir containing **both** arch wheels + an exact `==<version>` pin selects the compatible wheel and resolves deps (no conflict), and that the pin keeps the local `.dev` wheel instead of a stable PyPI release.

Tests updated: pin assertion, a multi-arch case (`flyte_controller_base` with arm64+amd64 wheels → single `==<version>`), and a missing-wheel error case. Full `test_docker_builder.py` passes (38 passed, 1 skipped); `ruff` clean.

PR #1219, opened 2026-06-17
fix(image): always install the local wheel for with_local_v2()

pingsutw:
## Why

When building an image with `Image.from_*().with_local_v2()`, the locally-built `flyte` wheel can be **silently replaced by the published PyPI release**. A user modifying `flyte` locally ends up running the released version in the image without any error.

`PythonWheelHandler.handle` emits two `uv pip install` steps for a `PythonWheels` layer:

1. **force-install** the local wheel: `--no-deps --no-index --reinstall flyte`
2. **resolve dependencies** from the index: `--find-links /dist flyte`

The problem is the order: the dependency step runs **last** and does a full resolve of `flyte`. If the local wheel can't be fully satisfied during that resolve — most commonly when a dev wheel pins a dependency version that isn't published yet (e.g. after rebasing onto a branch that bumped an internal dep) — uv backtracks and installs the highest **satisfiable** version, i.e. the stable PyPI release. The result silently undoes step 1:

```
#19  - flyte==2.5.2.dev1+g512b21d95.d20260617
#19  + flyte==2.5.1
```

Note this is **not** specific to the `--find-links` that was recently added to step 2 — old and new code backtrack identically. The fix is structural.

## What

Swap the order so the **dependency step runs first** and the **force-install step runs last**. Because the force-install uses `--no-index --reinstall`, it can't be re-resolved away, so the exact local wheel is always the final state regardless of what the resolver picks for the package itself.

- Dependencies still come from the index (resolved in the first step).
- Packages not published to PyPI (local plugin wheels) still resolve via `--find-links /dist` in the first step.
- `with_local_v2()` now always installs exactly the local wheel.

Trade-off: the local wheel is force-installed with `--no-deps`, so a genuinely-missing dependency now surfaces at runtime (ImportError) instead of being masked by silently running the older published package. This is the intended behavior for `with_local_v2()` — you get your local code.

## Repro / verification

With a local dev wheel whose declared dependency can't be satisfied, the pre-fix sequence downgrades to the PyPI release; the reordered sequence keeps the local wheel:

```
[A] uv pip install --find-links /dist flyte                                  -> picks stable + deps
[B] uv pip install --no-deps --no-index --reinstall --find-links /dist flyte -> local dev wheel (final)
```

## Tests

Added `test_python_wheel_handler_forces_local_wheel_last` asserting the force-install step (`--no-index --reinstall`) is emitted **after** the dependency step in the generated Dockerfile.

### DIFF
--- a/src/flyte/_internal/imagebuild/docker_builder.py
+++ b/src/flyte/_internal/imagebuild/docker_builder.py
@@ -245,7 +245,22 @@ class PythonWheelHandler:
         pip_install_args = layer.get_pip_install_args()
         secret_mounts = _get_secret_mounts_layer(layer.secret_mounts)
 
-        # First install: Install the wheel without dependencies using --no-deps
+        # First install: resolve and install the package's dependencies from the index. Keep /dist as
+        # a findlink so the package itself still resolves even when it isn't published to PyPI (e.g. a
+        # local plugin wheel); its dependencies come from the index. The exact local wheel is forced in
+        # the second step, so it does not matter which version of the package this step picks.
+        pip_install_args_deps = [*pip_install_args, "--find-links", "/dist", layer.package_name]
+        delta1 = UV_WHEEL_INSTALL_COMMAND_TEMPLATE.substitute(
+            PIP_INSTALL_ARGS=" ".join(pip_install_args_deps), SECRET_MOUNT=secret_mounts
+        )
+        dockerfile += delta1
+
+        # Second install (last): force the exact local wheel on top of whatever the dependency step
+        # installed. --no-index + --reinstall guarantees the local wheel wins and nothing re-resolves
+        # it to a published release afterwards. This must run after the dependency step: a full resolve
+        # can otherwise discard the local wheel in favor of a stable PyPI release -- e.g. when one of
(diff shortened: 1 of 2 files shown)


src/flyte/_internal/imagebuild/docker_builder.py
```diff
@@ -245,7 +245,22 @@ async def handle(layer: PythonWheels, context_path: Path, dockerfile: str) -> st
...
+        # a findlink so the package itself still resolves even when it isn't published to PyPI (e.g. a
+        # local plugin wheel); its dependencies come from the index. The exact local wheel is forced in
+        # the second step, so it does not matter which version of the package this step picks.
+        pip_install_args_deps = [*pip_install_args, "--find-links", "/dist", layer.package_name]
+        delta1 = UV_WHEEL_INSTALL_COMMAND_TEMPLATE.substitute(
+            PIP_INSTALL_ARGS=" ".join(pip_install_args_deps), SECRET_MOUNT=secret_mounts
+        )
+        dockerfile += delta1
+
+        # Second install (last): force the exact local wheel on top of whatever the dependency step
```
pingsutw, line 258: I just move it down so it make sure we always install flyte from the wheel

kumare3 (approved):

PR #1221, opened 2026-06-17
agents plugin

samhita-alla:
Run agents written with existing agent SDKs: OpenAI Agents SDK, Claude Agent SDK, Mistral Agents, and Google ADK, on Flyte as the durable orchestration runtime underneath. You keep writing agents in each framework's own idioms; Flyte provides replay, automatic retries / self-healing, per-tool containerized execution (CPU/GPU,
caching), cross-run memory and observability. No agent loop is reimplemented.

## Packages

Five co-located packages on a shared core:

| Package | Underlying SDK |
| --- | --- |
| `flyteplugins-agents-core` | shared contract (no SDK) |
| `flyteplugins-agents-openai` | OpenAI Agents SDK |
| `flyteplugins-agents-claude` | Claude Agent SDK |
| `flyteplugins-agents-mistral` | Mistral Agents (`mistralai` 2.x) |
| `flyteplugins-agents-google` | Google ADK |

## The shared model

Every adapter exposes the same surface: `function_tool` + `run_agent` and follows the same division of labor, so the call shape is identical across SDKs:

- The agent run is a Flyte `@env.task`: the durable parent (`retries=` = self-healing, `report=True` = the agent timeline in the report).
- Each tool is a Flyte task, invoked as a durable child action (its own container/resources, retries, caching) via `function_tool`.
- Each model turn is recorded for replay by tracing the seam below the SDK's own loop (`durable=True`), so a crashed or retried run replays completed turns instead of re-calling (and re-billing) the model. Where the SDK runs its loop in a subprocess (Claude), durability is the SDK's own session-resume instead.
- Cross-run memory via `memory_key`, backed by a durable keyed `MemoryStore`.

The rule: if the SDK owns the loop, we run its loop and trace the seam beneath it. Agent maps to a task, each tool maps to a task, each model turn maps to a trace.

## What you get

- Crash-and-resume: a crash mid-run replays completed model turns and cache-hits the tool calls on retry with no rework, no re-billing.
- Tools as first-class Flyte actions: each tool runs in its own container with its own resources and caching, and fans out in parallel.
- Observability: a per-run timeline (assistant turns, tool calls/results, token-usage summary) rendered into the task report.
- Cross-run memory: continue a conversation across separate runs and workers with a shared `memory_key`.
- Human-in-the-loop: gate a tool on `flyte.new_condition` -- a durable suspend that survives restarts, for approvals or for a customer sharing details mid-conversation.
- Native multi-agent: each SDK's own handoff / sub-agent pattern (OpenAI handoffs, Claude subagents, ADK transfer, Mistral handoffs) runs durably on Flyte.

*Attempt 1*
Screenshot 2026-06-29 at 10 01 59 PM

*Attempt 2 (after the crash)*
Screenshot 2026-06-29 at 10 02 31 PM


plugins/agents/core/src/flyteplugins/agents/core/_tools.py
```diff
@@ -0,0 +1,141 @@
...
+    except Exception:  # pragma: no cover - unresolved annotations; pass through
+        return kwargs
+    coerced = dict(kwargs)
+    for name, value in kwargs.items():
+        if hints.get(name) is float and isinstance(value, int) and not isinstance(value, bool):
+            coerced[name] = float(value)
+    return coerced
+
+
+def function_tool(
```
cosmicBboy: can we rename this to just `@tool`? So we make the naming consistent with the core library https://github.com/flyteorg/flyte-sdk/blob/main/src/flyte/ai/agents/_tools.py#L449, also conceptually it could support Callable classes, which would define the `__call__` method.

plugins/agents/core/src/flyteplugins/agents/core/_memory.py
```diff
@@ -0,0 +1,54 @@
...
+
+import typing
+
+from flyte._logging import logger
+
+if typing.TYPE_CHECKING:
+    from flyte.ai.agents.memory import MemoryStore
+
+
+async def resolve_memory(memory_key: str | None, *, audit: bool = True) -> "MemoryStore | None":
```
cosmicBboy, line 32: I wonder if this should live in `flyte.ai.agents`?

plugins/agents/core/src/flyteplugins/agents/core/_durable.py
```diff
@@ -0,0 +1,54 @@
...
+from __future__ import annotations
+
+import typing
+
+import flyte
+
+T = typing.TypeVar("T")
+
+
+async def durable_step(
```
cosmicBboy, line 21: would this be helpful in `flyte.ai.agents.Agent`?

plugins/agents/core/src/flyteplugins/agents/core/_observability.py
```diff
@@ -0,0 +1,105 @@
+"""Shared report timeline — render agent events into the Flyte task report.
```
cosmicBboy, line 1: would this module be helpful in the core `flyte.ai.agents`?

plugins/agents/core/src/flyteplugins/agents/core/_memory.py
samhita-alla: we already have `MemoryStore.get_or_create` in the core. this is more to ensure memory never breaks an agent run, so i'm not sure it'd be very useful to have it in the core. wdyt?

plugins/agents/core/src/flyteplugins/agents/core/_durable.py
samhita-alla: this is a trivial helper for the agent and i'm not sure it makes sense to move it into the core. it's simple enough to define manually and is not agent-specific. wdyt?

cosmicBboy (approved):

PR #1224, opened 2026-06-17
feat: populate code_bundle_uri in Task and App serialization

EngHabu:

The UI uses `code_bundle_uri` to display the source code bundle for a task or app, and `source_code.link` to provide a GitHub link to the defining code. These fields existed in the IDL but were never populated during serialization.

1. **`task_serde.py`**: Set `code_bundle_uri` in `TaskMetadata` to `serialize_context.code_bundle.tgz or .pkl` during task serialization.

2. **`app_serde.py`**: 
   - Set `code_bundle_uri` on `App.Meta` during app serialization (tgz or pkl bundle URI).
   - Set `source_code` on `App.Meta` with a GitHub tree URL (`<remote>/tree/<sha>`) constructed from `flyte.git.GitStatus`. Gracefully returns `None` when not in a git repo (`is_valid=False`).

## How was this patch tested?

The `code_bundle_uri` and `source_code` fields on `App.Meta` are added in companion PRs:

- https://github.com/flyteorg/flyte/pull/7549 — IDL adds `code_bundle_uri` and `source_code` to `App.Meta`
- https://github.com/unionai/cloud/pull/16585 — dataproxy reads `meta.code_bundle_uri` from App, falls back to container command args and pod primary container args for PodTemplate apps

### DIFF
--- a/src/flyte/_internal/runtime/task_serde.py
+++ b/src/flyte/_internal/runtime/task_serde.py
@@ -265,6 +265,9 @@ def get_proto_task(
             is_entrypoint=task.entrypoint,
             log_links=log_links,
             image_build_run=image_build_run,
+            code_bundle_uri=serialize_context.code_bundle.tgz or serialize_context.code_bundle.pkl
+            if serialize_context.code_bundle
+            else None,
         ),
         interface=transform_native_to_typed_interface(task.native_interface),
         custom=custom if len(custom) > 0 else None,
--- a/src/flyte/_internal/runtime/task_serde.py
+++ b/src/flyte/_internal/runtime/task_serde.py
@@ -265,9 +265,7 @@ def get_proto_task(
             is_entrypoint=task.entrypoint,
             log_links=log_links,
             image_build_run=image_build_run,
-            code_bundle_uri=serialize_context.code_bundle.tgz or serialize_context.code_bundle.pkl
-            if serialize_context.code_bundle
-            else None,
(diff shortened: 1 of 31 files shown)


ursucarina (dismissed):

src/flyte/_internal/runtime/task_serde.py
```diff
@@ -265,6 +265,9 @@ def get_proto_task(
             is_entrypoint=task.entrypoint,
             log_links=log_links,
             image_build_run=image_build_run,
+            code_bundle_uri=serialize_context.code_bundle.tgz or serialize_context.code_bundle.pkl
```
kumare3: was this no already there? 
btw, .pkl cannot be downloaded

EngHabu: Good catches on both:

1. `TaskMetadata.code_bundle_uri` was **not** set before — the task path only passed the bundle through container command args (`pyflyte-execute --tgz ...`), which the dataproxy parses. This populates the dedicated field so the UI doesn't have to scrape args (and so PodTemplate apps, which have no such args, work too).

2. Fixed — dropped the `.pkl` fallback. `code_bundle_uri` now only points at the downloadable `.tgz` archive. Done in both `task_serde.py` and `app_serde.py` (e919a9e).

kumare3 (dismissed):

kumare3 (approved):

PR #1229, opened 2026-06-19
feat(shell): stage shell-task File inputs under NAMED_DIR

k1sauce:
Follow-up to #1220 (which added the file_input_layout knob to ContainerTask but did not wire it into shell tasks). Shell tasks reference inputs through a glob (/var/inputs/<name>/*), so a single File must be staged inside a per-input directory under its original basename — otherwise CoPilot stages it extensionless and tools that sniff format by extension reject it (e.g. salmon on *.fastq.gz staged as "0").

- shell/_runtime.py: _ShellContainerTask now passes file_input_layout="NAMED_DIR" to ContainerTask, so shell tasks always stage File/list[File] into per-input dirs under their original basenames (the renderer already globs them).
- _container.py: local docker staging is now layout-aware so `--local` matches remote CoPilot — NAMED_DIR keeps original basenames (index-prefix dedup on collision), DIRECT uses bare indices (0, 1, ...). Replaces the prior unconditional name-preserving behavior.
- pyproject.toml: bump flyteidl2 pin to 2.0.24 — shell tasks always request NAMED_DIR, and the emit raises on older flyteidl2 that lacks DataLoadingConfig.file_input_layout.
- tests: lock DIRECT->indices for raw container list[File]; exercise name/extension preservation and collision dedup via the public file_input_layout="NAMED_DIR" param; assert single File renders to a glob; add data_loading_config emit tests (DIRECT unset, shell emits NAMED_DIR).


samhita-alla (dismissed):

src/flyte/extras/_container.py
```diff
@@ -266,26 +271,41 @@ async def execute(self, **kwargs) -> Any:
...
+        named = self._file_input_layout == "NAMED_DIR"
+
+        def _stage_files_into_dir(items: list) -> str:
+            local_dir = storage.get_random_local_directory()
+            for i, item in enumerate(items):
+                if named:
+                    base = (item.name or os.path.basename(item.path)) or str(i)
+                    target = pathlib.Path(local_dir) / base
+                    if target.exists():
+                        target = pathlib.Path(local_dir) / f"{i}_{base}"
```
machichima: I'm thinking about the collision here. If we have following files:
1. file.txt
2. 2_file.txt (somehow user's file name is like this)
3. file.txt -> already exists, change to 2_file.txt (collision again, but we do not handle it)

Then 3. will produce duplicate target

samhita-alla: you're right. would you mind fixing this, please?

samhita-alla (dismissed):

machichima (approved):

PR #1239, opened 2026-06-22
feat(connector): serve Connect (HTTP/1.1) alongside gRPC

pingsutw:
## What
Adds a **Connect (HTTP/1.1) transport** to the connector server, running **side by side with the existing gRPC server in a single process on two ports**:

- gRPC server on `--port` (default 8080, `h2c`) — unchanged, for the still-gRPC backend.
- Connect/HTTP1.1 server (uvicorn) on the new `--connect_port` (default 8090) — serves `AsyncConnectorService` + `ConnectorMetadataService` over the Connect protocol plus a plain HTTP `GET /healthz` probe.

The per-RPC logic was extracted into transport-agnostic `_do_*` handlers in `_server.py`; the gRPC servicers and the new connectRPC services in `_connect_server.py` are thin adapters over them (each with its own metrics + error mapping), so the two transports can't drift. Connector loading and the Prometheus metrics server are set up once and shared.

## Why
flyte core already moved from gRPC to buf/Connect, but the connector was still gRPC-only. This is the connector half of that migration. Because the Go backend plugin still speaks gRPC to the connector, both transports run simultaneously during the transition — zero-downtime: deploy the connector serving both, migrate the backend to the Connect port, then drop gRPC in a follow-up.

> **Note on the two-port design:** the gRPC *wire protocol* cannot be served reliably from connect-python on a single negotiated port — it requires HTTP/2 + response trailers, and connectRPC's docs flag the HTTP/2 ASGI servers (hypercorn/granian) as not passing conformance (only the very new `pyvoy` does). The Connect protocol (unary **and** the server-streaming used by `GetTaskLogs`) runs fine over HTTP/1.1 via uvicorn, so we serve Connect on its own HTTP/1.1 port and keep gRPC on grpcio.

## Changes
- `connectors/_server.py` — shared `_do_*` handlers + `_extract_metrics_meta`; gRPC servicers reduced to thin wrappers; `ConnectorService.run` gains `connect_port`.
- `connectors/_connect_server.py` *(new)* — connectRPC service impls, `ConnectError` mapping, and `build_asgi_app()` (router → both services + `/healthz`, handles the ASGI `lifespan` scope).
- `connectors/utils.py` — `_start_connector_servers` runs `_serve_grpc` + `_serve_connect` (uvicorn, signal handlers suppressed) via `asyncio.gather`; `timeout` honored for tests.
- `_bin/connect.py` — `--connect_port` option.
- `app/_connector_environment.py` — `container_args` passes `--connect_port` (internal container port; gRPC `port` stays the exposed one).
- `pyproject.toml` — add `uvicorn` to the `connector` extra (kept `grpcio*` for the transition).

## Test
- New `tests/flyte/connector/test_connect_connector_service.py` (7 tests): Connect create/get/delete/metrics/logs handlers, `FlyteConnectorNotFound` → `ConnectError(NOT_FOUND)`, `/healthz`→200, prefix routing, 404, and `lifespan`.
- Updated 6 `container_args` assertions for the new `--connect_port`.
- Local end-to-end (not in CI): started both servers, hit `/healthz`→200 and real Connect RPCs (`list_connectors`, `get_connector`, `get_task`→SUCCEEDED) over HTTP/1.1; not-found surfaced as `Code.NOT_FOUND` over the wire.
- `mypy`, `ruff check`/`format` clean; 44 connector/env + 216 app tests pass.

## Out of scope (Go-side / deploy, follow-up)
Exposing the Connect port at the k8s Service and repointing the backend plugin to it. This PR only makes the connector *serve* both. gRPC removal is a later step once the backend is migrated.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,6 +54,7 @@ connector = [
     "grpcio-health-checking",
     "httpx",
     "prometheus-client",
+    "uvicorn",
 ]
 examples-test = [
     # Dependencies for running integration tests on examples
--- a/src/flyte/_bin/connect.py
+++ b/src/flyte/_bin/connect.py
@@ -14,7 +14,14 @@ def _connect():
     default="8000",
     is_flag=False,
     type=int,
-    help="Grpc port for the connector service. Defaults to 8000",
+    help="gRPC port for the connector service. Defaults to 8000",
+)
+@click.option(
+    "--connect_port",
+    default="8090",
+    is_flag=False,
+    type=int,
+    help="Connect (HTTP/1.1) port for the connector service. Defaults to 8090",
 )
 @click.option(
     "--prometheus_port",
@@ -35,8 +42,8 @@ def _connect():
     default=None,
     is_flag=False,
     type=int,
-    help="It will wait for the specified number of seconds before shutting down grpc server. It should only be used "
-    "for testing.",
+    help="It will wait for the specified number of seconds before shutting down the connector servers. It should only "
+    "be used for testing.",
 )
 @click.option(
     "--modules",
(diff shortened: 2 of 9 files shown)


kumare3:
@pingsutw why do we need both?

PR #1240, opened 2026-06-22
fix(controllers): bound informer watch establishment to recover from half-open reconnects

pingsutw:
## What

Wire `watch_conn_timeout_sec` into `Informer.watch()` as a **per-stream establishment deadline** so a watch stream that connects but never delivers its initial snapshot can no longer wedge a parent action forever.

## Why (the incident)

A flyte2 server restart left ~13 in-flight parent actions stuck `Running` for hours. Root cause, traced end to end:

- The runtime's watch informer reconnected onto a **pooled keep-alive connection** (pyqwest/reqwest pools by default) that was **half-open** — the old server pod was SIGKILLed, peer gone, no RST.
- It wrote the `WatchForUpdates` request onto that dead connection and then **blocked forever reading the HTTP/1.1 chunked response** (the original failure was `unexpected EOF during chunk size line`).
- The transport has `read_timeout=None`, and the intended `watch_conn_timeout_sec` guard was **never wired in**, so a hung read raised no exception. The retry/backoff loop only fires on exceptions, so it never fired — the client logged `attempt 5/10` and went silent for ~3h.
- The server confirmed it never received the request (no `WatchForUpdates stream started` log), so the snapshot that would have re-synced the children's terminal states was never delivered. The 30s sleep children were long done; the parent just never observed it.

## The fix

Replace the unbounded `async for resp in watcher` with manual iteration that applies `watch_conn_timeout_sec` to each read **up to the sentinel**. A stalled reconnect now raises `TimeoutError`, the existing backoff/retry re-issues `watch_for_updates` on a fresh connection, and the server's snapshot (`ListChildActions` + sentinel) re-syncs child state.

Key correctness points:
- The deadline is **per-attempt** (a local `established` flag), **not** `self._ready` — the incident hung on a *reconnect* after the informer was already ready, so gating on the global ready flag would not have fixed it.
- The deadline is **dropped once the sentinel arrives**, so a healthy stream that idles between updates is never recycled.

### Known gap
This guards stream *establishment*. A stream that dies *silently after* the sentinel still needs server-side heartbeats on `WatchForUpdates` (none today) to detect — a separate flyte2 change. Marked with a `ponytail:`-style comment in the code.

## Tests

`tests/internal/controllers/test_informer_watch.py`:
- `test_informer_recovers_from_hung_reconnect` — reproduces the exact incident path (establish → stream dies → reconnect hangs half-open → recover). Times out against the old code; passes with the fix.
- `test_informer_does_not_recycle_idle_established_stream` — a healthy stream that idles far longer than the deadline must not reconnect.

Full controllers suite: 98 passed.

### DIFF
--- a/src/flyte/_internal/controllers/remote/_informer.py
+++ b/src/flyte/_internal/controllers/remote/_informer.py
@@ -260,9 +260,7 @@ class Informer:
                         if established:
                             resp = await watch_iter.__anext__()
                         else:
-                            resp = await asyncio.wait_for(
-                                watch_iter.__anext__(), timeout=self._watch_conn_timeout_sec
-                            )
+                            resp = await asyncio.wait_for(watch_iter.__anext__(), timeout=self._watch_conn_timeout_sec)
                     except StopAsyncIteration:
                         break
                     retries = 0
--- a/src/flyte/_internal/controllers/remote/_informer.py
+++ b/src/flyte/_internal/controllers/remote/_informer.py
@@ -253,29 +253,8 @@ class Informer:
                             parent_action_id=parent_action_id,
                         ),
                     )
-                # Bound the time to receive each message of the initial snapshot
-                # up to the sentinel. After a server restart the transport can
-                # hand back a pooled keep-alive connection that is half-open
-                # (peer gone, no RST); the request is written but the chunked
-                # response read never returns, and with read_timeout unset on the
-                # transport this would wedge the watch -- and the parent action --
-                # forever. The deadline turns that hang into a TimeoutError so the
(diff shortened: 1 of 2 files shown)


src/flyte/_internal/controllers/remote/_informer.py
```diff
@@ -253,7 +253,16 @@ async def watch(self):
                             parent_action_id=parent_action_id,
                         ),
                     )
-                async for resp in watcher:
```
kumare3, line 256: does it hang?

pingsutw: Yep, but it’s an issue with the OSS Action service, so ignore it for now.

PR #1246, opened 2026-06-24
fix: flyte run deployed-task ignores pinned version in input discovery

samhita-alla:
Fixes https://linear.app/unionai/issue/ENG26-758/flyte-run-deployed-task-ignores-pinned-version-in-input-discovery

### DIFF
--- a/src/flyte/cli/_run.py
+++ b/src/flyte/cli/_run.py
@@ -737,7 +737,10 @@ Missing required parameter(s): {", ".join(f"--{p[0]} (type: {p[1]})" for p in mi
             sync_local_sys_paths=not self.run_args.no_sync_local_sys_paths,
         )
 
-        task = flyte.remote.Task.get(self.task_name, auto_version="latest")
+        # Build the input form from the pinned version, exactly like the execution path.
+        task = flyte.remote.Task.get(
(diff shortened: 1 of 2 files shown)


src/flyte/cli/_run.py
```diff
@@ -737,7 +737,8 @@ def get_params(self, ctx: click.Context) -> List[click.Parameter]:
             sync_local_sys_paths=not self.run_args.no_sync_local_sys_paths,
         )
 
-        task = flyte.remote.Task.get(self.task_name, auto_version="latest")
+        # Build the input form from the pinned version, exactly like the execution path.
+        task = flyte.remote.Task.get(self.task_name, version=self.version, auto_version="latest")
```
cosmicBboy, line 741: does this work? I forget if `version` and `auto_version` args are mutually exclusive

samhita-alla: it works. https://github.com/flyteorg/flyte-sdk/blob/f11befb67ce6abcc7c5823beb52720c59d2826b6/src/flyte/remote/_task.py#L201-L221 - `auto_version` gets ignored whenever `version` is set.

cosmicBboy (approved):

PR #1247, opened 2026-06-24
typed output/input API on ActionDetails/Run

samhita-alla:
Fetching another action's output via ActionDetails.outputs() eagerly reconstructs the whole interface and fails fast if any one type can't be guessed; the result is also a synthetic schema-derived look-alike, and there's no first-class way to ask
for an output as a known type. This adds an API addressing those gaps:

- output_literals() / input_literals(): raw {name: Literal} straight from the data proxy, with no interface reconstruction.
- typed_outputs(types) / typed_inputs(types): re-hydrate only the requested slots into caller-supplied types via literal_map_to_kwargs(python_types=...), which skips guess_python_type. Sibling types are never touched and the result is the caller's real class with its validators/methods.
- deserializers={Type: loader}: optional hook handing the raw pre-validation payload to a caller loader instead of the default decode, for versioned-schema models that migrate historical payloads before validating.

The data-proxy fetch is factored into a cached _fetch_action_data() shared by _cache_data and the new methods. Wrappers are surfaced on RunDetails (async) and Run (syncified).

Fixes https://linear.app/unionai/issue/ENG26-759/first-class-typed-output-fetch-without-eager-interface-reconstruction, https://linear.app/unionai/issue/ENG26-760/caller-supplied-deserializer-hook-for-pydantic-schema-migration

### DIFF
--- a/src/flyte/remote/_action.py
+++ b/src/flyte/remote/_action.py
@@ -9,6 +9,7 @@ from typing import (
     Any,
     AsyncGenerator,
     AsyncIterator,
+    Callable,
     Dict,
     Iterator,
     List,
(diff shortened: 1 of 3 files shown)


cosmicBboy (approved):
nice, I've faced this issue before too

PR #1250, opened 2026-06-25
Add related_to field as a run parameter

afbrock:
This commit adds related_to to the runspec.
It is an immutable property of the run, entered at create run time via either a cli parameter or through the sdk withRunContext().

### DIFF
--- a/src/flyte/cli/_run.py
+++ b/src/flyte/cli/_run.py
@@ -289,6 +289,18 @@ class RunArguments:
             )
         },
     )
+    related_to: str | None = field(
+        default=None,
+        metadata={
+            "click.option": click.Option(
+                ["--related-to"],
+                type=str,
+                default=None,
+                help="Name of a parent run this run is derived from (provenance). "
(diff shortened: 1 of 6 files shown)


src/flyte/cli/_run.py
```diff
@@ -289,29 +289,15 @@ class RunArguments:
             )
         },
     )
-    # TODO: add a `--recover-from <run>` option (recover a fresh run from a prior run: reuse its
-    # succeeded actions) once the flyteidl2 RunSpec.recover field + backend support ship. Hidden
-    # options still surface in `flyte gen docs`, so it's omitted entirely for now.
-    rerun_from: str | None = field(
+    related_to: str | None = field(
```
kumare3, line 292: Let's not add this. It's weird. When we use --rerun-from or called flyte rerun we should
Populate it. Doing it on cli is odd

kumare3:
Superseded by #1279 + #1303, which shipped `related_to` with the fully-automatic design instead of a run parameter. Closing.

PR #1255, opened 2026-06-25
Bump connectrpc version and update usage

katrogan:
Can't properly merge until 5 days after connect 0.11.0 release (06.30.2026)

Follow up to https://github.com/flyteorg/flyte-sdk/pull/1254 which pinned to avoid breaking changes, this actually bumps the connectrpc lib to the latest and adds a cap at the minor version since the breakage occurred with a minor version release.


machichima (approved):
LGTM!

PR #1260, opened 2026-06-26
fix: return App wrapper from dryrun app deploy (FLYTE-SDK-5W)

EngHabu:
## Summary

`--dryrun` deploy of an `AppEnvironment` crashed the CLI with `AttributeError: name`.

`_deploy_app` returns a `remote.App` in the normal path but returned the raw
`app_definition_pb2.App` IDL proto in the dryrun branch. The proto has no
top-level `.name` attribute (its name is at `metadata.id.name`), so
`DeployedAppEnvironment.table_repr()` raised `AttributeError: name` at
`("name", self.deployed_app.name)` while the CLI rendered the deploy table.

## Fix

Wrap the dryrun proto in `App(app_idl)` so both branches return an `App` and
the `.name`/`.revision`/`.desired_state`/... accessors work uniformly (matches
the declared `-> App` return type). Adds a regression test that exercises the
dryrun path and asserts `table_repr()` no longer raises.

## Sentry

fixes FLYTE-SDK-5W

### DIFF
--- a/src/flyte/app/_deploy.py
+++ b/src/flyte/app/_deploy.py
@@ -108,8 +108,10 @@ async def _deploy_app(
             app_idl.spec.container.image = serialization_context.image_cache.image_lookup[app.name]
 
         if dryrun:
-            # Dryrun intentionally short-circuits with the translated IDL proto.
-            return typing.cast("App", app_idl)
+            # Wrap the raw IDL proto in an App so callers (e.g. table_repr) can use the same
+            # attribute accessors (.name, .revision, ...) as the non-dryrun path, which returns
+            # an App. Returning the bare proto here caused `AttributeError: name` in table_repr.
+            return App(app_idl)
         ensure_client()
         resolved_image = app_idl.spec.container.image if app_idl.spec.HasField("container") else image_uri_for_log
         msg = f"Deploying app {app.name}, with image {resolved_image} version {serialization_context.version}"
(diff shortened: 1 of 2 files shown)


EngHabu:
Superseded by #1580, which landed on `main` on 2026-09-11.

#1580 makes the identical production change this PR proposed — replacing `return typing.cast("App", app_idl)` with `return App(app_idl)` in `_deploy_app`'s dryrun short-circuit — and carries an equivalent regression test (`test_deploy_app_dryrun_returns_app_facade`) alongside the one here. It also goes further, fixing the entity-table rendering that mixed task and app rows under one set of headers.

That overlap is why this branch went `CONFLICTING`. Closing rather than rebasing: there is nothing left in it that #1580 did not land.

**FLYTE-SDK-5W** (39 events, last fired 2026-09-11 at release 2.7.1) is fixed on `main` by #1580 and I have marked it *resolved in next release* (2.7.2) in Sentry.

Verified the stack before resolving: 5W is `cli/_deploy.py:invoke → <listcomp> → _deploy.table_repr → app/_deploy.table_repr`, i.e. the dryrun path. Two other issues opened 2026-09-11 share the exact same `AttributeError: name` title but are a different bug in `load_python_modules` (FLYTE-SDK-8E/8F, covered by #1473) — same message, unrelated cause.

PR #1264, opened 2026-06-26
Classify task load EIO as system error

pvditt:
## Problem

Task import/load failures happen before the normal task execution wrapper. The runtime already uploads an `error.pb` for these failures, but a raw `OSError` with `errno.EIO` was converted through the generic fallback instead of being attributed as a system failure.

That means filesystem/image read failures like `[Errno 5] Input/output error` during module import can be surfaced with the wrong attribution, even though `EIO` indicates the runtime failed to read bytes from the underlying filesystem/storage layer rather than a deterministic user-code error.

## Solution

Classify only task-load `OSError` values with `errno.EIO` as `RuntimeSystemError` before converting and uploading the load failure error document.

All other load-time exceptions, including non-EIO `OSError` values such as `FileNotFoundError`, continue through the existing generic path.

## Testing

- `uv run python -m pytest tests/flyte/internal/bin/test_runtime.py tests/flyte/internal/runtime/test_entrypoints.py`
- `uv run python -m ruff check src/flyte/_internal/runtime/entrypoints.py tests/flyte/internal/runtime/test_entrypoints.py`

### DIFF
--- a/src/flyte/_internal/runtime/entrypoints.py
+++ b/src/flyte/_internal/runtime/entrypoints.py
@@ -1,3 +1,4 @@
+import errno
 import importlib
 import os
 import traceback
@@ -137,6 +138,15 @@ def load_task(resolver: str, *resolver_args: str) -> TaskTemplate:
         raise ModuleNotFoundError(msg) from e
 
 
+def _classify_load_error(err: Exception) -> Exception:
+    if isinstance(err, OSError) and err.errno == errno.EIO:
+        return flyte.errors.RuntimeSystemError(
+            "FilesystemIOError",
+            f"Filesystem I/O error while loading task: {err}",
+        )
+    return err
+
+
 def load_pkl_task(code_bundle: CodeBundle) -> TaskTemplate:
     """
     Loads a task from a pickled code bundle.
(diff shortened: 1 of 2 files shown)


ursucarina (approved):

src/flyte/_internal/runtime/entrypoints.py
```diff
@@ -137,6 +138,15 @@ def load_task(resolver: str, *resolver_args: str) -> TaskTemplate:
         raise ModuleNotFoundError(msg) from e
 
 
+def _classify_load_error(err: Exception) -> Exception:
```
wild-endeavor, line 141: why here?

wild-endeavor (commented):
i think i'm missing context here.  cc @pingsutw
why are we handling the error here instead of in convert_from_native_to_error?

will this handling be needed in rusty.py?  can this error happen in the worker-v2 code?

src/flyte/_internal/runtime/entrypoints.py
```diff
@@ -246,14 +256,17 @@ async def load_and_run_task(
...
         # Import/load failures happen before the contextual_run wrapper below, so they must
         # also be uploaded to the error file -- otherwise the UI shows an empty message.
         logger.exception(f"Failed to load task before execution: {e}")
         if output_path:
             from .io import upload_error
 
-            error = convert_from_native_to_error(e)
+            error = convert_from_native_to_error(classified_error)
             await upload_error(error.err, output_path, recoverable=error.recoverable)
+        if classified_error is not e:
```
wild-endeavor, line 268: Isn't this the same as raise classified_error?

PR #1265, opened 2026-06-27
feat: use ruff to bundle lazy/conditional imports for loaded_modules

### LINKED ISSUE #7595: feat: use `ruff analyze graph` to detect files for the code bundle
## Why

When you `flyte.run(...)` with the default `copy_style="loaded_modules"`, the SDK decides which local files to ship in the code bundle by inspecting `sys.modules` at runtime (`list_imported_modules_as_files`). That has two downsides:

- **It can miss files.** Only modules already imported by the time bundling runs are captured. Lazy imports, conditional imports (`if TYPE_CHECKING:`, inside-function imports), or modules imported later on the cluster can be silently left out of the bundle, surfacing as a `ModuleNotFoundError` at execution time.
- **It depends on interpreter state**, so the bundle contents can vary with how/where `flyte run` is invoked.

A user asked for a more reliable, static way to determine the dependency set. [`ruff analyze graph`](https://docs.astral.sh/ruff/) produces a first-party import dependency graph statically (no need to actually import anything) and is fast. This is a *nice-to-have*, not a blocker — a good community contribution.

### Concrete example: works with `ruff analyze graph`, broken today

```python
# helper.py
def compute() -> int:
    return 42
```

```python
# main.py  (the entrypoint you `flyte.run`)
import flyte

env = flyte.TaskEnvironment(name="demo")

@env.task
async def main() -> int:
    from helper import compute  # lazy, function-level import
    return compute()
```

Run it: `flyte.run(main)`.

- **Today (`loaded_modules`):** at packaging time on your laptop, `main()` never executes, so the
  `from helper import compute` line never runs and `helper` is **not** in `sys.modules`. The bundle
  ships without `helper.py`, and the task fails on the cluster with `ModuleNotFoundError: No module
  named 'helper'`.
- **With `ruff analyze graph`:** static analysis sees the `from helper import ...` edge regardless of
  where it appears, so `helper.py` is included and the run succeeds.

The same gap applies to imports guarded by `if TYPE_CHECKING:` that are later needed at runtime, and
to any module only reachable through a code path that isn't exercised locally before packaging.

## What to change

> **Note:** the code bundling logic lives in the **[flyteorg/flyte-sdk](https://github.com/flyteorg/flyte-sdk)** repo (Python SDK), under `src/flyte/_code_bundle/`. The paths below refer to that repo. This issue is tracked here for v2 visibility.

Reimplement the existing `loaded_modules` detection on top of `ruff analyze graph`, which emits a JSON map of `file -> [files it imports]`. No new `copy_style` value or public API change — just make `loaded_modules` discover files statically from the import graph instead of from runtime `sys.modules`. Starting from the entrypoint module(s), walk that graph to collect the transitive set of first-party files, then feed them into the existing bundling path.

- `src/flyte/_code_bundle/_utils.py` — in `ls_files()`, the `copy_file_detection == "loaded_modules"` branch currently calls `list_imported_modules_as_files(str(source_path), sys_modules)`. Replace that implementation so it shells out to `ruff analyze graph`, parses the JSON, and resolves the transitive imports of the entrypoint under `source_path`. Keep filtering to files within `source_path` (drop third-party/stdlib), so the returned file list matches the existing contract.
- The `CopyFiles` literal stays as-is (`Literal["loaded_modules", "all", "none", "custom"]`) — no new value.
- Handle `ruff` not being installed: detect `shutil.which("ruff")` and fall back to the current `sys.modules` approach with a clear log. Don't hard-fail.
- `tests/` — cover a project where a module is imported lazily/conditionally and assert `loaded_modules` now includes it.

## Outcome

- [ ] `copy_style="loaded_modules"` (the default) bundles the full transitive first-party import set of the entrypoint, including lazily/conditionally imported local modules
- [ ] Third-party and stdlib files are excluded (bundle stays minimal)
- [ ] Graceful fallback to the old `sys.modules` behavior when `ruff` is not on `PATH`
- [ ] Tests cover the lazy/conditional-import case that `loaded_modules` previously missed
- [ ] No public API change (`CopyFiles` values unchanged)

## Getting started

- See the current logic: `list_imported_modules_as_files` and `ls_files` in `src/flyte/_code_bundle/_utils.py` (the `copy_file_detection == "loaded_modules"` branch).
- Try the tool: `ruff analyze graph path/to/entrypoint.py` (outputs JSON). Docs: https://docs.astral.sh/ruff/
- Entry points to read: `src/flyte/_code_bundle/bundle.py` (`build_code_bundle`), `src/flyte/_run.py` (`copy_style`).
- Test: `pytest tests/` — add a fixture project under a tmp dir with a lazy import.
- Setup: see [CONTRIBUTING](https://github.com/flyteorg/flyte-sdk/blob/main/CONTRIBUTING.md) in the flyte-sdk repo.

paullongtan:
## Issue
With the default copy_style="loaded_modules", the SDK decides which local files to ship by inspecting sys.modules at runtime (list_imported_modules_as_files). Only modules already imported by the time bundling runs are captured, so **lazily or conditionally imported local modules get silently dropped** from the bundle and surface on the cluster as ModuleNotFoundError. This affects:
- **Function-level imports**: `from helper import compute` inside a task body never runs at packaging time, so `helper` is never in sys.modules.
- **if TYPE_CHECKING: imports**: never executed at runtime, but needed when annotations are evaluated on the cluster.

Closes flyteorg/flyte#7595.

## Summary
Augments the runtime sys.modules discovery with a static first-party import graph from [ruff analyze graph](https://docs.astral.sh/ruff/). Falls back to the existing behavior whenever ruff is unavailable or the analysis fails — the feature is purely additive and never blocks bundling.

`ls_files()` now picks a discovery strategy up front:

```python
if _ruff_is_available():
    all_files = list_imported_modules_as_files_with_ruff(str(source_path), sys_modules)
else:
    all_files = list_imported_modules_as_files(str(source_path), sys_modules)
```

New functions:
- `list_imported_modules_as_files_with_ruff()` — sys.modules discovery as seeds, expanded with everything reachable through the import graph. Returns the seeds unchanged if the analysis fails.
- `_ruff_is_available()` — PATH check, logs at debug when ruff is missing.
- `_create_ruff_import_dependency_graph()` — runs `ruff analyze graph`, parses the JSON, resolves ruff's cwd-relative paths to absolute.
- `_collect_transitive_dependencies()` — BFS over the graph from the seeds.
- `_ruff_analyze_timeout_seconds()` — reads the timeout, with validation.
- Extracted the user-file filter into `_is_user_file()` / `_build_invalid_directories()`.

## Timeout
The subprocess is capped at **30s** (`_RUFF_ANALYZE_TIMEOUT_SECONDS`), overridable via **`FLYTE_RUFF_ANALYZE_TIMEOUT_SECONDS`**. Invalid or non-positive values log a warning and fall back to the default.

This is a ceiling, not a delay — `ruff analyze graph` completes in ~0.15s on flyte-sdk.

## Notes for reviewers
- **ruff is a dev dependency, not a runtime one.** Users without ruff on PATH get the existing sys.modules behavior unchanged. The feature is opt-in by environment.
- **Bundle digests now depend on the environment.** Whether ruff is installed, its version, and the user's ruff `exclude` config all influence which files are discovered, so the same source can digest differently across machines. The bundles genuinely differ in those cases, so this is correct — but it means bundle cache hit rates depend on environment uniformity across a team.
- Digest ordering was raised in review: `ls_files` already sorts before hashing, so set iteration order cannot reach the digest. That behavior predates this PR.

## Test
```
uv run pytest tests/flyte/code_bundle/test_code_bundle.py -q
```
Result: **30 passed**. Six cover this feature:
- `test_ls_files_loaded_modules_includes_function_level_import`
- `test_ls_files_loaded_modules_includes_type_checking_import`
- `test_ls_files_loaded_modules_falls_back_when_ruff_times_out`
- `test_ruff_analyze_timeout_defaults_when_env_var_unset`
- `test_ruff_analyze_timeout_env_var_override`
- `test_ruff_analyze_timeout_ignores_invalid_env_var` (parametrized)


src/flyte/_code_bundle/_utils.py
```diff
@@ -321,24 +372,84 @@ def list_imported_modules_as_files(source_path: str, modules: List[ModuleType])
...
 
-    return list(files)
+    Returns a mapping of absolute file path -> absolute paths it imports, or `None` when ruff is
+    unavailable or the analysis fails (callers fall back to the runtime `sys.modules` discovery).
+
+    Unlike the runtime snapshot, ruff statically detects imports inside function bodies and
+    `if TYPE_CHECKING:` blocks, so lazily/conditionally imported local modules are discovered even
+    when they were never executed at bundle time.
+    """
+    if shutil.which("ruff") is None:
```
kumare3, line 408: should we add a debug log here to say that we are unable to find ruff so not using it

paullongtan: Definitely. I've added a debug message when ruff is not found and falls back to the current runtime sys.modules discovery method.

paullongtan:
Hi @kumare3 and @pingsutw,
Can I get a review of this updated PR?
Thanks!

src/flyte/_code_bundle/_utils.py
```diff
@@ -125,6 +128,19 @@ def ls_files(
     if copy_file_detection == "loaded_modules":
         sys_modules = list(sys.modules.values())
         all_files = list_imported_modules_as_files(str(source_path), sys_modules)
+        # Augment with a static import graph (ruff) to include lazily/conditionally imported local modules
+        # Falls back to the runtime-only set when ruff is unavailable.
```
pingsutw: Could we add another function like `list_imported_modules_as_files_with_ruff`, and then we use this if ruff is installed?
```python
if _ruff_is_enabled:
  all_files = list_imported_modules_as_files_with_ruff(...)
else:
  all_files = list_imported_modules_as_files(...)
```

src/flyte/_code_bundle/_utils.py
```diff
@@ -321,24 +372,89 @@ def list_imported_modules_as_files(source_path: str, modules: List[ModuleType])
...
+    if shutil.which("ruff") is None:
+        logger.debug(
+            "ruff not found on PATH; skipping static import graph analysis and falling back to runtime "
+            "sys.modules discovery. Lazily/conditionally imported local modules may be omitted from the "
+            "code bundle."
+        )
+        return None
+
+    try:
+        proc = subprocess.run(
```
pingsutw, line 455: Could we add a timeout for subprocess to avoid hanging?

src/flyte/_code_bundle/_utils.py
```diff
@@ -321,24 +372,89 @@ def list_imported_modules_as_files(source_path: str, modules: List[ModuleType])
...
+def _collect_transitive_dependencies(
+    seeds: typing.Set[str],
+    graph: typing.Dict[str, List[str]],
+    source_path: str,
+    invalid_directories: List[str],
+) -> typing.Set[str]:
+    """
+    Collect every user file reachable from `seeds` by following `graph`'s import edges.
+
+    Iteratively traverses the import graph (a queue-based walk; order is irrelevant since the
```
pingsutw, line 503: Could we make sure the digest will be the same if the order could be different?
https://github.com/flyteorg/flyte-sdk/blob/80dc33e7bb840f51cdbd130bfb33624a7176d1ff/src/flyte/_code_bundle/_utils.py#L174

src/flyte/_code_bundle/_utils.py
paullongtan: Done. Added `list_imported_modules_as_files_with_ruff()` which holds the ruff + graph traversal logic, and ls_files now branches on a _ruff_is_available() check:
```python
if _ruff_is_available():
    all_files = list_imported_modules_as_files_with_ruff(str(source_path), sys_modules)
else:
    all_files = list_imported_modules_as_files(str(source_path), sys_modules)
```

src/flyte/_code_bundle/_utils.py
paullongtan: Added. 
Default 30s via `_RUFF_ANALYZE_TIMEOUT_SECONDS`, overridable with `FLYTE_RUFF_ANALYZE_TIMEOUT_SECONDS`.

A ruff analyze graph test on the flyte-sdk runs in 0.15s, so a 30s timeout should provide enough head room.

src/flyte/_code_bundle/_utils.py
paullongtan: Checked!
It's already handled: `all_files.sort()` runs before the hashing loop, so set iteration order can't reach the digest.

https://github.com/flyteorg/flyte-sdk/blob/0e64ce96c74c4877f4856211ac297963b19245e8/src/flyte/_code_bundle/_utils.py#L166

src/flyte/_code_bundle/_utils.py
```diff
@@ -24,6 +27,11 @@
 
 CopyFiles = Literal["loaded_modules", "all", "none", "custom"]
 
+# Ceiling on the `ruff analyze graph` subprocess, not a delay: the call returns as soon as ruff
+# exits, which is under a second for typical repos. Test on Flytekit shows 0.15s over 575 Python files.
```
pingsutw:
```suggestion
# exits, which is under a second for typical repos. Test on flyte-sdk shows 0.15s over 575 Python files.
```

pingsutw (commented):
one nit. LGTM, It's nice feature to add. I just tested it locally, it works well for me, thanks

src/flyte/_code_bundle/_utils.py
pingsutw: there are some linting errors. https://github.com/flyteorg/flyte-sdk/actions/runs/30502644563/job/90791342451?pr=1265

paullongtan:
Hi @pingsutw ,
Thanks for the review!
I've fixed the spelling and description in the comment. Can we get another workflow run and review?

pingsutw (approved):

PR #1266, opened 2026-06-27
Centralize EIO system error conversion

pvditt:
## Problem

PR #1264 fixed task-load `OSError(errno.EIO)` attribution in the normal `a0` entrypoint path, but the EIO classification lived in `entrypoints.py`. Review feedback pointed out that this is really exception-to-Flyte-error conversion behavior, and other runtime paths such as Rust/worker-v2 loading should not need their own copy of the same classification rule.

## Solution

Move the raw `OSError(errno.EIO)` mapping into `convert_from_native_to_error()` so any runtime path that serializes native exceptions through the shared converter gets the same result:

- `ExecutionError.kind = SYSTEM`
- `code = FilesystemIOError`
- `recoverable = True`

The entrypoint load-failure path now simply converts the original exception and re-raises it, removing the local classifier and special `raise classified_error from e` block.

## Testing

- `uv run python -m pytest tests/flyte/internal/runtime/test_convert.py tests/flyte/internal/runtime/test_entrypoints.py tests/flyte/internal/bin/test_runtime.py`
- `uv run python -m ruff check src/flyte/_internal/runtime/convert.py src/flyte/_internal/runtime/entrypoints.py tests/flyte/internal/runtime/test_convert.py tests/flyte/internal/runtime/test_entrypoints.py`

### DIFF
--- a/src/flyte/_internal/runtime/convert.py
+++ b/src/flyte/_internal/runtime/convert.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import asyncio
 import base64
 import contextvars
+import errno
 import hashlib
 import inspect
 from dataclasses import dataclass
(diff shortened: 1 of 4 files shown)


kumare3:
it should be handled in the backend right?

PR #1277, opened 2026-07-01
Add Nsight Systems GPU profiling example

kumare3:
Adds `examples/accelerators/nsys_profiling` — profile a GPU training script with Nsight Systems inside a Flyte task.

**What it does**
- Runs `nsys profile --trace=cuda,nvtx,osrt --output=lightning_nsys_trace --force-overwrite=true python train.py` on an L4 GPU.
- Returns the `.nsys-rep` as a `flyte.io.File` output (downloadable, open in the Nsight Systems GUI).
- Renders an `nsys stats` summary (kernel / NVTX / OS-runtime) into the task's HTML report deck (`report=True` + `flyte.report`).
- `train.py` ships as a real file via `include=`, not an inline string, so it's editable/runnable on its own.

**Notes**
- Base is the NGC PyTorch image (`nvcr.io/nvidia/pytorch:24.08-py3`) for a preinstalled `nsys` + torch; the module comments document the `clone(extendable=True, python_version=(3,10))` + `flyte`/`uv`/`kubernetes` layers needed to make an arbitrary base flyte-runnable.
- `PodTemplate.allow_nested_sandboxing()` supplies the `CAP_SYS_ADMIN` + AppArmor-unconfined posture that `--trace=osrt` and GPU counters require.

Validated end-to-end on `flyte run` (run succeeded; profiling ran under `/usr/bin/python3.10`, `.nsys-rep` generated and uploaded).


chrismatteson:
The subprocess method just nukes all the flyte goodness after that doesn't it? I suppose if you only do this at leaf tasks that's fine.

kumare3:
@chrismatteson because this is still python sdk, you can invoke other tasks before the subprocess, after and around it right?

chrismatteson:
I think what would make me happy with this is wrapping this task in another a0 task that maybe maps over this with a comment about putting nsys on leaf nodes only.

samhita-alla:
worked on a plugin for nys https://github.com/flyteorg/flyte-sdk/pull/1305, so closing this PR.

PR #1279, opened 2026-07-02
feat: populate RunSpec.related_to on rerun and in-container flyte.run

kumare3:
## Why

The backend added `RunSpec.related_to` (field 15, `flyteidl2.common.RunIdentifier`): an optional, set-at-creation pointer to the single parent run this run was derived from (re-run, recover, clone, or explicit provenance). This PR makes the SDK populate it automatically at CreateRun time.

## What

Two implicit provenance sources, both resolved by a new `_Runner._resolve_related_to` helper and stamped in `_apply_overrides`:

1. **Rerun** — every rerun surface (`flyte rerun`, `flyte run --rerun-from`, `flyte.rerun`, `with_runcontext().rerun()`) converges on `_Runner.rerun`; the new run points at the run being rerun. The source id comes from the fetched `action_details.pb2.id.run`, with a fallback to the `run_name` argument when the fetched id is empty.
2. **In-container `flyte.run`** — when a run is created from inside a running remote task (`TaskContext.is_in_cluster()`), the new run points at the invoking run (`flyte.ctx().action`). Local and hybrid executions are excluded; in a rerun the explicit source wins over the ambient task context.

### Gating & scope rules

- **Descriptor-gated, silent skip**: the published flyteidl2 (2.0.26) doesn't ship the field yet. Unlike `recover` — explicitly user-requested, so its absence raises — implicit provenance silently skips with a debug log, so reruns keep working on older builds.
- **Same-scope, complete-id only**: per the proto contract `related_to` is scoped to the new run's org/project/domain, and all four `RunIdentifier` fields are server-required (buf.validate `min_len=1`). The resolver stamps only when the source scope (empty fields inherited from the init config) equals the new run's target scope and the id is complete — a provenance pointer must never fail run creation.
- **Rerun base-copy path**: the field is unconditionally cleared before stamping, so a rerun-of-rerun overwrites the inherited grandparent pointer, and a cross-scope rerun (e.g. `with_runcontext(project="other").rerun(...)`) clears it rather than propagating a wrong-scope pointer.

### Tests

- New `tests/flyte/test_related_to.py`: resolver rules (ctx detection, mode gating, explicit-source precedence, scope mismatch, empty-org, cfg-scope leniency) — runs on any flyteidl2 build.
- `test_run_runspec_chars.py` / `test_rerun.py`: wire-level tests gated on `RunSpec.DESCRIPTOR.fields_by_name` — they skip on 2.0.26 and were verified to pass against the generated protos that ship the field (stamping, overwrite, clear, name fallback). Two inverse-gated tests pin the silent no-op on old builds and flag when the gate becomes obsolete.
- `_init_for_testing` gains an `org=` parameter (stamping requires a non-empty org); existing tests init without org and are unaffected — the RunSpec snapshot oracle stays byte-for-byte identical.

### Notes for reviewers

- **Backend caveat**: if the server stamps `related_to` itself, a rerun performed with an *old* SDK fetches the base spec carrying field 15 as an unknown field, which `CopyFrom` preserves and `ClearField` cannot touch — the backend should treat client-sent `related_to` at create time as advisory.
- Follow-ups deliberately out of scope: explicit `with_runcontext(related_to=...)`, provenance for trigger-created runs.

### DIFF
--- a/src/flyte/_initialize.py
+++ b/src/flyte/_initialize.py
@@ -858,6 +858,7 @@ async def _init_for_testing(
     root_dir: Path | None = None,
     log_level: int | None = None,
     client: ClientSet | None = None,
+    org: str | None = None,
 ):
     global _init_config  # noqa: PLW0603
 
@@ -871,6 +872,7 @@ async def _init_for_testing(
             project=project,
             domain=domain,
             client=client,
+            org=org,
         )
 
 
--- a/plugins/anthropic/uv.lock
+++ b/plugins/anthropic/uv.lock
@@ -393,7 +393,7 @@ requires-dist = [
     { name = "docstring-parser", specifier = ">=0.16" },
     { name = "fastapi", marker = "extra == 'examples-test'" },
     { name = "flyte-controller-base", marker = "extra == 'rust-controller'", directory = "../../rs_controller" },
-    { name = "flyteidl2", specifier = "==2.0.26" },
+    { name = "flyteidl2", specifier = "==2.0.27" },
     { name = "fsspec", specifier = ">=2025.3.0" },
     { name = "grpcio", marker = "extra == 'connector'", specifier = ">=1.71.0" },
     { name = "grpcio-health-checking", marker = "extra == 'connector'" },
@@ -460,7 +460,7 @@ dev = [
 
 [[package]]
 name = "flyteidl2"
-version = "2.0.26"
+version = "2.0.27"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "googleapis-common-protos" },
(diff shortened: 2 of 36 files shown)


ursucarina (dismissed):

src/flyte/_initialize.py
```diff
@@ -858,6 +858,7 @@ async def _init_for_testing(
     root_dir: Path | None = None,
     log_level: int | None = None,
     client: ClientSet | None = None,
+    org: str | None = None,
```
kumare3, line 861: we should not add org

src/flyte/_initialize.py
```diff
@@ -871,6 +872,7 @@ async def _init_for_testing(
             project=project,
             domain=domain,
             client=client,
+            org=org,
```
kumare3, line 875: no org

afbrock (approved):

EngHabu (approved):

PR #1291, opened 2026-07-13
fix: restore original std streams while cloudpickling for version derivation

### LINKED ISSUE #7660: flyte deploy: PicklingError ("Cannot pickle files that are not opened for reading: w") — version hash is cloudpickled inside rich's stderr-redirecting spinner
## Describe the bug

`flyte deploy` fails during version derivation with a `PicklingError` whenever any task's module globals reference an object that holds a reference to the real `sys.stderr` — loguru's default logger being the common case:

```
Error invoking command: Failed to compute deployment version: your task or environment
captures an unpicklable object (PicklingError: Cannot pickle files that are not opened
for reading: w). This is usually caused by closing over `sys.stdin` / `sys.stdout` /
`sys.stderr`, an open file handle, a thread, or a lock from module-level code. Move the
captured value inside the task function, or pass an explicit `version=...` to
`flyte.deploy(...)` to skip version derivation.
```

The error message is misleading: the captured value belongs to a third-party library, and the suggested remedy ("move the captured value inside the task function") is not actionable by the user.

## Root cause

Three things interact:

1. `apply()` derives the deployment version by hashing `cloudpickle.dumps(deployment_plan.envs)` (`flyte/_deploy.py:497`). Task functions are pickled **by value**, so the module globals they reference are pulled into the pickle graph — including loguru's global `logger`, whose default sink holds a direct reference to the original `sys.__stderr__` (a `mode='w'` file object).

2. cloudpickle's `_file_reduce` pickles a stream **by reference** only via an identity check (`cloudpickle.py:961-964`):

   ```python
   if obj is sys.stdout:
       return getattr, (sys, "stdout")
   if obj is sys.stderr:
       return getattr, (sys, "stderr")
   ...
   if "r" not in obj.mode and "+" not in obj.mode:
       raise pickle.PicklingError(
           "Cannot pickle files that are not opened for reading: %s" % obj.mode
       )
   ```

3. That `cloudpickle.dumps` call runs **inside a rich status spinner** — `with common.cli_status(obj.output_format, "Deploying..."):` (`flyte/cli/_deploy.py:216`). Rich's `Live` display replaces `sys.stdout` / `sys.stderr` with `rich.file_proxy.FileProxy` objects for its duration.

So at pickling time `sys.stderr` is a `FileProxy`, the `obj is sys.stderr` identity check misses, cloudpickle falls through to the mode check, and raises.

I confirmed this by wrapping cloudpickle's file reducer during a real deploy:

```
repr          : <_io.TextIOWrapper name='<stderr>' mode='w' encoding='utf-8'>
is sys.stderr : False
sys.stderr is now: <rich.file_proxy.FileProxy object at 0x7f0f67382ce0>
is sys.__stderr__ : True
```

And the reference chain out of the envs dict:

```
envs['<env>']._tasks['<env>.<task>'].func.__globals__['logger']._core.handlers[0]._sink._stream
```

Either condition alone is harmless — the envs pickle fine outside the spinner, and they pickle fine under a swapped stderr if the task globals don't hold a stream. It needs both.

## Expected behavior

`flyte deploy` succeeds regardless of whether a task module imports loguru (or any library that holds a stream reference).

## To Reproduce

```python
# tasks.py
import flyte
from loguru import logger

env = flyte.TaskEnvironment(name="demo")

@env.task
async def hello() -> str:
    logger.info("hi")
    return "hi"
```

```
flyte deploy --all --recursive .
```

Fails with the `PicklingError` above. Adding `--output-format table-simple` (which makes `cli_status` return a `nullcontext` instead of a spinner, leaving `sys.stderr` intact) makes the same deploy succeed — which isolates the spinner as the trigger.

## Suggested fix

Compute the version hash **outside** the `Live` display, or restore `sys.stdout` / `sys.stderr` around the `cloudpickle.dumps` calls in `apply()`. Any code that cloudpickles user-supplied object graphs shouldn't run while stdout/stderr are monkeypatched, since cloudpickle's stream handling is identity-based.

Secondary: the error message should not attribute the capture to the user's task/environment, since the offending object is frequently a third-party global the user cannot move.

## Workarounds (for anyone hitting this)

- `flyte --output-format table-simple deploy ...` — no spinner, `sys.stderr` untouched.
- Give loguru a callable sink that resolves `sys.stderr` at write time rather than capturing it at import:
  ```python
  logger.remove()
  logger.add(lambda m: sys.stderr.write(m), level="INFO")
  ```
- `--version=...` to skip version derivation entirely (loses content-derived versions).

## Environment

- `flyte` 2.3.1 (PyPI)
- `cloudpickle` 3.1.2
- `loguru` 0.7.3
- Python 3.11, Linux

cosmicBboy:

Fixes flyteorg/flyte#7660.

`flyte deploy` fails with `PicklingError: Cannot pickle files that are not opened for reading: w` whenever any task module's globals hold a reference to the real `sys.stderr` — loguru's default logger being the common case.

Two things interact:

1. `apply()` derives the deployment version by `cloudpickle.dumps(deployment_plan.envs)`. Task functions are pickled by value, so module globals they reference (e.g. loguru's `logger`, whose default sink holds the original stderr) are pulled into the pickle graph.
2. The CLI runs `apply()` inside a rich status spinner, which rebinds `sys.stdout`/`sys.stderr` to `FileProxy` objects for its duration. cloudpickle only pickles streams by reference via an identity check against the *current* `sys.stdout`/`sys.stderr`, so the captured real stderr misses the check, falls through to the write-mode check, and raises.

Reproduced on main with a task module holding a stream reference (loguru-style) pickled inside `console.status(...)` — fails exactly as reported; succeeds outside the spinner or with `--output-format table-simple`.

- Add `flyte._utils.original_std_streams()`, a context manager that temporarily rebinds `sys.stdout`/`sys.stderr` to `sys.__stdout__`/`sys.__stderr__`.
- Wrap the version-hash `cloudpickle.dumps` calls in `flyte/_deploy.py` (`apply()`) and `flyte/_serve.py` with it, so cloudpickle's identity-based stream handling sees the original streams even while a spinner is active.
- Reword the version-derivation error message: the unpicklable object is frequently a third-party module global the user cannot move, so it no longer attributes the capture to "your task or environment" unconditionally.

## How was this patch tested?

- New regression test `test_apply_version_derivation_under_redirected_std_streams`: an env graph holding `sys.__stderr__` (mimicking loguru's default sink) is deployed while `sys.stdout`/`sys.stderr` are swapped for proxies. It fails on main with the exact error from the issue and passes with this change.
- `tests/flyte/test_deploy.py`, `tests/flyte/test_serve.py`, `tests/utils/` all pass (81 tests).
- ruff format/check and mypy (pre-commit) pass.

### DIFF
--- a/src/flyte/_deploy.py
+++ b/src/flyte/_deploy.py
@@ -526,19 +526,27 @@ async def apply(deployment_plan: DeploymentPlan, copy_style: CopyFiles, dryrun: 
 
             import click
 
+            from ._utils import original_std_streams
+
             h = hashlib.md5()
             try:
-                h.update(cloudpickle.dumps(deployment_plan.envs))
-                h.update(code_bundle.computed_version.encode("utf-8"))
-                h.update(cloudpickle.dumps(image_cache))
+                # Pickle with the original std streams in place: a UI spinner (rich Live)
+                # may have swapped sys.stdout/sys.stderr for proxies, which breaks
+                # cloudpickle's identity-based handling of stream references held by
+                # module globals (e.g. loguru's default sink).
+                with original_std_streams():
+                    h.update(cloudpickle.dumps(deployment_plan.envs))
+                    h.update(code_bundle.computed_version.encode("utf-8"))
+                    h.update(cloudpickle.dumps(image_cache))
             except (_pickle.PicklingError, TypeError) as e:
                 raise click.ClickException(
-                    "Failed to compute deployment version: your task or environment captures an "
-                    f"unpicklable object ({type(e).__name__}: {e}). This is usually caused by closing "
-                    "over `sys.stdin` / `sys.stdout` / `sys.stderr`, an open file handle, a thread, "
-                    "or a lock from module-level code. Move the captured value inside the task "
-                    "function, or pass an explicit `version=...` to `flyte.deploy(...)` to skip "
-                    "version derivation."
+                    "Failed to compute deployment version: the deployment captures an "
+                    f"unpicklable object ({type(e).__name__}: {e}). This is usually caused by a "
+                    "reference to `sys.stdin` / `sys.stdout` / `sys.stderr`, an open file handle, "
+                    "a thread, or a lock reachable from module-level code — either in your task "
+                    "module or in a third-party library it imports. If the value is yours, move it "
+                    "inside the task function; otherwise pass an explicit `version=...` to "
+                    "`flyte.deploy(...)` to skip version derivation."
                 ) from e
             version = h.hexdigest()
 
--- a/src/flyte/_serve.py
+++ b/src/flyte/_serve.py
@@ -668,11 +668,18 @@ class _Serve:
         elif app_deployment.version:
             version = app_deployment.version
         else:
+            from ._utils import original_std_streams
+
             h = hashlib.md5()
-            h.update(cloudpickle.dumps(app_deployment.envs))
-            if code_bundle:
-                h.update(code_bundle.computed_version.encode("utf-8"))
-            h.update(cloudpickle.dumps(image_cache))
+            # Pickle with the original std streams in place: a UI spinner (rich Live)
+            # may have swapped sys.stdout/sys.stderr for proxies, which breaks
+            # cloudpickle's identity-based handling of stream references held by
+            # module globals (e.g. loguru's default sink).
+            with original_std_streams():
+                h.update(cloudpickle.dumps(app_deployment.envs))
+                if code_bundle:
+                    h.update(code_bundle.computed_version.encode("utf-8"))
+                h.update(cloudpickle.dumps(image_cache))
             version = h.hexdigest()
 
         # Create serialization context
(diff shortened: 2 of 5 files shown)


ursucarina (approved):

src/flyte/_deploy.py
```diff
@@ -526,19 +526,27 @@ async def apply(deployment_plan: DeploymentPlan, copy_style: CopyFiles, dryrun:
...
             import click
 
+            from ._utils import original_std_streams
+
             h = hashlib.md5()
             try:
-                h.update(cloudpickle.dumps(deployment_plan.envs))
-                h.update(code_bundle.computed_version.encode("utf-8"))
-                h.update(cloudpickle.dumps(image_cache))
+                # Pickle with the original std streams in place: a UI spinner (rich Live)
```
wild-endeavor, line 533: can we make these ai comments less distracting?  can we re-write, and place in function definition instead of call-site?

PR #1292, opened 2026-07-13
fix(auth): store tokens in one keychain item to avoid double password prompt

pingsutw:
## Why

On macOS, running a workflow prompted for the keychain password **twice**. macOS prompts once per keychain item accessed, and `KeyringStore` stored `access_token` and `refresh_token` as two separate keychain items — so every credential retrieve triggered two prompts.

## What changed

- `KeyringStore.store()` now writes both tokens as a single JSON blob under one `tokens` key → `retrieve()` does exactly one keychain read, so at most one password prompt (zero after "Always Allow").
- `retrieve()` parses the JSON blob; unparseable/missing data falls back to `None` (same as no cached credentials).
- `delete()` removes the new `tokens` key and also cleans up the legacy per-token `access_token`/`refresh_token` items.
- Public API unchanged; callers in `_authenticators/base.py` are untouched.

Note: previously cached tokens (old two-item format) are intentionally not read back — a legacy fallback would require a second keychain read, which is the exact bug. Users get one fresh browser login after upgrading, then a single prompt thereafter.

## Tests

Updated `tests/flyte/keyring/test_keyring_store.py`:
- single `set_password` call with JSON payload on store
- exactly one `get_password` call on retrieve
- unparseable blob returns `None`
- delete removes `tokens` + legacy keys

`uv run pytest tests/flyte/keyring/ -q` → all passing.

## Summary by Bito

Fixes double macOS keychain password prompts by consolidating access and refresh tokens into a single keychain item, ensuring credential retrieval performs exactly one keychain read.

### DIFF
--- a/src/flyte/remote/_client/auth/_keyring.py
+++ b/src/flyte/remote/_client/auth/_keyring.py
@@ -58,7 +58,8 @@ class KeyringStore:
     # Both tokens live in ONE keychain item so macOS only prompts for the
     # keychain password once per retrieve (one prompt per item otherwise).
     _tokens_key = "tokens"
-    # Legacy per-token keys, only used for cleanup in delete().
+    # JSON field names inside the tokens item; also the legacy per-token
+    # keychain keys cleaned up in delete().
     _access_token_key = "access_token"
     _refresh_token_key = "refresh_token"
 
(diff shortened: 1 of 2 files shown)


wild-endeavor (approved):
i think this is good.  would also like to know why click 'always allow' never seems to always allow.

PR #1298, opened 2026-07-14
build: type-check all examples with mypy and ty; add ty make target + pre-commit hook

cosmicBboy:

`make mypy` only checked `src/` plus two hand-picked example files (`examples/basics/hello.py`, `hello_v2.py`), so type regressions in the other ~500 example scripts — the SDK's primary user-facing documentation — went unnoticed, and several SDK typing gaps forced users into unannotatable code. This PR makes the entire tree (`src/` + `examples/`) pass **both mypy and ty**, and enforces it on every commit.

Baseline before this change: **548 mypy errors** and **1059 ty diagnostics** across the tree.

### Tooling
- `make mypy` now checks `src/ examples/`; new `make ty` target checks the same scope; both run as prek pre-commit hooks.
- `[tool.mypy]`: `explicit_package_bases` + `mypy_path = ["src"]` so example scripts with recurring filenames (`main.py`, `app.py`, ...) resolve to unique module names in one invocation.
- `[tool.ty.rules]`: `unresolved-import` ignored (same policy as mypy's `ignore_missing_imports` — examples import optional heavy deps like torch/sklearn); unused-ignore-comment rules disabled in both directions since the tree is checked by two checkers; fixed two stale rule names (`possibly-unbound-*` → `possibly-missing-*`) that ty had renamed and were **silently not applying**.
- ty added to the dev dependency group, with an `exclude-newer-package` override (ty releases outpace the repo's 5-day recency cutoff).

### SDK — typing-only, no runtime or user-facing API changes
- `File`/`Dir` generic `T` now defaults to `Any` (PEP 696): `File(path=...)` and `File.new_remote()` type-check without an annotation on the assignment target (~150 errors).
- `File.open()` gains mode-based overloads (`"rb"` → readable handle, `"wb"` → writable). The overloads live on a plain method returning `AbstractAsyncContextManager` delegating to the `@asynccontextmanager` generator, because neither mypy nor ty matches overloads through that decorator (~170 errors across both checkers).
- `AppEnvironment.on_startup`/`server`/`on_shutdown` accept sync **or** async callables — matching their documented behavior and the `iscoroutinefunction` dispatch in `_serve.py` — and preserve the decorated function's type.
- `AsyncConnector` is now `Generic[M: ResourceMeta]` with a default, so subclasses can parameterize (`AsyncConnector[MyJobMetadata]`) and narrow `create`/`get`/`delete`/`get_logs` honestly. `metadata_type` corrected to `type[M]` (every real connector assigns a class; internal code calls `.decode` on it), and `get_logs` is typed to permit the async-generator form the connector server already documents and handles. Unparameterized subclasses type-check exactly as before.
- ~280 further ty diagnostics fixed across 84 src files (casts, narrowed annotations, TYPE_CHECKING imports). 28 justified `# ty: ignore[rule]` comments remain, each marking a documented checker limitation.

### Examples (~470 fixes across ~180 files)
- Recurring honest patterns: `flyte.ctx()` None-guards, coroutine-list annotations, `BaseException` filtering after `gather(return_exceptions=True)`.
- `Runner.run` is now honestly typed `-> Run` in every mode: hybrid mode (test-only) wraps its natively computed outputs in the same inline Run wrapper local mode uses, instead of returning raw task outputs. This removed the `R | Run` union that would otherwise have required `assert isinstance(run, Run)` narrowing at ~75 example call sites.
- Renamed `examples/plugins/polars` → `polars_examples`: the directory shadowed the real polars package in ty's module resolution, producing 72 false diagnostics.
- Only 4 `# ty: ignore` comments across all examples.

### Real bugs the checkers caught (fixed here)
- Missing `@syncify` on `CodeModeAgent.run` — `agent.run.aio(...)` call sites would `AttributeError` at runtime.
- `typing.NotRequired` import that crashes on Python 3.10 (project floor).
- A stray `flyte.g` statement (AttributeError when run).
- A custom image builder example returning `str` where the `ImageBuilder` protocol requires `ImageBuild`.
- Shadowed FastAPI `app` variables (serve handles overwriting the ASGI app), deprecated `datetime.utcnow()`, and a `gpu_memory_utilization` kwarg silently dropped by pydantic.

### Known follow-ups (deliberately not fixed — behavior-change constraint)
- `src/flyte/cli/_tui/_remote/_screens.py:767-772` reads nonexistent proto fields (`pb2.id`/`pb2.task_template`), so that TUI panel always hits its error fallback; likely intended fields noted in a comment.
- `src/flyte/storage/_remote_fs.py:148` — dead code path awaits a syncify sync wrapper.
- Protocol/stub drift in `remote/_client/_protocols.py` worked around with ~7 casts.

### API behavior changes (intentional)
- `flyte.run(...)` / `with_runcontext(...).run(...)` now returns a `Run` in **every** mode. Hybrid mode (test-only) previously returned the task's raw outputs; it now wraps them in the same inline Run wrapper local mode uses (`wait()` no-op, `outputs()` serves the values).
- `flyte.ctx()` now always returns a `TaskContext`. Outside a task it returns a **falsy null context** (`NULL_TASK_CONTEXT`) whose fields are all `None`, instead of `None` itself — so task code reads `flyte.ctx().action` without None-guards. Detect task execution with truthiness (`if flyte.ctx():`). **Compat note:** code using identity checks (`flyte.ctx() is None`) must switch to truthiness; all in-repo call sites (src, plugins, examples, tests) were converted. One such site — the log-record factory in `flyte/_logging.py` — initially slipped through and hung any await on a `DynamicBatcher` future whose `process_fn` logged an error; fixed and covered by the previously-hanging `test_error_propagation`.

## How was this patch tested?

- `make mypy` — clean (810 files)
- `make ty` — clean
- `make unit_test` — 4268 passed, 9 skipped
- bigquery / databricks / snowflake plugin suites (subclass `AsyncConnector`) — 91 passed
- `File.open` read/write round-trip verified at runtime; `ruff format`/`check` clean
- All pre-commit hooks (fmt, mypy, ty, uvlock) passed on commit


src/flyte/ai/agents/agent.py
```diff
@@ -332,7 +332,8 @@ class Agent:
 
     _registry: dict[str, AgentTool] = field(init=False, repr=False, default_factory=dict)
     _mcp_loaded: bool = field(init=False, repr=False, default=False)
-    _mcp_loader: _MCPToolLoader = field(init=False, repr=False, default=None)  # type: ignore[assignment]
+    # Placeholder default; always assigned in __post_init__.
```
pingsutw:
```suggestion
```

src/flyte/_logging.py
```diff
@@ -25,7 +25,9 @@ def _flyte_record_factory(*args: Any, **kwargs: Any) -> logging.LogRecord:
         c = _flyte_ctx()
     except Exception:
         c = None
-    if c is not None:
+    # Truthiness, not identity: outside a task ctx() returns a falsy null context
+    # whose attributes are all None.
```
pingsutw:
```suggestion
```

src/flyte/_task_environment.py
```diff
@@ -372,15 +375,17 @@ def decorator(func: F) -> AsyncFunctionTaskTemplate[P, R, F]:
...
                 task_resolver=task_resolver,
             )
             self._tasks[task_name] = tmpl
             return tmpl
 
         if _func is None:
-            return cast(Callable[[F], AsyncFunctionTaskTemplate[P, R, F]], decorator)
-        return cast(AsyncFunctionTaskTemplate[P, R, F], decorator(_func))
+            # P/R are bound by the overloads; they are unbound in the implementation body, so
+            # cast through the unparameterized generic here.
```
pingsutw:
```suggestion
```

pingsutw (commented):
Most of the changes look good to me. There are quite a few unnecessary comments that could be removed to keep the code cleaner.

examples/apps/custom_model_serving.py
```diff
@@ -159,6 +159,7 @@ def export_model_to_pt2() -> flyte.io.File:
 # ---------------------------------------------------------------------------
 
 
+# NOTE: the SDK types on_startup as accepting a sync callable, but async functions are supported at runtime.
```
cosmicBboy:
```suggestion
```

examples/apps/pickled_gradio_app.py
```diff
@@ -32,9 +32,10 @@
 )
 
 # State that can be shared between startup and server
-state = {}
+state: dict[str, str] = {}
 
 
+# NOTE: the SDK types on_startup as accepting a sync callable, but async functions are supported at runtime.
```
cosmicBboy:
```suggestion
```

examples/apps/pickled_model_serving.py
```diff
@@ -50,6 +50,7 @@ def train_model() -> flyte.io.File:
     return flyte.io.File.from_local_sync("./model.joblib")
 
 
+# NOTE: the SDK types server as accepting a sync callable, but async functions are supported at runtime.
```
cosmicBboy:
```suggestion
```

src/flyte/_internal/controllers/remote/_client.py
```diff
@@ -97,18 +97,20 @@ def state_service(self) -> StateService:
         """
         The state service.
         """
-        return self._state_service
+        # The generated connect client matches the protocol at runtime (parameter names and
+        # streaming return types differ from the hand-written protocol declarations).
```
cosmicBboy:
```suggestion
```

src/flyte/_internal/controllers/__init__.py
```diff
@@ -34,7 +34,8 @@ def next_seq(self, task_obj: object, action_key: str) -> int:
             name = task_obj.name
 
         sequencer = self._counters[action_key]
-        task_id: int | str = name or id(task_obj)
+        # `__name__`/`name` are strings in practice; the attributes are accessed dynamically.
```
cosmicBboy:
```suggestion
```

src/flyte/_internal/resolvers/internal.py
```diff
@@ -48,7 +48,10 @@ def load_task(self, loader_args: List[str]) -> TaskTemplate:
 
         return builder(**parsed)
 
-    def loader_args(self, task: TaskTemplate, root_dir: Optional[Path] = None) -> List[str]:
+    # The `Resolver` protocol names this parameter `t`, but the runtime call site
+    # (flyte._task.TaskTemplate.container_args) passes it as `task=`, so renaming here would
+    # break at runtime. The proper fix is renaming the protocol parameter in common.py.
```
cosmicBboy:
```suggestion
```

src/flyte/cli/_tui/_remote/_screens.py
```diff
@@ -763,8 +763,12 @@ def _fetch_detail_lines(self) -> list[str]:
             t = remote.Task.get(name=self._entity_name, version="latest")
             details = t.fetch()
             lines.append(f"name:       {self._entity_name}")
-            lines.append(f"version:    {details.pb2.id.version}")
-            lines.append(f"type:       {details.pb2.task_template.type}")
+            # NOTE: `pb2.id` / `pb2.task_template` do not exist on TaskDetails (ty is right);
+            # accessing them raises AttributeError which is caught by the caller's fallback.
+            # Kept as-is to preserve existing runtime behavior; likely intended:
+            # `pb2.task_id.version` and `pb2.spec.task_template.type`.
```
cosmicBboy:
```suggestion
```

src/flyte/connectors/_connector.py
```diff
@@ -88,20 +89,30 @@ class Resource:
     custom_info: Optional[typing.Dict[str, Any]] = None
 
 
-class AsyncConnector(ABC):
+# The resource-meta type a connector round-trips through create/get/delete. Defaults to
+# ResourceMeta (PEP 696) so unparameterized `class MyConnector(AsyncConnector)` keeps
+# type-checking exactly as before.
```
cosmicBboy:
```suggestion
```

src/flyte/connectors/_server.py
```diff
@@ -194,7 +197,9 @@ async def GetTaskLogs(self, request: GetTaskLogsRequest, context: grpc.ServicerC
             async for msg in result:
                 yield msg
             return
-        response = await result
+        # isasyncgen() narrowed away the generator shape at runtime, but mypy can't
+        # follow that check — the remaining shape is the single-response coroutine.
```
cosmicBboy:
```suggestion
```

src/flyte/connectors/utils.py
```diff
@@ -64,7 +64,8 @@ async def _start_grpc_server(
 
     print_metadata()
 
-    server = grpc.aio.server(futures.ThreadPoolExecutor(max_workers=worker))
+    # grpc/__init__ imports `aio` unconditionally at runtime; the stubs mark it conditional.
```
cosmicBboy:
```suggestion
```

src/flyte/io/_dataframe/basic_dfs.py
```diff
@@ -77,10 +77,10 @@ def __init__(self):
 
     async def decode(
         self,
-        proto_value: literals_pb2.StructuredDataset,
+        flyte_value: literals_pb2.StructuredDataset,
```
cosmicBboy, line 80:
```suggestion
        proto_value: literals_pb2.StructuredDataset,
```

src/flyte/io/_dataframe/basic_dfs.py
```diff
@@ -77,10 +77,10 @@ def __init__(self):
...
         current_task_metadata: literals_pb2.StructuredDatasetMetadata,
     ) -> "pd.DataFrame":
-        uri = proto_value.uri
+        uri = flyte_value.uri
```
cosmicBboy, line 83:
```suggestion
        uri = proto_value.uri
```

src/flyte/io/_dataframe/basic_dfs.py
```diff
@@ -176,7 +176,7 @@ def __init__(self):
     async def encode(
         self,
         dataframe: DataFrame,
-        dataframe_type: types_pb2.StructuredDatasetType,
+        structured_dataset_type: types_pb2.StructuredDatasetType,
```
cosmicBboy, line 179:
```suggestion
        dataframe_type: types_pb2.StructuredDatasetType,
```

src/flyte/io/_dataframe/basic_dfs.py
```diff
@@ -195,7 +195,7 @@ async def encode(
         pq.write_table(dataframe.val, strip_protocol(path), filesystem=filesystem)
         return literals_pb2.StructuredDataset(
             uri=uri,
-            metadata=literals_pb2.StructuredDatasetMetadata(structured_dataset_type=dataframe_type),
+            metadata=literals_pb2.StructuredDatasetMetadata(structured_dataset_type=structured_dataset_type),
```
cosmicBboy, line 198:
```suggestion
            metadata=literals_pb2.StructuredDatasetMetadata(structured_dataset_type=dataframe_type),
```

src/flyte/io/_dataframe/basic_dfs.py
```diff
@@ -205,12 +205,12 @@ def __init__(self):
 
     async def decode(
         self,
-        proto_value: literals_pb2.StructuredDataset,
+        flyte_value: literals_pb2.StructuredDataset,
```
cosmicBboy, line 208:
```suggestion
        proto_value: literals_pb2.StructuredDataset,
```

src/flyte/io/_dataframe/basic_dfs.py
```diff
@@ -205,12 +205,12 @@ def __init__(self):
...
         current_task_metadata: literals_pb2.StructuredDatasetMetadata,
     ) -> "pa.Table":
         import pyarrow.parquet as pq
 
-        path = proto_value.uri
+        path = flyte_value.uri
```
cosmicBboy, line 213:
```suggestion
        path = proto_value.uri
```

src/flyte/_task.py
```diff
@@ -526,6 +529,7 @@ async def execute(self, *args: P.args, **kwargs: P.kwargs) -> R:
             await self.post(v)
         return v
 
+    # Parameter is named `serialize_context` (base uses `sctx`); renaming would break keyword callers.
```
cosmicBboy:
```suggestion
```

src/flyte/remote/_client/controlplane.py
```diff
@@ -555,11 +557,13 @@ def task_service(self) -> TaskService:
 
     @property
     def app_service(self) -> AppService:
-        return self._app_service
+        # Generated connect clients satisfy the protocols at runtime; the stubs'
+        # streaming-method return annotations differ from the protocol's.
```
cosmicBboy:
```suggestion
```

src/flyte/remote/_client/controlplane.py
```diff
@@ -358,7 +358,9 @@ async def _select_and_build(self, req: cluster_payload_pb2.SelectClusterRequest)
 
         endpoint = resp.cluster_endpoint
         if not endpoint or endpoint == self._session_config.endpoint:
-            return self._default_client
+            # The generated connect client structurally satisfies the DataProxyService
+            # protocol at runtime; the stubs lag behind the protocol definition.
```
cosmicBboy:
```suggestion
```

src/flyte/io/_file.py
```diff
@@ -378,16 +381,54 @@ async def process_existing_file() -> str:
...
         return cls(path=remote_path, hash=file_cache_key)
 
-    @asynccontextmanager
-    async def open(
+    # The overloads below are typing-only: they narrow the yielded type by open mode
+    # (read modes yield a readable handle, write modes a writable one) so callers
+    # don't have to discriminate the union themselves. The overloads live on a plain
+    # method delegating to the @asynccontextmanager generator because neither mypy
+    # nor ty reliably matches overloads through that decorator. Runtime behavior is
+    # unchanged: open() still returns the async context manager the generator yields.
```
cosmicBboy:
```suggestion
```

src/flyte/io/_dir.py
```diff
@@ -271,9 +273,9 @@ def lazy_uploader(self, lazy_uploader: Callable[[], Coroutine[Any, Any, tuple[st
         self._lazy_uploader = lazy_uploader
 
     @classmethod
-    def _deserialize(cls, file_dump: Dict[str, Optional[str]]) -> Dir:
+    def _deserialize(cls, value: Dict[str, Optional[str]]) -> Dir:
```
cosmicBboy, line 276:
```suggestion
    def _deserialize(cls, file_dump: Dict[str, Optional[str]]) -> Dir:
```

src/flyte/io/_dir.py
```diff
@@ -271,9 +273,9 @@ def lazy_uploader(self, lazy_uploader: Callable[[], Coroutine[Any, Any, tuple[st
...
         """Internal: Deserialize Dir from dictionary. Not intended for direct use."""
-        return cls.model_validate(file_dump)
+        return cls.model_validate(value)
```
cosmicBboy, line 278:
```suggestion
        return cls.model_validate(file_dump)
```

examples/advanced/resource_tuner.py
```diff
@@ -46,6 +46,7 @@ async def tune_memory(udf: typing.Callable, inputs: dict) -> str:
     """
     Retry foo with more memory if it fails.
     """
+    assert isinstance(udf, flyte.TaskTemplate)  # tasks passed as inputs are TaskTemplates
```
wild-endeavor, line 49: why not just change the input signature to `flyte.TaskTemplate` instead of `typing.Callable`?

src/flyte/_bin/serve.py
```diff
@@ -281,7 +281,7 @@ async def run_user_shutdown_hooks() -> None:
             # Run the function on a separate thread, in case the sync function
             # relies on third party libraries that use an event loop internally.
             def run_sync():
-                return app_env._server(**bound_params)
+                return typing.cast("typing.Callable", app_env._server)(**bound_params)
```
wild-endeavor: can we add typing to _server?

src/flyte/__init__.py
```diff
@@ -60,6 +61,7 @@ def version() -> str:
     "HABANA_GAUDI",
     "TPU",
     "AppHandle",
+    "AsyncFunctionTaskTemplate",
```
wild-endeavor, line 64: are these really needed?  this is only for passing tasks themselves?  should we expose a new headline package?

Is it possible to only expose these for typing?

src/flyte/_context.py
```diff
@@ -155,9 +155,9 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, exc_tb):
         """Exit the context, restoring the previous context."""
         try:
-            root_context_var.reset(self._token)
+            root_context_var.reset(cast(contextvars.Token["Context"], self._token))
```
wild-endeavor: is this cast only to assert it's not None?  would an actual assert that it's not None work?

src/flyte/_utils/file_handling.py
```diff
@@ -70,7 +70,7 @@ def compute_digest_for_dir(source: os.PathLike):
                 compute_digest_for_file(Path(abspath), Path(relpath))
 
     if isinstance(source, list):
-        for src in source:
+        for src in cast("List[Union[str, os.PathLike[str]]]", source):
```
wild-endeavor, line 73: there's a bunch of places that look like ty just isn't capturing yet.  this may be fixed in the future. we should revisit in a year.

wild-endeavor (commented):
Can we make sure we tag this as a breaking change?

`flyte.ctx() is None` will now not return true, and i think there's some argument renames that people might be using.

src/flyte/_bin/serve.py
wild-endeavor: if this is an optional thing can we assert not none before it?

examples/advanced/resource_tuner.py
cosmicBboy: the types will break, since the resource tuner is trying to access attributes that don't exist in `Callable`

src/flyte/__init__.py
cosmicBboy: They're needed at runtime, so they can't be `TYPE_CHECKING`-only: `resource_tuner` uses `isinstance(udf, flyte.TaskTemplate)`, and durable_agent.py:151 uses `_tool_tasks: list[flyte.AsyncFunctionTaskTemplate] = [...]` (a module-level annotation, evaluated at import). TaskTemplate is also the legitimate public type for tasks-passed-as-values. Left as-is.

src/flyte/_bin/serve.py
cosmicBboy: updating this, good catch

src/flyte/_context.py
cosmicBboy: updating this

wild-endeavor (approved):

PR #1299, opened 2026-07-14
feat(auth): FLYTE_USE_FILE_KEYRING opt-out of macOS keychain prompts

pingsutw:
Followup to #1292 — answers https://github.com/flyteorg/flyte-sdk/pull/1292#pullrequestreview-4696509642

## Why does "Always Allow" never always allow?

Two independent macOS keychain behaviors, both verified:

1. **python-keyring's macOS backend implements `set_password` as delete + `SecItemAdd`** (`keyring/backends/macOS/api.py`). Every token refresh recreates the keychain item from scratch with a fresh ACL, so the "Always Allow" grant is wiped on the next store.
2. **Keychain ACLs identify the client app by code signature.** uv / python-build-standalone interpreters are **ad-hoc signed with no identity** (`codesign -dv` → `Identifier=-`, `Signature=adhoc`), so the keychain can't build a durable app identity — every new venv (or uv python upgrade) is a "different app" and re-prompts for the login password even though the token is already cached.

Neither is fixable while tokens live in the macOS keychain.

## What changed

`FLYTE_USE_FILE_KEYRING=1` raises the priority of the existing `SimplePlainTextKeyring` backend (already shipped for notebook environments) above the macOS keychain backend, so tokens are stored in `~/.flyte/keyring.cfg` (created `0600`) — same model as `gcloud`/`aws`/`kubectl` credentials, zero prompts.

**Default behavior is unchanged**: without the env var the system keyring is still used, and notebook detection still works as before.

## Tests

- `test_priority_forced_by_env_var`: unset/"0" → below macOS priority; "1"/"true" → above.
- Manual: `FLYTE_USE_FILE_KEYRING=1` selects `SimplePlainTextKeyring` and round-trips a token through `~/.flyte/keyring.cfg`.
- `uv run pytest tests/flyte/keyring/ -q` → 22 passed.

### DIFF
--- a/src/flyte/_keyring/file.py
+++ b/src/flyte/_keyring/file.py
@@ -1,3 +1,4 @@
+import os
 from base64 import decodebytes, encodebytes
 from configparser import ConfigParser, NoOptionError, NoSectionError
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


pingsutw:
Superseded by #1300 — tokens stay in the macOS keychain; access is routed through /usr/bin/security so every venv is the same app to the keychain and prompts disappear.

PR #1300, opened 2026-07-14
fix(auth): route macOS keychain access through /usr/bin/security to kill per-venv password prompts

pingsutw:
Followup to #1292 (and replaces #1299's env-var approach — tokens stay in the keychain).

## Problem

Creating a new python venv makes flyte prompt for the login password again even though a valid token is already cached in the keychain, and "Always Allow" never sticks. Two causes:

1. The keychain authorizes clients by **code signature**. uv / python-build-standalone interpreters are ad-hoc signed with no identity (`codesign -dv` → `Identifier=-`, `Signature=adhoc`), so every venv is a *different app* to the keychain.
2. python-keyring's native macOS backend implements `set_password` as delete + `SecItemAdd`, recreating the item with a fresh ACL on every token refresh — so even a granted "Always Allow" is wiped.

## Fix

`SecurityCliKeyring` talks to the keychain through **`/usr/bin/security`** instead of the Security-framework bindings. The CLI is Apple-signed and identical for every venv, so the keychain sees the *same app* no matter which interpreter is running — items it creates are read back with **zero prompts**, and tokens stay in the encrypted default keychain.

On Darwin, `KeyringStore` uses this class directly; on Linux/Windows the `keyring` package is used exactly as before (and the lazy-`keyring`-import cold-start invariant for cluster pods is preserved).

## Hardening from external security review

An independent security review of the first revision was addressed in d6ef828:

- **No command interpreter in the secret path** (was High): the first revision piped secrets through `security -i`, whose line parser executes embedded newlines as further commands and truncates at 4096 bytes — corrupting long tokens and potentially echoing fragments into stderr. Secrets now ride in **argv via execve**: no parsing, no quoting, no length limit (verified live with a 9 KB token payload). argv is briefly visible to same-user `ps`, which adds no exposure given the ACL model below.
- **No global keyring hijack** (was High): the first revision registered the backend in the process-global `keyring.backends` entry-point group with top priority, silently changing credential storage for *every* package using keyring in the venv. The entry point is removed; the class is instantiated only inside flyte's `KeyringStore`.
- **Honest error semantics** (was Medium): `get_password` returns `None` only for `errSecItemNotFound` (exit 44); locked-keychain / denied-access now raise `KeyringError` instead of masquerading as "no password stored". All errors use keyring's typed exceptions; all subprocess calls have timeouts.
- **Correct migration**: `set_password` recreates the item (delete + add) instead of `-U` — an in-place update preserves the old item's ACL, so legacy interpreter-owned items would have kept prompting forever.

### Documented tradeoff (deliberate, please weigh in)

Items created by `/usr/bin/security` trust that binary, so **any process running as the same macOS user can read them through the same CLI without a prompt**. That is the same exposure as the file-based credential stores of `gcloud`/`aws`/`kubectl`, with the added benefit that the keychain is encrypted at rest and locks with the session. The alternative — a dedicated Flyte-signed helper binary with its own code signature — would restore per-app ACLs and is noted as a future upgrade path if wanted.

## Tests

- `tests/flyte/keyring/` → 40 passed: argv-verbatim secret handling (newlines, quotes, 8 KiB payloads), delete-then-add recreation, not-found vs error exit codes, timeouts, platform-based backend selection, and `KeyringStore` behavior against a mocked backend (no real keychain access in CI).
- Live on macOS: full `KeyringStore.store/retrieve/delete` roundtrip including a 9 KB token — no prompts; `keyring.get_keyring()` confirmed unchanged for other packages.


cosmicBboy:
@pingsutw looks like there's a merge conflict... not sure which branch is the valid implementation

cosmicBboy (approved):

EngHabu (approved):

PR #1303, opened 2026-07-14
Update relation fields

afbrock:
Update the rerun method to use the relation type enum set to RERUN and use the new relation message instead of the old related_to field


cosmicBboy:
need to resolve conflict @afbrock

cosmicBboy (approved):

PR #1305, opened 2026-07-15
nsight systems plugin

samhita-alla:
image


cosmicBboy:
nice! minor nit: can we prettify the css on the report? probably something claude can handle pretty easily

cosmicBboy (approved):

PR #1316, opened 2026-07-20
docs: fix inverted/inaccurate reset_root_logger docstrings

ppiegaze:
## What

The `reset_root_logger` parameter docstrings describe the **opposite** of the actual behavior:

- **`_run.py`** (`with_runcontext`): *"If true, the root logger will be preserved and not modified by Flyte."* — but `reset_root_logger=True` **wipes** the root logger's handlers and installs Flyte's own (`_logging._setup_root_logger` → `root.handlers.clear()` + `root.addHandler(...)`).
- **`_initialize.py`** (`init`): *"By default, we clear out root logger handlers and set up our own."* — but the **default is `False`**, which leaves the root handlers in place (they're wrapped, not cleared).

## Fix

Rewrite both docstrings to match the code, consistently:

> If True, wipe the root logger's existing handlers and install Flyte's own handler (JSON or Rich, matching `log_format`) at the root, so logs from third-party libraries that propagate to the root logger are captured in Flyte's format. Defaults to False, which leaves the root logger's handlers in place, wrapping them so their output carries run/action context.

Docstring-only (no behavior change). These surface in the generated API reference. Verified against `_logging._setup_root_logger` on `main`.

*Surfaced by the Union.ai docs team while documenting JSON logging + `FLYTE_RESET_ROOT_LOGGER` (unionai-docs#1234 / DOC-1257) — the docs had to follow the code because the docstrings were inverted.*

### DIFF
--- a/src/flyte/_initialize.py
+++ b/src/flyte/_initialize.py
@@ -234,7 +234,10 @@ async def init(
       defaults to the editable install directory if the cwd is in a Python editable install, else just the cwd.
     :param log_level: Optional logging level for the logger, default is set using the default initialization policies
     :param log_format: Optional logging format for the logger, default is "console"
-    :param reset_root_logger: By default, we clear out root logger handlers and set up our own.
+    :param reset_root_logger: If True, wipe the root logger's existing handlers and install Flyte's own
+        handler (JSON or Rich, matching ``log_format``) at the root, so logs from third-party libraries that
+        propagate to the root logger are captured in Flyte's format. Defaults to False, which leaves the root
+        logger's handlers in place, wrapping them so their output carries run/action context.
     :param api_key: Optional API key for authentication
     :param endpoint: Optional API endpoint URL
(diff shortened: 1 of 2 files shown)


ppiegaze:
Closing and re-authoring against `main` rather than rebasing.

The correction here is **still valid** — I re-verified it against today's code before closing, not just the diff. What is no longer valid is the *form*: this branch edits Sphinx `:param` lines, and `main` has **none left**. Commit `838b41f4` ("strip inert reStructuredText from docstrings, and gate it in CI", #1376) converted the file to Google style, so there are 60-70 `:param` lines on this branch and zero on the target.

That is not a conflict to resolve, it is a rewrite — and it is faster and clearer to redo the few lines than to reconcile 130 commits across a style migration.

Replacement PR: #TBD, with the same substance in Google style.

---

ppiegaze:
Replacement opened: #1493, same substance in Google style.

---

PR #1320, opened 2026-07-21
fix: route Dir.download_sync through obstore-aware storage.get

cosmicBboy:
## Summary

Fixes [ENG26-937](https://linear.app/unionai/issue/ENG26-937/bug-reading-a-dir-input-from-a-sync-task-via-dirdownload-sync-fails).

Reading a `Dir` input from a **sync** task via `Dir.download_sync()` failed with `Object at location <dir-uri> not found`, even though the directory and its files exist in object storage. The async equivalent, `await Dir.download()`, on the same `Dir` worked.

### Root cause

`Dir.download_sync()` bypassed the SDK's obstore-aware `storage.get()` and called the raw fsspec `fs.get(self.path, local_dest, recursive=True)`. For obstore-backed filesystems (S3/GCS/ABFS), the obstore backend resolves `recursive=True` on a directory prefix as a **single-object GET** on that prefix, which 404s. The async `Dir.download()` didn't hit this because it routes through `storage.get()`, which detects obstore protocols and lists + downloads the prefix contents.

### Fix

`download_sync` now mirrors the async path, running the obstore-aware `storage.get()` synchronously via the existing `syncify` bridge:

```python
return syncify(storage.get)(self.path, local_dest, recursive=True)
```

The local-filesystem and `local_path is None` branches are unchanged.

## Tests

Added two `@pytest.mark.sandbox` regression tests in `tests/flyte/io_types/test_dirs.py`:
- `test_download_dir_sync` — downloads an S3 dir to a target path, asserting nested subdirectory contents arrive.
- `test_download_dir_sync_no_local_target` — downloads without a target path.

Both were verified to **fail with the original `FileNotFoundError`** when the fix is reverted and **pass** with it (run against a local minio S3 backend).

## Reproduction example

Added `examples/basics/dir_download_sync_repro.py`, which builds a multi-file `Dir` (with a nested subdirectory) in an async task and downloads it from a **sync** task via `download_sync()` — the exact scenario from the issue. It only manifests the bug against real object storage; the docstring notes it must be run against a remote storage config.

## Test plan

- [x] `test_download_dir_sync` / `test_download_dir_sync_no_local_target` fail without the fix, pass with it (minio S3)
- [x] Full non-integration io_types suite passes (54 tests)
- [x] Reproduction example fails without the fix, succeeds with it
- [x] ruff + mypy clean

### DIFF
--- a/examples/basics/dir_download_sync_repro.py
+++ b/examples/basics/dir_download_sync_repro.py
@@ -1,7 +1,4 @@
 """
-Reproduction for ENG26-937:
-https://linear.app/unionai/issue/ENG26-937/bug-reading-a-dir-input-from-a-sync-task-via-dirdownload-sync-fails
-
 Reading a ``Dir`` input from a *sync* task via ``Dir.download_sync()`` used to fail with::
 
     Object at location <dir-uri> not found
@@ -54,8 +51,6 @@ async def create_remote_directory() -> Dir:
 def download_directory_sync(d: Dir) -> list[str]:
     """
     The failing path: a *sync* task receives a Dir input and calls ``download_sync()``.
-
-    This is the exact scenario from ENG26-937.
     """
     local_path = d.download_sync()
     print(f"Downloaded dir sync to: {local_path}")
--- a/examples/basics/dir_download_sync_repro.py
+++ b/examples/basics/dir_download_sync_repro.py
@@ -1,23 +1,4 @@
-"""
-Reading a ``Dir`` input from a *sync* task via ``Dir.download_sync()`` used to fail with::
-
-    Object at location <dir-uri> not found
-
-even though the directory and its files exist in object storage. The async equivalent,
-``await Dir.download()``, on the same ``Dir`` works. The sync path bypassed the SDK's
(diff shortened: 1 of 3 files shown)


examples/basics/dir_download_sync_repro.py
```diff
@@ -0,0 +1,77 @@
...
+listing + downloading its contents.
+
+IMPORTANT: this bug only manifests against real object storage (S3/GCS/ABFS), i.e. an
+obstore-backed filesystem. Run it against a remote storage config, not the local filesystem::
+
+    flyte run --project <p> --domain <d> examples/basics/dir_download_sync_repro.py main
+
+Expected behavior *after* the fix: ``download_directory_sync`` downloads the directory
+(including the nested subdirectory) and prints the local files. Before the fix, it raised
+``FileNotFoundError: Object at location ... not found``.
```
pingsutw: nit: should we remove this comments and rename file name to `dir_download_sync` maybe? It's odd to show it in the example

cosmicBboy: done

pingsutw (approved):

PR #1321, opened 2026-07-22
fix: sync local sys paths for deploy

### LINKED ISSUE #7675: [BUG] [v2] `flyte deploy` never records `_F_SYS_PATH`, breaking code-bundle imports under `src/`
## Describe the bug

The [deploy command docs](https://www.union.ai/docs/v2/flyte/user-guide/task-deployment/deploy-command-options/#--no-sync-local-sys-paths) document sys.path synchronization as the **default behavior of `flyte deploy`**:

> Default behavior (path synchronization enabled): Flyte captures local sys.path entries that are under the root directory. These paths are passed to the remote container via the `_F_SYS_PATH` environment variable. At runtime, the remote container adds these paths to its sys.path, maintaining the same import environment.

This is not implemented. The deploy code path never computes or records `_F_SYS_PATH`; only the run path does.

`flyte run` and `flyte deploy` handle code-bundle import paths asymmetrically:

- **run**: the client computes `_F_SYS_PATH` from the local `sys.path` and injects it as a run-level env var — `_RunLauncher._build_env_dict` in `flyte/_run.py`:

  ```python
  # These paths will be appended to sys.path at runtime.
  if cfg.sync_local_sys_paths:
      root_dir_abs = pathlib.Path(cfg.root_dir).resolve()
      env[FLYTE_SYS_PATH] = ":".join(
          f"./{pathlib.Path(p).relative_to(root_dir_abs)}"
          for p in sys.path
          if pathlib.Path(p).is_relative_to(root_dir_abs)
      )
  ```

- **deploy**: nothing equivalent exists. `flyte/_deploy.py`, `task_serde.py`, and `trigger_serde.py` never reference `sync_local_sys_paths` or `FLYTE_SYS_PATH`; `translate_task_to_wire` serializes only the user-declared `env_vars` into `container.env`. The deployed task spec and its triggers carry no record of the sys.path entries the code bundle needs.

At runtime, `adjust_sys_path` (`flyte/_utils/module_loader.py`) puts only `.`, the code-bundle destination, and whatever `_F_SYS_PATH` contains onto `sys.path`. For a client-launched run, `_F_SYS_PATH` arrives via the run envs and imports work. For a run of a deployed task fired by a trigger/schedule, there is no client, `_F_SYS_PATH` is unset, and only the bundle root is importable.

Note `flyte deploy` even accepts `--no-sync-local-sys-paths` and stores it in the init config, but the deploy code path never reads it — the flag is a no-op, suggesting this was intended but never wired up.

## Consequence

Any project with a `src/` layout (the `uv init --lib` default and standard modern packaging) using a `with_uv_project` image is broken for scheduled runs: `project_install_mode` defaults to `dependencies_only`, so the project's own package exists only in the code bundle at `./src/<pkg>`. `flyte run` injects `_F_SYS_PATH=./src:...` and works; the identical deployed task fired by a `Cron` trigger fails every time with `ModuleNotFoundError: No module named '<pkg>'`. The workaround is to compute the paths yourself and set `TaskEnvironment(env_vars={"_F_SYS_PATH": ...})` — verified to reach the pod on trigger-fired runs. (`Trigger(env_vars=...)` does not reach the pod; reported separately.)

## Steps to reproduce

1. `uv init --lib myproj` (code lands in `src/myproj/`), add a task:

   ```python
   # src/myproj/wf.py
   import flyte
   from myproj import hello  # any first-party import

   env = flyte.TaskEnvironment(
       "myproj",
       image=flyte.Image.from_debian_base().with_uv_project("pyproject.toml"),
   )

   @env.task(triggers=flyte.Trigger("minutely", automation=flyte.Cron("* * * * *")))
   async def main() -> str:
       return hello()
   ```

2. `flyte run src/myproj/wf.py main` — succeeds; the run's env shows `_F_SYS_PATH=./src:...`.
3. `flyte deploy src/myproj/wf.py env` and wait for the trigger to fire — the run fails with `ModuleNotFoundError: No module named 'myproj'`. The run has no `_F_SYS_PATH`.

## Expected behavior

`deploy` should capture the same bundle-relative sys.path entries `run` computes (the code bundle is built from the same workspace, so they're known at deploy time) and persist them in the deployed task so every run of the task — trigger-fired or otherwise — resolves imports identically to `flyte run`.

## Environment

- flyte SDK 2.5.9 (Python 3.14), self-hosted Flyte v2 backend

nightcityblade:
Fixes flyteorg/flyte#7675

## Summary
- persist bundle-relative local import paths in deployed task environments
- honor the no-sync-local-sys-paths configuration for deploys
- keep the original task environment variables unchanged

## Testing
- uv run python -m pytest tests/flyte/test_deploy.py -q (24 passed)
- uv run python -m ruff check src/flyte/_deploy.py tests/flyte/test_deploy.py
- uv run python -m ruff format --check src/flyte/_deploy.py tests/flyte/test_deploy.py

### DIFF
--- a/src/flyte/_deploy.py
+++ b/src/flyte/_deploy.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import copy
 import hashlib
 import os
 import pathlib
@@ -14,6 +15,7 @@ import rich.repr
 from flyte.models import ActionID, NativeInterface, RawDataPath, SerializationContext, TaskContext
 from flyte.syncify import syncify
 
+from ._constants import FLYTE_SYS_PATH
 from ._environment import Environment
 from ._image import Image
 from ._initialize import ensure_client, get_client, get_init_config, requires_initialization
@@ -163,6 +165,23 @@ class Deployment:
         return tuples
 
 
+def _with_local_sys_paths(task: TaskTemplate, root_dir: pathlib.Path) -> TaskTemplate:
+    """Return a task copy whose runtime env mirrors local imports under ``root_dir``."""
+    if not get_init_config().sync_local_sys_paths:
+        return task
+
+    root_dir_abs = pathlib.Path(root_dir).resolve()
+    env_vars = dict(task.env_vars or {})
+    env_vars[FLYTE_SYS_PATH] = ":".join(
+        f"./{pathlib.Path(path).relative_to(root_dir_abs)}"
+        for path in sys.path
+        if pathlib.Path(path).is_relative_to(root_dir_abs)
+    )
+    task_copy = copy.copy(task)
+    task_copy.env_vars = env_vars
+    return task_copy
+
+
 async def _deploy_task(
     task: TaskTemplate, serialization_context: SerializationContext, dryrun: bool = False
 ) -> DeployedTask:
@@ -181,6 +200,8 @@ async def _deploy_task(
     from ._internal.runtime.task_serde import lookup_image_in_cache, translate_task_to_wire
     from ._internal.runtime.trigger_serde import offload_trigger_inputs, to_task_trigger
 
+    assert serialization_context.root_dir is not None
+    task = _with_local_sys_paths(task, serialization_context.root_dir)
     assert task.parent_env_name is not None
     if isinstance(task.image, Image):
         image_uri: str | None = lookup_image_in_cache(serialization_context, task.parent_env_name, task.image)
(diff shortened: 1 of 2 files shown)


nightcityblade:
Follow-up pushed in 15e3911: the deployment root-dir invariant now narrows the optional type for mypy, and the amended commit includes the required DCO sign-off. Local validation: 24 deploy tests passed, Ruff check/format passed, and the full make mypy target passed (303 source files). The refreshed GitHub workflows are currently awaiting maintainer approval to run.

pingsutw (approved):
thank you for fixing it

PR #1334, opened 2026-07-28
Support plugin_config in TaskTemplate.override

pingsutw:
## Why

Setting `plugin_config` via `override()` had no effect — the task still ran as a plain python task.

Plugin behavior lives on the **template class**, not on a field. `TaskEnvironment.task()` resolves the class from `TaskPluginRegistry` at decoration time (`_task_environment.py:343`), and that class carries the `task_type` and the custom config serialization (`RayFunctionTask` / `task_type="ray"`). Because `dataclasses.replace` preserves `type(self)`, `override(plugin_config=...)` produced an `AsyncFunctionTaskTemplate` with `task_type="python"` and empty custom config on the wire.

## What

- Expose `plugin_config` as an explicit keyword argument on `TaskTemplate.override()`.
- Look the plugin class up in the same `TaskPluginRegistry` that `TaskEnvironment.task()` uses, with the same error when no plugin is registered, and rebuild the overridden task as that class. `task_type` is dropped from the copied fields so the plugin's own default applies.
- Overriding the config of a task that is already the plugin class keeps the class and replaces the config; omitting `plugin_config` preserves the existing one.
- Update `examples/plugins/ray_example.py` to configure the plugin at call time.

```python
run = flyte.run(hello_ray_nested.override(plugin_config=ray_config))
```

## Tests
```python
import asyncio
import typing

import ray
from flyteplugins.ray.task import HeadNodeConfig, RayJobConfig, WorkerNodeConfig

import flyte.remote
import flyte.storage


@ray.remote
def f(x):
    return x * x


ray_config = RayJobConfig(
    head_node_config=HeadNodeConfig(ray_start_params={"log-color": "True"}),
    worker_node_config=[WorkerNodeConfig(group_name="ray-group", replicas=2)],
    runtime_env={"pip": ["numpy", "pandas"]},
    enable_autoscaling=False,
    shutdown_after_job_finishes=True,
    ttl_seconds_after_finished=300,
)

image = (
    flyte.Image.from_debian_base(name="ray")
    .with_apt_packages("wget")
    .with_pip_packages("ray[default]==2.46.0", "flyteplugins-ray", "pip", "mypy")
)

task_env = flyte.TaskEnvironment(
    name="hello_ray", resources=flyte.Resources(cpu=(1, 2), memory=("400Mi", "1000Mi")), image=image
)
ray_env = flyte.TaskEnvironment(
    name="ray_env",
    image=image,
    resources=flyte.Resources(cpu=(3, 4), memory=("3000Mi", "5000Mi")),
    depends_on=[task_env],
)


@task_env.task()
async def hello_ray():
    await asyncio.sleep(20)
    print("Hello from the Ray task!")


@ray_env.task
async def hello_ray_nested(n: int = 3) -> typing.List[int]:
    print("running ray task")
    t = asyncio.create_task(hello_ray())
    futures = [f.remote(i) for i in range(n)]
    res = ray.get(futures)
    await t
    return res


if __name__ == "__main__":
    flyte.init_from_config()
    run = flyte.run(hello_ray_nested.override(plugin_config=ray_config))
    print("run name:", run.name)
    print("run url:", run.url)
    run.wait()

    action_details = flyte.remote.ActionDetails.get(run_name=run.name, name="a0")
    for log in action_details.pb2.attempts[-1].log_info:
        print(f"{log.name}: {log.uri}")

```
Added to `tests/user_api/test_override.py` using a locally registered dummy plugin: class swap, re-override on a plugin task, preserve-on-omit, and the unregistered-config error.

Full suite, this branch vs. a clean `main` worktree (`pytest tests/flyte tests/user_api`):

| | passed | failed |
|---|---|---|
| this branch | 3633 | 18 |
| `main` | 3630 | 18 |

No regressions — the +3 passed are the new tests. All 18 failures reproduce on `main` and are environmental (docker `buildx --push` to a registry at `localhost:30000`, obstore/minio storage tests). The two failures that are not identical between the runs — `test_image.py::test_install_flyte_false_builds_full_image` and `test_int_image_builder.py::test_real_build` — are cwd-sensitive and flaky respectively; the first fails on unmodified `main` source too when run from the repo root.

Verified against the real Ray plugin — `override(plugin_config=RayJobConfig(...))` now yields `RayFunctionTask`, and `get_proto_task` emits `type: "ray"` with a populated `rayCluster` custom config.

https://claude.ai/code/session_01F4ZdQ2hENGqdWF57j1GfEt

### DIFF
--- a/src/flyte/_task.py
+++ b/src/flyte/_task.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import weakref
-from dataclasses import dataclass, field, fields, replace
+from dataclasses import MISSING, dataclass, field, fields, replace
 from inspect import iscoroutinefunction
 from typing import (
     TYPE_CHECKING,
@@ -53,6 +53,34 @@ AsyncFunctionType: TypeAlias = Callable[P, Coroutine[Any, Any, R]]
 SyncFunctionType: TypeAlias = Callable[P, R]
 FunctionTypes: TypeAlias = AsyncFunctionType | SyncFunctionType
 F = TypeVar("F", bound=FunctionTypes)
+
+
+def _rebuild_as(
+    task: TaskTemplate[P, R, F], plugin_task_class: type[AsyncFunctionTaskTemplate[P, R, F]]
+) -> AsyncFunctionTaskTemplate[P, R, F]:
+    """
+    Rebuild `task` as `plugin_task_class`, which is how a plugin's behavior is applied since
+    `dataclasses.replace` cannot change the class.
+
+    A field is carried over unless the plugin class redefines its default (`task_type`,
+    `task_type_version`, `debuggable`, ...) and `task` still holds its own class default, in which
+    case the plugin's default wins. That way a task that never touched a field picks up the
+    plugin's value, while anything explicitly set on the task is preserved.
+    """
+    task_defaults = {f.name: f.default for f in fields(type(task))}
+    plugin_defaults = {f.name: f.default for f in fields(plugin_task_class)}
+
+    carried = {}
+    for f in fields(task):
+        if not f.init:
+            continue
+        value = getattr(task, f.name)
+        default = task_defaults.get(f.name, MISSING)
+        plugin_default = plugin_defaults.get(f.name, MISSING)
+        if plugin_default is not MISSING and plugin_default != default and value == default:
+            continue
+        carried[f.name] = value
+    return plugin_task_class(**carried)
 
 
 @dataclass(kw_only=True)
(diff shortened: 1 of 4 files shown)


src/flyte/_task.py
```diff
@@ -467,6 +482,16 @@ def override(
             **kwargs,
         )
 
+        if task_template_class is not None and not isinstance(new_task, task_template_class):
+            # Plugin behavior (task_type, custom config serialization) lives on the template class, which
+            # `replace` cannot change, so rebuild as the plugin's class.
+            task_template_class = cast(type[AsyncFunctionTaskTemplate[P, R, F]], task_template_class)
+            new_task = task_template_class(
+                **{f.name: getattr(new_task, f.name) for f in fields(new_task) if f.init and f.name != "task_type"}
```
AdilFayyaz: doing this means that we want the class default value to pass for task_type but the same can happen for other fields the user didn't actually set. Shouldn't all default values pass through?

AdilFayyaz (approved):
LGTM

PR #1338, opened 2026-07-29
feat(image): Allow extra docker buildx flags via env var

earlybard:
## Why

Allow custom `docker buildx build` arguments to be appended to the image build. The case I'm specifically after is swapping to Zstd compression by appending `--output=type=image,push=true,compression=zstd,oci-mediatypes=true `

I can see other use cases too, such as `--label`, `--annotation`, or `--sbom`.

## What

Adds a `FLYTE_DOCKER_BUILD_EXTRA_ARGS` env var. Its value is split with `shlex` and appended to the flags flyte generates, after them and before the positional build context.


pingsutw (approved):
Thanks for adding it

PR #1339, opened 2026-07-29
fix: redact signed-URL credentials from upload error messages (FLYTE-SDK-6R)

EngHabu:
The non-retryable-error branch of `_upload_with_retry` embedded the full pre-signed object-store URL in the `RuntimeSystemError` message:

```
Failed to upload /tmp/.../fast48d4...tar.gz to https://...s3.us-west-2.amazonaws.com/...?X-Amz-Algorithm=...&X-Amz-Credential=ASIA...&X-Amz-Security-Token=IQoJb3JpZ2luX2Vj...
```

That query string is the part that makes a pre-signed URL usable — `X-Amz-Signature`, `X-Amz-Credential`, and for STS-issued upload locations a complete `X-Amz-Security-Token`. So every failed bundle/spec upload leaked short-lived credentials into local logs *and* into Sentry crash reports.

**Fix:** keep scheme/host/path (the diagnostic part — it identifies the bucket and key the PUT targeted) and replace the query string with `?<redacted>`.

Side benefit: the signature and expiry differ on every attempt, so the un-redacted message gave each failure a unique Sentry fingerprint. Redacting collapses these into a single groupable issue.

Tests: unit coverage for `_redact_signed_url` plus a regression test asserting the token/signature never appear in the raised message.

fixes FLYTE-SDK-6R

### DIFF
--- a/src/flyte/remote/_data.py
+++ b/src/flyte/remote/_data.py
@@ -76,6 +76,23 @@ def _parse_retry_after(value: typing.Optional[str], cap_sec: float) -> typing.Op
     return min(seconds, cap_sec)
 
 
+def _redact_signed_url(url: str) -> str:
+    """Strip the query string off a pre-signed object-store URL.
+
+    The query string of a pre-signed URL carries the credential material that makes
+    it usable: ``X-Amz-Signature``, ``X-Amz-Credential`` and, for STS-issued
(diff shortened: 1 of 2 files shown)


EngHabu:
LGTM

pingsutw (approved):

PR #1343, opened 2026-07-30
fix(trigger): render trigger automation without crashing in `flyte get trigger`

jmonty42:

`flyte get trigger` crashes for any project that has triggers:

```
$ flyte get trigger -p ketan -d development
...
AttributeError: type
```

`Trigger.__rich_repr__` is called for every trigger in the list, and it delegates to
`_rich_automation`, which reads the automation type enum off the *message class*:

```python
if automation.type == common_pb2.TriggerAutomationSpec.type.TYPE_NONE:
```

`TriggerAutomationSpec` has no `type` attribute — in the generated `flyteidl2.task.common_pb2`
the enum is the module-level `TriggerAutomationSpecType` (`TYPE_UNSPECIFIED`/`TYPE_NONE`/`TYPE_SCHEDULE`).
Listing triggers is therefore unusable in every output format (`table`, `table-simple`, `json`).

Fixing only the enum accessor still renders the wrong thing, because the rest of the function has
three more defects:

1. `automation.schedule.cron is not None` — `cron` is a sub-message, so it is *never* `None`. Every
   schedule takes the cron branch and yields an (often empty) `Cron` message as the value; the
   fixed-rate branch is dead code.
2. `automation.HasField('start_time')` raises `ValueError: Protocol message TriggerAutomationSpec has
   no "start_time" field` — `start_time` lives on `FixedRate`, not on the automation spec. So the
   fixed-rate branch would crash if it were reachable.
3. The `cron_expression` member of the `Schedule.expression` oneof is never handled.

- Read the automation type from the module-level `TriggerAutomationSpecType` enum.
- Select the schedule kind from the oneof tag (`schedule.WhichOneof("expression")`) instead of
  `is not None` checks on sub-messages, and handle all three members (`cron`, `cron_expression`,
  `rate`).
- Read `start_time` off the `FixedRate` message, and render the rate unit by enum name rather than as
  a raw int.
- Yield a single `automation` field describing the schedule in one line, instead of a different field
  name per automation kind. `flyte.cli._common._table_format` takes its column headers from the first
  row, so variable field names misalign columns when a project mixes cron, fixed-rate, and
  unautomated triggers.

Sample output, previously an unconditional crash:

```
                                    Triggers
┌──────────────────────────────────────┬───────────────────┬─────────┬─────────┐
│ Task_name                            │ Name              │ Automat │ Auto_ac │
│                                      │                   │ ion     │ tivate  │
╞══════════════════════════════════════╪═══════════════════╪═════════╪═════════╡
│ sdk_version_tester.test_all_versions │ test_sdk_every_5m │ every 5 │ True    │
│                                      │                   │ minutes │         │
│                                      │                   │ startin │         │
│                                      │                   │ g at    │         │
│                                      │                   │ now     │         │
└──────────────────────────────────────┴───────────────────┴─────────┴─────────┘
```

## How was this patch tested?

- New `tests/flyte/remote/test_trigger_rich_repr.py`: parametrized coverage of every automation
  branch (unset type, `TYPE_NONE`, `cron` with and without a timezone, `cron_expression`, fixed rate
  with and without `start_time`, and `TYPE_SCHEDULE` with no expression set), plus a test asserting
  the CLI table keeps one stable column set across mixed automation kinds.
- `pytest tests/flyte/remote tests/flyte/cli` — 496 passed.
- Full `pytest -k "not integration and not sandbox" tests` — the only failures are pre-existing in
  this environment (`tests/flyte/ai/mcp`, `test_caller_frame_resolution`, `test_keyring_store`);
  they fail identically with this change stashed.
- `ruff format --check`, `ruff check`, `mypy`, and `ty check` clean on the changed files.
- Verified live against a real trigger (`flyte get trigger -p ketan -d development` on a canary
  deployment) in `table`, `table-simple`, and `json` output formats, plus the single-trigger detail
  view.

### Labels

fixed

## Check all the applicable boxes

- [ ] I updated the documentation accordingly.
- [x] All new and existing tests passed.
- [x] All commits are signed-off.

### DIFF
--- a/src/flyte/remote/_trigger.py
+++ b/src/flyte/remote/_trigger.py
@@ -20,6 +20,37 @@ from ._common import ToJSONMixin
 from ._task import Task, TaskDetails
 
 
+def _describe_automation(automation: common_pb2.TriggerAutomationSpec) -> str:
+    """
+    Render a trigger's automation specification as a single human-readable line.
+
+    One line (rather than one field per automation kind) keeps the columns aligned when a list of
+    triggers with differing automations is formatted as a table.
+    """
+    if automation.type == common_pb2.TriggerAutomationSpecType.TYPE_NONE:
+        return "none"
+    if automation.type != common_pb2.TriggerAutomationSpecType.TYPE_SCHEDULE:
+        return common_pb2.TriggerAutomationSpecType.Name(automation.type)
+
+    schedule = automation.schedule
+    # `schedule` is a message, so its sub-messages are never None; the oneof tag is what says
+    # which kind of schedule was actually set.
+    match schedule.WhichOneof("expression"):
+        case "cron":
+            return f"cron: {schedule.cron.expression} ({schedule.cron.timezone or 'UTC'})"
+        case "cron_expression":
+            return f"cron: {schedule.cron_expression}"
+        case "rate":
+            rate = schedule.rate
+            unit = common_pb2.FixedRateUnit.Name(rate.unit).removeprefix("FIXED_RATE_UNIT_").lower()
+            if rate.value != 1:
+                unit += "s"
+            start = rate.start_time.ToDatetime() if rate.HasField("start_time") else "now"
+            return f"every {rate.value} {unit} starting at {start}"
+        case _:
+            return "schedule: unset"
+
+
 @dataclass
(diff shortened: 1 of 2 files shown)


pingsutw (approved):
Thanks for fixing it

PR #1350, opened 2026-07-30
feat: populate TaskTemplate.reuse_policy for reusable tasks

pingsutw:
## Overview

Populate the structured `TaskTemplate.reuse_policy` field from `ReusePolicy` during serialization, so a task's environment-reuse configuration is carried as a first-class field instead of being encoded inside the untyped `custom` Struct.

- **ray**: the reuse policy moves to `reuse_policy`; `custom_config` no longer writes a `reusePolicy` key into the RayJob spec (the "exactly 1 replica" validation stays). Cluster-backed plugins read the policy from `reuse_policy`.
- **actor**: also populates `reuse_policy`, while keeping the existing `actor` spec in `custom` for backward compatibility with readers that predate the field.

A shared `reuse_policy_to_pb()` helper converts the `ReusePolicy` dataclass into the proto message; `reuse.py` gains `from __future__ import annotations` so the module still imports against older `flyteidl2`.

## Dependency

`TaskTemplate.reuse_policy` landed upstream (flyteorg/flyte#7731) and shipped in **flyteidl2 2.0.37**; this PR bumps the pin (and `rs_controller`'s) to 2.0.37, so serialization works against the released package — no local/unpublished build needed.

## Test plan

- Verified end-to-end that serializing a reusable **actor** task sets `reuse_policy` (min/max replicas, concurrency, idle/scaledown TTLs) **and** keeps the `actor` spec in `custom`.
- Updated `plugins/ray/tests/test_task.py`: `test_custom_config_omits_reuse_policy` asserts `custom` no longer carries `reusePolicy`; the multi-replica rejection test is unchanged.


examples/plugins/ray_reusable_cluster.py
```diff
@@ -41,7 +41,7 @@ def square(x: int) -> int:
     plugin_config=ray_config,
     image=image,
     resources=flyte.Resources(cpu=(1, 2), memory=("2000Mi", "4000Mi")),
-    reusable=flyte.ReusePolicy(replicas=1, idle_ttl=600),
+    reusable=flyte.ReusePolicy(replicas=1, idle_ttl=600, scope="run"),
```
pingsutw, line 44: will update union doc

AdilFayyaz (approved):

AdilFayyaz:
The PR desc still says 2.0.32 btw

PR #1353, opened 2026-07-31
Upgrade flyteidl to 2.0.31

iaroslav-ciupin:
Upgrade flyteidl to 2.0.31 which contains a breaking change cluster to queue renaming


iaroslav-ciupin:
already done in other PR

PR #1355, opened 2026-07-31
opentelemetry tracing and grafana agent observability for agents

samhita-alla:
Adds distributed tracing to Flyte plus a bridge that sends agent generations to Grafana Agent Observability. Two new plugins and one small hook in core.

## The problem

A durable run is not one process. It crashes, resumes, retries and each attempt starts a fresh OpenTelemetry SDK that knows nothing of the earlier ones. Point a normal OTel setup at a durable agent and two things go wrong: every attempt mints its own trace id, so one agent run arrives as several unrelated traces; and every step the resumed run replayed out of its durable log is missing entirely because replayed steps never execute and nothing instruments them. The trace ends up with holes exactly where durability did its job.

## What is here

`flyte._observe` (core): a dependency-free hook exposing two moments: a task starting and a `flyte.trace` step running. Context-manager based so an observer's span is active while user code runs and any instrumentation library's spans nest inside it. Core never imports an observability SDK. Called from `_trace.py` (all four wrappers) and `taskrunner.run_task`;
when nothing is registered the path is a dict lookup and no allocation.

`flyteplugins-otel`: turns those callbacks into spans. Ordinary distributed tracing for ordinary tasks plus two durability-specific behaviours. When no trace context arrives from outside, the trace id is derived from the run identity, so every process computes the same value and a crashed-and-resumed run is one trace with no coordination. Steps served from the durable log are recorded as spans marked `flyte.replayed` so a resumed trace has no holes.

`flyteplugins-agento11y`: sends generations, tool calls, token usage and cost to Grafana, nested inside the Flyte task span and grouped by Flyte run. Binds Flyte identity onto agento11y's own context: run name to conversation id, task name to agent name, task version to agent version. Both bindings are switchable since a task driving several agents should
keep each framework's own names.

`flyteplugins-agents-core`: two registries so an out-of-tree package can instrument frameworks the adapters drive. Agent instrumentation libraries attach at the call site and the adapter owns that call, so a user going through `run_agent` has nowhere to put a handler. Adapters offer their framework-native payload on the way past and separately can have their call substituted. Both know nothing about any vendor.

Six adapters wired: langchain, langgraph, openai, claude, google, pydantic_ai. 

Grafana links for the Flyte UI. `GrafanaTrace` queries on `flyte.run_name` rather than addressing a trace by id so it works whatever the trace ids turn out to be. `GrafanaAgentObservability` opens the run's conversation.

1. Links show up in the UI when configured
image

2. Agents show up in the Grafana UI
image

3. Agent conversations
image

4. Agent traces
Screenshot 2026-07-31 at 10 06 14 PM


cosmicBboy:
@samhita-alla I think `uv lock` needs to be run again

samhita-alla:
@cosmicBboy done

cosmicBboy (approved):

PR #1356, opened 2026-07-31
Add control-plane reporting for tracked runs (flyte run --tracked)

EngHabu:
## Overview

Tracked runs: the SDK keeps orchestrating everything locally while reporting run state — actions, attempts, inputs/outputs, HTML reports — to the control plane, so local runs appear live in the console. Builds on the IDL in flyteorg/flyte#7737 and the `TrackedRunService` rename (released as flyteidl2 2.0.40, which this PR pins).

### How it works

- **RemoteRunReporter** — a third `RunRecorder` sink alongside the TUI tracker and SQLite store. The recorder observer contract is synchronous and fires from async controller code *and* the threads that run sync tasks, so the sink is a sync enqueue facade over async I/O: `record_*` calls capture a fully-computed event under a lock (per-attempt monotonic versions; attempts start at 1 per the server contract) and a dedicated background event loop batches `TrackedRunService.ReportActions` calls and performs the signed-URL uploads (the same background-loop shape as `flyte.syncify`). Terminal events trigger a bounded flush barrier at run end. Reporting is best-effort and never fails or hangs the run (unless strict mode is on).
- **Data path routes like everything else**: artifacts upload via `DataProxyService.CreateUploadLocation` signed PUT URLs with the `tracked-runs/<run>/<action>[/<attempt>]` filename_root scheme, routed by `ClusterService.SelectCluster(OPERATION_TRACKED_RUN_DATA)` — an org with a directly-reachable data plane cluster uploads straight to that cluster's store; an org without one is served by the control plane's own store. The reporter stamps the routing cluster onto reported attempt events so reads later route to the same store. Inputs upload before CreateRun (deterministic path, no URI on the event); outputs/report URIs attach to the terminal event after their uploads complete. Reports also flush live mid-run so the report tab updates while the run executes.
- **Opt-in surface** (one naming family, anchored on the flag):
  - `flyte run --local --tracked [--tracked-strict]`
  - config: `local.tracked` / `local.tracked_strict` (section mirrors `local.persistence`)
  - `flyte.init(local_tracked=..., local_tracked_strict=...)`; `flyte create config --local-tracked`
  - runcontext: `flyte.with_runcontext(mode="local", tracked=True)`
- Requires project/domain; degrades gracefully with one warning when no client is initialized. `--tracked-strict` fails the run loudly on the first reporting failure (for debugging reporting itself). SIGINT/SIGTERM abort-flushes in-flight actions as ABORTED. `Run.url` points at the console tracked-runs page. Falsy outputs (`0`, `""`, `[]`, `False`) report correctly (presence, not truthiness). File/Dir raw data never uploads — only the three metadata artifacts (inputs.pb / outputs.pb / report.html) reach the dataproxy.

The deck feature (`@env.task(report=True)`, `flyte.report`) and the `@trace` decorator are unrelated and untouched.

## Test Plan

- 688 unit tests green on the released flyteidl2 pin (reporter lifecycle incl. retries/ordering/flush barrier/failure isolation/strict mode, upload helper, client protocol, CLI flags, config plumbing).
- Live E2E against three environments running a backend that implements `TrackedRunService`:
  - a single-process development stack: full lifecycle, watches, replay idempotency.
  - a hosted control plane with a directly-reachable data plane cluster: runs + inputs/outputs + reports render in the console; artifacts land in the cluster's metadata store, with the routing cluster recorded on attempts.
  - a hosted control plane where artifacts are served by the control plane's own store — the fallback leg of the routed data path, verified end to end on the released flyteidl2 pin.

## Rollout Plan

The feature is opt-in per run/config and requires a control plane that implements `TrackedRunService`; nothing changes for users who don't pass `--tracked`.

## Rollback Plan

Revert; the feature is client-side and opt-in.

### DIFF
--- a/src/flyte/_persistence/_remote_reporter.py
+++ b/src/flyte/_persistence/_remote_reporter.py
@@ -1,0 +1,1190 @@
+"""Report tracked-run state to the control plane via `TrackedRunService`.
+
+`RemoteRunReporter` is a third `RunRecorder`
+backend (next to the TUI tracker and the SQLite `RunStore`).
+
+Why the facade is synchronous when the SDK is async-first: the `RunRecorder`
+observer contract is synchronous by design and fires from async controller code
+*and* from the worker threads that run synchronous tasks — there is no single
+event loop the reporter could await on, and a recorder call must never block or
+fail the run it is observing. So every `record_*` call only captures a
+lightweight, fully-computed event and enqueues it on a thread-safe queue. All
+actual I/O is async: a dedicated background worker thread hosts its own event
+loop (the same shape as `flyte.syncify`'s `_BackgroundLoop` and the remote
+controller's worker model) and runs the async pipeline there — batching queued
+events into `TrackedRunService.ReportActions` calls on the async client and
+performing the outputs/report signed-URL uploads for terminal events. The
+dedicated loop also keeps the terminal flush working during interpreter
+shutdown and `KeyboardInterrupt`, when the caller's loop may already be gone.
+Async callers can use `aclose`; `close` is the sync barrier used at
+process-exit points.
+
+Reporting is strictly best-effort: enqueue methods never raise, the worker retries a
+bounded number of times and then drops the batch with a warning, and the terminal
+flush (`close`) waits a bounded amount of time so
+`flyte run --local` never hangs on a slow control plane.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import queue
+import threading
+import uuid
+from dataclasses import dataclass, field
+from datetime import datetime, timezone
+from typing import TYPE_CHECKING, Any
+
+from flyte._logging import logger
+
+if TYPE_CHECKING:
+    from flyteidl2.common import identifier_pb2
+    from flyteidl2.workflow import tracked_run_service_pb2
+
+    from flyte._task import TaskTemplate
+    from flyte.remote._client.controlplane import ClientSet
+
+ROOT_ACTION_NAME = "a0"
+_MAX_RUN_NAME_LENGTH = 30
+# Run names starting with these prefixes are reserved by the platform.
+_RESERVED_RUN_NAME_PREFIXES = ("u", "r")
+
+# `flyteidl2.common.ActionIdentifier.name` is capped at 30 characters. Local action ids
+# have no such bound — they are built by concatenating the parent id, the task name and a
+# sequence number — so nesting past the first level produces ids the API rejects.
+_MAX_ACTION_NAME_LENGTH = 30
+_ACTION_NAME_DIGEST_LENGTH = 8
+
+
+def _bounded_action_name(action_id: str) -> str:
+    """Shorten a local action id to something the control plane will accept.
+
+    Truncating alone would collide (sibling actions share long common prefixes, since the
+    parent id is a prefix of every child's), so the tail is a digest of the full id. The
+    result is a pure function of the input: two call sites that never coordinated — a
+    child reporting itself and its parent being referenced — still agree.
+    """
+    if len(action_id) <= _MAX_ACTION_NAME_LENGTH:
+        return action_id
+    digest = hashlib.sha256(action_id.encode("utf-8")).hexdigest()[:_ACTION_NAME_DIGEST_LENGTH]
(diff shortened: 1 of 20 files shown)


katrogan (dismissed):

src/flyte/_persistence/_remote_reporter.py
```diff
@@ -0,0 +1,1176 @@
+"""Report tracked-run state to the control plane via `TrackedRunService`.
```
kumare3, line 1: i think this is doing a lot of IO and I feel should be async

src/flyte/config/_config.py
```diff
@@ -179,12 +179,19 @@ class LocalConfig(object):
     """Configuration for local execution settings."""
 
     persistence: bool = False
+    # Tracked runs: mirror a locally-orchestrated run onto the control plane via
+    # TrackedRunService. The `local.` section name is kept (these configure local
+    # execution, and renaming it would break existing config files).
+    tracked: bool = False
+    tracked_strict: bool = False
```
kumare3, line 186: what is the run name? do you want to support supplying your own name

kumare3:
ideally should use async in the io layer

src/flyte/_persistence/_remote_reporter.py
EngHabu: The IO layer is async — every network call (`ReportActions` batching, retries, the signed-URL uploads) is a coroutine running on the async connect clients (`_process_batch`, `_send_with_retries`, `_upload_inputs`, `_upload_terminal_artifacts`, `_upload_live_report`). What's synchronous is only the thin `record_*` enqueue facade, and that's forced by the `RunRecorder` observer contract: it fires synchronously from async controller code *and* from the threads running sync tasks, so there's no single loop to await on, and an observer call must never block or fail the run it's observing. Events go on a thread-safe queue and a dedicated background event loop drains them — the same background-loop shape as `flyte.syncify`'s `_BackgroundLoop` and the remote controller's worker model. The dedicated loop is also what keeps the terminal flush working during `KeyboardInterrupt`/interpreter shutdown when the caller's loop is already gone. Async callers get `aclose()`; `close()` is the sync exit barrier. The module docstring now documents this rationale (pushed in 9920d335).

src/flyte/config/_config.py
EngHabu: Supplying your own name is supported through the existing surface: `flyte run --name my-run --local --tracked` or `flyte.with_runcontext(name="my-run", mode="local", tracked=True)`. A user-supplied name is validated against the reporting contract (≤30 chars, and it can't start with the reserved `u`/`r` prefixes the platform uses for routing generated names). When no name is given we generate `local-<8 hex chars>` — the `l` prefix keeps generated tracked-run names disjoint from platform-generated ones. This config key is just the on/off switch for reporting; naming stays with the run, not the config.

EngHabu:
Re: async in the IO layer — it is async end to end: all network IO (`ReportActions` batches with retries, the signed-URL artifact uploads) runs as coroutines on the async connect clients, hosted on a dedicated background event loop. The synchronous part is only the `record_*` enqueue facade, which the `RunRecorder` observer contract forces (it fires from async controller code and from sync-task worker threads, and must never block the run). Same background-loop pattern as `flyte.syncify`'s `_BackgroundLoop`. Wrote the rationale into the module docstring in 9920d335; details in the inline reply on `_remote_reporter.py`.

src/flyte/_persistence/_remote_reporter.py
EngHabu: Follow-up after a line-by-line audit of the I/O paths (prompted by this comment): one genuinely blocking call was hiding in the async pipeline — `_read_report` did a synchronous `read_bytes()` of the local report.html inside `_upload_terminal_artifacts` on the worker loop. Fixed in 2e77dc5d (`asyncio.to_thread`). Everything else checks out as async: the bootstrap (`CreateRun` + root inputs upload) is awaited on the runner's loop; the worker awaits `report_actions` on the async connect client (pyqwest transport — per-call awaitables, so safe from the dedicated loop), the signed-PUT uploads use an `httpx.AsyncClient` created inside the worker loop, and backoffs are `asyncio.sleep`. The remaining sync work is CPU-only (proto serialization at enqueue time, MD5 over in-memory bytes).

kumare3 (approved):

PR #1357, opened 2026-07-31
Revert #1312

pingsutw:
Revert #1312 because the backend plugin has been updated to use the latest Dask script to start the workers and scheduler, making this change no longer necessary.

https://github.com/flyteorg/flyte/pull/7707


katrogan (approved):
woohoo!

PR #1362, opened 2026-08-03
Artifacts and Artifact-Triggers

kumare3:
## Summary

Stacked on `artifacts`. Adds the third `Trigger` automation type alongside `Cron` and `FixedRate` — fire a run when a new version of a named artifact is created:

```python
retrain = flyte.Trigger(
    name="retrain-on-new-model",
    automation=flyte.OnArtifact(name="customer_model"),  # any new version
    inputs={"model": flyte.TriggeredArtifact, "threshold": 0.5},
)

@env.task(triggers=[retrain])
async def consumer(model: File, threshold: float = 0.5) -> str: ...
```

- `OnArtifact(name, version=None)` — watch an artifact by name; a version pin fires only on exactly that version
- `flyte.TriggeredArtifact` sentinel binds the triggering artifact's value to one task input (the artifact-trigger analogue of `TriggerTime`); other inputs keep their registered defaults, offloaded exactly as schedule triggers do today
- Serde emits `TriggerAutomationSpec(TYPE_ARTIFACT, ArtifactTrigger{artifact_name, version, input_arg})` with the sentinel key as `input_arg`, validated against the task interface; no kickoff context key is emitted — the backend fire step (leaseworker plugin) injects the real literal into the offloaded inputs
- Cross-validation at construction: at most one sentinel, sentinel requires `OnArtifact`, `TriggerTime` rejected on artifact triggers
- `flyte get trigger` / rich repr render `on artifact: name@version -> input`
- Runnable example: `examples/triggers/artifact_trigger.py`

## Dependencies

- Protos: flyteorg/flyte#7749 (`TYPE_ARTIFACT` + `ArtifactTrigger`). **CI will fail on the new serde tests until a flyteidl2 release carries it and the pin is bumped** — developed against a local wheel build.
- Backend stack: unionai/cloud#17381–#17385.

E2E-verified on devbox: deploy → producer publishes a new version → downstream run fires with the artifact bound to `model` and `threshold=0.5` from the trigger defaults; republishing the same version does not re-fire.


kumare3:
Retargeted to main: this PR now carries the full SDK artifacts + triggers work in one (the `artifacts` branch content included). Supersedes #191.

src/flyte/_internal/runtime/convert.py
```diff
@@ -291,20 +291,38 @@ async def convert_from_native_to_outputs(o: Any, interface: NativeInterface, tas
...
+    produced: list[common_pb2.ProducedArtifact] = []
     for (output_name, python_type), v in zip(interface.outputs.items(), o):
+        # Preserve artifact metadata attached via flyte.artifacts.new(...): capture it
+        # before to_literal unwraps the wrapper (and would otherwise discard it), then
+        # emit a first-class ProducedArtifact declaration on the Outputs envelope so
+        # the backend can register the artifact when the task declares
+        # produces_artifacts. The declaration carries the declared output type (this
+        # SDK is authoritative for it). Top-level outputs only; a wrapper nested
+        # inside a model or container is rejected outright — it would otherwise
+        # serialize its inner value and silently drop the metadata.
```
wild-endeavor: don't really understand this comment.  should be inside `raise_if_nested_wrapper`.  Inside that should just be:
> Artifacts are supported only as top-level outputs from tasks.  Artifacts declared inside container structures like dataclasses or pydantic models are rejected.

src/flyte/_internal/runtime/trigger_serde.py
```diff
@@ -177,9 +186,25 @@ async def to_task_trigger(
     if kickoff_arg_name is not None:
         context_kvs.append(literals_pb2.KeyValuePair(key=KICKOFF_TIME_INPUT_ARG_CONTEXT_KEY, value=kickoff_arg_name))
 
-    # Keep the kickoff arg on the schedule too: the backend uses it for the scheduled-trigger
-    # contract (and folds run_start_time into the cache key on fire).
-    automation = _to_schedule(t.automation, kickoff_arg_name=kickoff_arg_name)
+    if isinstance(t.automation, OnArtifact):
+        # No kickoff context key: the artifact's value is injected into the offloaded inputs
```
wild-endeavor: what is kickoff context key?

examples/artifacts/artifact_example.py
```diff
@@ -0,0 +1,228 @@
...
+    image=flyte.Image.from_debian_base().with_pip_packages("pandas", "pyarrow"),
+    # pandas + pyarrow need more than the default task memory.
+    resources=flyte.Resources(cpu=1, memory="1Gi"),
+)
+
+
+# 1. A File artifact: the file uploads to blob storage; the artifact stores the
+#    typed reference.
+@env.task(produces_artifacts=True)
+async def produce_file() -> File:
```
pingsutw, line 41: Does it work if outputs are
```python
@env.task(produces_artifacts=True)
async def produce_file() -> (File, int):
```

src/flyte/types/_string_literals.py
```diff
@@ -58,24 +58,58 @@ def _scalar_to_string(scalar: literals_pb2.Scalar) -> Any:
...
+    through every copy. (Produced-artifact declarations live on the Outputs envelope —
+    see ``produced_artifact_annotation``.)
+    """
+    if not lit.HasField("artifact_id"):
+        return None
+    key = lit.artifact_id.key
+    return f"artifact: {key.org}/{key.project}/{key.domain}/{key.name}@{lit.artifact_id.version}"
+
+
+def produced_artifact_annotation(decl: "common_pb2.ProducedArtifact") -> str | None:
```
wild-endeavor: can we import the type at least for type checking so we don't have to use ""

src/flyte/_internal/imagebuild/docker_builder.py
```diff
@@ -275,12 +275,16 @@ async def handle(layer: PythonWheels, context_path: Path, dockerfile: str) -> st
...
+        # Second install (last): force the exact local wheels on top of whatever the dependency step
+        # installed. --no-index + --reinstall guarantees the local wheels win and nothing re-resolves
+        # them to published releases afterwards. This must run after the dependency step: a full resolve
+        # can otherwise discard a local wheel in favor of a stable PyPI release -- e.g. when the local
+        # wheel is a dev version or one of its dependencies can't be satisfied, uv backtracks to the
+        # published version (silently swapping a `with_local_v2()` build back to the released package).
+        # Every wheel in the dir is forced (not just layer.package_name) so sibling wheels -- e.g. a
+        # locally built flyteidl2 that the local SDK depends on -- also win over their PyPI releases;
+        # the layer hash already covers all wheels in the dir.
+        wheel_paths = sorted(f"/dist/{wheel.name}" for wheel in layer.wheel_dir.glob("*.whl") if wheel.is_file())
```
pingsutw: I think we should revert it. The build will fail if the dist directory contains wheels for different architectures. or only do it if package name is None

src/flyte/artifacts/_card.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+async def _upload_card_from_local(
+    local_path: pathlib.Path, format: CardFormat = "html", card_type: CardType = "generic"
+) -> Card:
+    # Implement upload. If in task context, upload to current metadata location, if not, upload using control plane.
+    uri = ""
+    ctx = flyte.ctx()
+    if ctx:
+        output_path = ctx.output_path + "/" + f"{card_type}.{format}"
```
pingsutw: Should we push a hash of file to the path?

src/flyte/cli/_create.py
```diff
@@ -62,6 +62,79 @@ def project(cfg: common.CLIConfig, id: str, name: str, description: str, label:
...
+
+    The file is uploaded to blob storage and stored in the artifact service as a
+    File artifact. Primitive values (strings, numbers) are not allowed as
+    artifacts: an artifact is an addressable asset, not a scalar.
+
+    \b
+    Example usage:
+
+    ```bash
+    flyte create artifact my_model --from-file model.pt -d framework=torch
```
pingsutw:
```suggestion
    flyte create artifact my_model --from-file model.pt --data framework=torch
```

src/flyte/artifacts/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+```
+
+Retrieve a set of artifacts and pass them as a list
+```python
+from flyte.remote import Artifact
+flyte.run(main, x=[Artifact.get("name1", version="1.0"), Artifact.get("name2", version="2.0")])
+```
+OR
+```python
+flyte.run(main, x=flyte.remote.Artifact.list("name_prefix", partition_match="x"))
```
pingsutw: should it be listall or list_names

src/flyte/artifacts/_card.py
pingsutw: Someone may run 2 `create_from` at the same time

examples/artifacts/artifact_example.py
kumare3: yes, it does, but let me make the test

src/flyte/artifacts/_card.py
kumare3: why?

src/flyte/_run.py
```diff
@@ -54,6 +60,78 @@
...
+    python value stored in its literal, which is what tasks actually consume.
+    Non-artifact values are returned unchanged. Also returns the artifact source (tracker
+    string, or per-element trackers for lists) so callers can record provenance.
+    """
+    # Imported lazily so ``import flyte`` does not eagerly pull in the remote package.
+    from flyte.remote import Artifact
+
+    if isinstance(value, Artifact):
+        return await value.to_python(), value.artifact_version_id
+    if isinstance(value, list) and len(value) > 0:
```
wild-endeavor, line 79: not entirely following this section.  is this from `flyte.run(...)`?  should we handle dict in addition to list?  should we error on dict?

src/flyte/_run.py
```diff
@@ -54,6 +60,78 @@
...
+    for k, v in kwargs.items():
+        unwrapped, source = await _unwrap_artifact_value(v)
+        new_kwargs[k] = unwrapped
+        if source is not None:
+            sources[k] = source
+    return tuple(new_args), new_kwargs, sources
+
+
+def _stamp_artifact_inputs(
+    inputs: Any, input_names: List[str], sources: Dict[Union[int, str], _ArtifactSource]
```
wild-endeavor: inputs is `Any`... but then we call `inputs.proto_inputs.literals` below?

src/flyte/_run.py
```diff
@@ -725,6 +803,14 @@ async def _run_remote(self, obj: TaskTemplate[P, R, F] | LazyEntity, *args: P.ar
         project = self._project or cfg.project
         domain = self._domain or cfg.domain
 
+        # Artifacts bind as normal inputs: materialize each one to its python value
```
wild-endeavor: I don't follow this section... I feel like we're taking input Artifacts, which have literals, literals which are presumably already stamped by the Artifacts service.  we're then converting into python... and then converting back into literal, and then stamping the origin.  am i reading this just all wrong?

side question - when Artifacts service stamps a literal, does it need to sign it?

src/flyte/_internal/imagebuild/docker_builder.py
kumare3: Reverted — you were right. `package_name` is always set so the other option wasn't available; a sibling wheel that needs to win now gets its own layer.

src/flyte/artifacts/_card.py
kumare3: You were right — two cards of the same kind in one task overwrote each other, so the path now includes a hash of the content.

src/flyte/artifacts/__init__.py
kumare3: Good catch — that example named a method that doesn't exist; it now uses `listall`.

src/flyte/cli/_create.py
kumare3: Already fixed, thanks.

examples/artifacts/artifact_example.py
kumare3: Confirmed it works and added tests — one caveat worth knowing: the `(File, int)` form fails in files that use `from __future__ import annotations`, so the example uses `tuple[File, float]`, which works everywhere.

src/flyte/_internal/runtime/convert.py
kumare3: Moved it — the explanation now lives in `raise_if_nested_wrapper` in your words, and the comment here is down to a couple of lines.

src/flyte/_internal/runtime/trigger_serde.py
kumare3: Fair — the comment never said what it was; rewritten to name the key and say why a schedule needs one and an artifact trigger doesn't.

src/flyte/types/_string_literals.py
kumare3: Fixed — the module already imported it, so the quotes were just noise.

src/flyte/_run.py
kumare3: Yes, this is from `flyte.run(...)`. Dicts stay unsupported, but they now raise a clear message instead of a confusing type error; lists still work.

src/flyte/_run.py
kumare3: Good catch — that function is gone now, replaced by the change below.

src/flyte/_run.py
kumare3: You read it right and it was wasteful — the artifact's literal already carries the stamp from the service, so we now use that literal directly instead of rebuilding an equivalent one, after checking it matches the input's declared type. On signing: nothing verifies the stamp today, so before this change whoever submitted a run could claim any source; using the service's literal removes that, but if we ever depend on this for audit the service would need to sign it.

src/flyte/_internal/runtime/convert.py
```diff
@@ -170,6 +170,187 @@ async def convert_from_native_to_inputs(
...
+                for field_name in model_fields:
+                    _raise_if_nested_artifact(getattr(value, field_name, None), arg_name, _depth + 1)
+
+
+def _one_line(msg: types_pb2.LiteralType) -> str:
+    """Proto text format is multi-line; flatten it so it reads inside an error message."""
+    return " ".join(str(msg).split())
+
+
+def _assert_artifact_type(artifact: Any, arg_name: str, declared_type: type) -> None:
```
wild-endeavor: i think we don't need this whole bit.  the `literal_types_compatible` function above feels like it should be in the type engine, and it already is in some sense.  we should either move the code there, or formalize what it does internally into a public function and invoke it from here. or just let it fail at runtime.

src/flyte/_internal/runtime/convert.py
```diff
@@ -170,6 +170,187 @@ async def convert_from_native_to_inputs(
...
+    stored = artifact.pb2.spec.type
+    if not literal_types_compatible(stored, declared):
+        raise ValueError(
+            f"artifact '{artifact.name}@{artifact.version}' cannot bind to input '{arg_name}': "
+            f"the artifact stores `{_one_line(stored)}` but the input is declared as "
+            f"{declared_type} (`{_one_line(declared)}`)."
+        )
+
+
+async def bind_artifact_literals(
```
wild-endeavor, line 229: i don't think necessarily think we should do this but i'm curious what claude/copilot says: should we just handle Artifacts in the type engine in dict_to_literal_map?

wild-endeavor (dismissed):

src/flyte/_internal/runtime/convert.py
kumare3: how?

kumare3: you mean in type engine when we are converting them from "what" it is actually literal type. we are just binding them. else are you saying we just pass them in

src/flyte/_internal/runtime/convert.py
kumare3: i want to compare types. the failure at runtime is weird. I think we could just "serialize" and compare

kumare3: Agreed the comparator was reinventing the type engine — it's gone. Artifacts now have a `coerce_to_literal(python_type)` method that round-trips the stored value through the type engine, so the engine decides compatibility (this also fixed `Optional[File]`, which the comparator approved but bound in the wrong shape).

src/flyte/_internal/runtime/convert.py
kumare3: Looked into it — the type engine picks its transformer from the declared type (`File`), never from the value, so an artifact would need a special check inside `to_literal` plus a remote import in the types package. Felt like the wrong layer to pay that cost, so the handling stays at run submission, but the type engine now does all the real work via the round-trip.

wild-endeavor (approved):

PR #1363, opened 2026-08-03
fix: give the signed upload URL a lifetime that covers the upload (FLYTE-SDK-5F)

EngHabu:
## The bug

`_upload_single_file` asks the control plane for a pre-signed URL valid for **60 seconds**:

```python
_UPLOAD_EXPIRES_IN = timedelta(seconds=60)
_UPLOAD_TIMEOUT_SECONDS = float(os.environ.get("FLYTE_UPLOAD_TIMEOUT", "600"))
```

...and then hands it to `_upload_with_retry`, which is allowed to spend **600s** on a single PUT and to retry it three times on top of that. The two numbers sat two lines apart and had nothing to do with each other. Any upload that takes longer than a minute to serve dies on:

```xml
<Error><Code>AccessDenied</Code><Message>Request has expired</Message>
<X-Amz-Expires>60</X-Amz-Expires><Expires>2026-08-01T01:18:50Z</Expires>...
```

A minute is easier to burn than it looks. `upload_dir` fans every file out with `asyncio.gather`, so a URL can be minted well before its PUT actually gets a turn — the FLYTE-SDK-4B events blow the window by 17 seconds while uploading `spec.pb`, a few hundred bytes.

Between them the two Sentry issues are **27 events** across releases 2.5.6 and 2.4.x, and both were bucketed as "S3-side, backend's problem" on the strength of the 403.

## The fix

- Derive the requested lifetime from the upload timeout plus a margin covering the retry backoff, the queueing above, and clock skew — the object store's clock is the one that decides whether a URL has expired. Default goes 60s → 900s, comfortably inside the hour flyteadmin will sign for. `FLYTE_UPLOAD_EXPIRES_IN` overrides it.
- Report an expired URL as its own error. Object stores answer an expired signature with the same status they use for "you may not write here" (403, or 400 when the STS token behind it lapsed), so the old message dumped S3's XML and left the user to work out which one it was. These need different fixes, and neither is worth re-PUTting the same dead signature for.

- Clamp the *derived* default to 1h: dataproxy validates `expires_in` against its `upload.maxExpiresIn` and fails `CreateUploadLocation` outright above it, so someone who raises `FLYTE_UPLOAD_TIMEOUT` should get a longer upload, not no upload. An explicit `FLYTE_UPLOAD_EXPIRES_IN` is passed through as given — the escape hatch for a deployment that raised its own cap.

## Not in scope

Re-minting the URL per retry attempt would make the retry loop correct regardless of lifetime, and bounding `upload_dir`'s fan-out would stop URLs aging in a queue. Both are real, both are bigger than this; happy to do either as a follow-up.

## Tests

11 new cases in `tests/flyte/remote/test_upload_retry.py`, covering the lifetime-vs-timeout invariant (including under a raised `FLYTE_UPLOAD_TIMEOUT`), the env override, expiry detection across AWS/GCS/Azure bodies plus the negative cases (a genuine `AccessDenied`, a 500 that happens to contain the phrase), the new error message, and redaction on the new branch. `tests/flyte/remote/` is 497 passed.

Both assertions were verified failing on `main` first, via a standalone repro rather than the test module — the module imports a new symbol, so stashing `src/` would have failed at collection and proved nothing:

```
=========== ON MAIN ===========      =========== WITH FIX ===========
expires_in=60.0s timeout=600.0s      expires_in=900.0s timeout=600.0s
[FAIL] URL outlives one attempt      [PASS] URL outlives one attempt
[FAIL] error names the knob          [PASS] error names the knob
```

## Three more issues in this class (added 2026-08-19)

A full-corpus sweep dated every `X-Amz-Date` in the unresolved issues against the event timestamp and compared it to `X-Amz-Expires`. Three more issues are the same 60s window closing mid-upload, and had been bucketed elsewhere on the strength of the 403:

| issue | events | URL age at failure | expiry |
|---|---|---|---|
| FLYTE-SDK-6C | 5 | 113s | 60s |
| FLYTE-SDK-7G | 1 | 67s | 60s |
| FLYTE-SDK-7H | 1 | 186s | 60s |

6C and 7G carry S3's `<Message>Request has expired</Message>` in the body; 7H's body is past the message truncation limit, but its URL had been dead for 126s by the time the PUT was rejected.

Worth noting where they fire: 6C, 7G and 7H are the **image-build context upload** (`remote_builder._validate_configuration` -> `upload_file`), not the code bundle. Same `_upload_single_file` call site and the same single `_UPLOAD_EXPIRES_IN` constant, so this PR covers them unchanged — but it means the 60s window is closing on `context.tar.gz` too, which is the larger of the two payloads.

## A sixth, from the next sweep (added 2026-08-20)

FLYTE-SDK-7K (1 event, 2026-08-19, release 2.6.2) is the same failure on the code-bundle
upload: `403 <Code>AccessDenied</Code><Message>Request has expired</Message>` out of
`_put_signed_url_with_retry` -> `_classify`. Its query string is redacted by #1339 so there
is no `X-Amz-Expires` to date, but the store's own message settles it.

Total for this PR is now **52 events across 6 issues**.

fixes FLYTE-SDK-5F
fixes FLYTE-SDK-4B
fixes FLYTE-SDK-6C
fixes FLYTE-SDK-7G
fixes FLYTE-SDK-7H
fixes FLYTE-SDK-7K

## A second stratum in 5F: the stalled upload (added 2026-08-25)

Re-reading all 39 of 5F's events rather than just the latest one shows it is a **mixed group**, not one bug:

| stratum | events | covered by |
|---|---|---|
| `403 AccessDenied` / `Request has expired` | 24 | the lifetime change above |
| **`400 RequestTimeout`** | **6** | **nothing — this commit** |
| retries exhausted on a network error | 8 | already retried |
| `400 ExpiredToken` | 1 | the expiry branch above |

The `RequestTimeout` stratum is a different failure with a different remedy:

```xml
<Error><Code>RequestTimeout</Code><Message>Your socket connection to the server was not read
from or written to within the timeout period. Idle connections will be closed.</Message></Error>
```

That is S3 saying the PUT body stopped arriving and it hung up — a transient stall on a slow or flaky link. But it arrives as **400**, which `_classify`'s status table reads as "malformed, do not retry", so the upload died on the first stall having never retried once. Raising the URL lifetime does not help: the request was well-formed and the signature was valid.

`RequestTimeout` is in the AWS SDKs' standard retryable set for exactly this reason, and re-sending is safe here — the PUT is idempotent, the same pre-signed URL is reused, and the file is re-opened per attempt.

The match is on the store's own `<Code>` element at status 400 only, so a genuine malformed-request 400 still fails fast, and `ExpiredToken` — same status, opposite meaning — keeps its own unretried branch.

7 new cases: the predicate's positives and negatives (including a 403/500 carrying the same body, and prose that merely mentions a timeout), a stall that recovers on retry, one that exhausts the budget and reports itself as retried, and a regression guard that `ExpiredToken` still beats the retry branch. `tests/flyte/remote/test_upload_retry.py` is 39 passed.

Both behavioral assertions were verified failing without the fix by reverting only the `_classify` wiring and keeping the new symbol defined — stashing `src/` would have failed at collection and proved nothing. Pre-fix, the assertion error reproduces 5F's Sentry message verbatim:

```
Failed to upload .../bundle.tar.gz to https://signed.url/upload, status 400:
<Error><Code>RequestTimeout</Code>...
```


EngHabu:
Rebased onto `main` — it had drifted to `CONFLICTING`.

The conflict was substantive rather than mechanical, so flagging what changed. `main` has since refactored `_upload_with_retry` into a thin wrapper over a new shared `_put_signed_url_with_retry`, which also backs the tracked-run metadata (bytes) upload path. This branch's expired-signature branch lived in the old inline body.

Resolution: took `main`'s structure and moved the `_is_expired_signed_url` check into the shared helper's `_classify`, rather than keeping a private copy in the wrapper. Net effect is that the bytes upload path now gets expired-URL detection too, which it would not have had if I'd resolved the other way.

`tests/flyte/remote/`: 581 passed, including all 28 in `test_upload_retry.py`. Sign-off preserved on both commits.

EngHabu (commented):
LGTM

PR #1368, opened 2026-08-04
Expose task serialization as a public method

katrogan:


src/flyte/extras/_serialize.py
```diff
@@ -0,0 +1,49 @@
...
+
+# Default version stamped onto the TaskSpec when the caller does not supply one
+# through a SerializationContext.
+_PLACEHOLDER_VERSION = "serialized"
+
+
+def _default_ctx() -> SerializationContext:
+    return SerializationContext(
+        version=_PLACEHOLDER_VERSION,
+        code_bundle=None,
```
pingsutw, line 31: Do we expect users to add the source code to image?

pingsutw (dismissed):

src/flyte/extras/_serialize.py
```diff
@@ -0,0 +1,49 @@
...
+        root_dir=pathlib.Path.cwd(),
+    )
+
+
+def serialize(task: TaskTemplate, ctx: Optional[SerializationContext] = None) -> task_definition_pb2.TaskSpec:
```
pingsutw, line 36: I feel like we can reuse this https://github.com/flyteorg/flyte-sdk/blob/4955aa4d70514f92380eb9029cbd6372d281d7a9/src/flyte/_run.py#L325-L335

pingsutw: If we just want to use in cloud, we can move it to https://github.com/flyteorg/flyte-sdk/blob/main/src/flyte/extras/__init__.py, so that we can import it in cloud.

src/flyte/extras/_serialize.py
katrogan: The source must be baked into the image (or mounted at run time); serialize just doesn't add it. Added this to the docstring.

src/flyte/extras/_serialize.py
katrogan: Serialize already reuses the same function from there: `translate_task_to_wire`. I chose to use that rather than call `_build_task_spec_from_template` directly because this also builds images + code bundles and needs a _Runner/client, which would defeat the offline, code-agnostic purpose here. Updated the docstring.

katrogan: > If we just want to use in cloud, we can move it to https://github.com/flyteorg/flyte-sdk/blob/main/src/flyte/extras/__init__.py, so that we can import it in cloud.

Done, thanks!

pingsutw (approved):

PR #1370, opened 2026-08-04
fix: serialize concurrent code-bundle download/extract to prevent worker startup crashes in clustered tasks

AdilFayyaz:
## Issue
Tracks the following Issue on Linear: [ENG26-1039](https://linear.app/unionai/issue/ENG26-1039/clustered-task-environments-integration-test-failing)

## Motivation

  Clustered (JobSet) tasks launch nproc_per_node separate a0 processes per pod via torchrun, and each one independently runs download_bundle into the same destination directory. The concurrent downloads and tar --overwrite extractions race with each other, intermittently crashing workers at startup (tar: <file>: Cannot open: No
  such file or directory); each crash burns a JobSet restart, which is what has been failing test_clustered_failure_cascade in the scheduled integration tests ("jobset failed due to reaching max number of restarts").

  ## Summary

  - download_bundle now serializes download+extract per bundle behind an exclusive flock (dest/.<basename>.lock), with a completion marker (.extracted / .done) as the "ready" signal — the archive's existence alone is not trusted, since it appears on disk before extraction finishes.
  - Downloads land atomically via mkstemp + os.replace, fixing the non-atomic fsspec fallback that could expose a half-written tgz/pkl to a peer process.
  - The lock waits without a timeout and is never unlinked; the kernel releases it if the holder dies, and the next waiter redoes the work (no marker → re-download).
  Where locking is unavailable (no fcntl, read-only destination) behavior degrades to the previous semantics.
  - The pkl branch gets the same protection (previously a peer could gzip.open a half-written pkl). No new dependencies; public signature unchanged.

 ## Test Plan

  - New tests/flyte/code_bundle/test_download_bundle.py (11 tests): happy path, marker fast-path, lock-wait, stale partial archive, corrupt bundle retry, same-process and multi-process concurrency, degraded mode, marker-write failure, pkl branch, cross-bundle non-contention.
  - The multi-process test reproduces the production race against the pre-fix code (fails 3/3 with the same tar collision) and passes with the fix: git stash && uv  run python -m pytest tests/flyte/code_bundle/test_download_bundle.py::test_multi_process_concurrency vs. unstashed.
  - Full unit suite passes (make unit_test); only pre-existing unrelated failure (test_keyring_store).
  - End-to-end: 5/5 consecutive examples/clustered/failure_cascade.py runs succeeded on playground canary with the fixed wheel baked into the image and a CI-sized bundle (1863 files, 38 MB, --copy-style all), each exercising crash → JobSet restart → completion with no tar errors.

### DIFF
--- a/src/flyte/_code_bundle/bundle.py
+++ b/src/flyte/_code_bundle/bundle.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import asyncio
+import contextlib
 import gzip
 import hashlib
 import logging
@@ -8,10 +9,16 @@ import os
 import pathlib
 import random
 import sqlite3
+import sys
 import tempfile
 import time
 from pathlib import Path
-from typing import TYPE_CHECKING, ClassVar, Type
+from typing import TYPE_CHECKING, AsyncIterator, ClassVar, Type
+
+try:
+    import fcntl
+except ImportError:  # pragma: no cover - Windows local dev; task containers are POSIX
+    fcntl = None  # type: ignore[assignment]
 
 from async_lru import alru_cache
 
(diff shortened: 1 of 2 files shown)


pingsutw (approved):

pingsutw:
Thanks for fixing it

PR #1375, opened 2026-08-05
chore(deps): bump flyteidl2 to 2.0.37 for the settings desc renumber

afbrock:
## What

Bumps `flyteidl2` `2.0.31` → `2.0.37` across all four pinned sites (`pyproject.toml`, `rs_controller/pyproject.toml`, `rs_controller/Cargo.toml`, `plugins/ray/pyproject.toml`) and regenerates the four lockfiles.

## Why

[flyteorg/flyte#7776](https://github.com/flyteorg/flyte/pull/7776) moved the `desc` field option in `flyteidl2/settings/settings_definition.proto` off extension number **50000** to **1364**, from the block the protobuf global extension registry assigns to Flyte.

50000 sits in the range protobuf reserves for use within a single organization — only safe for schemas that never share a descriptor pool with third-party protos. `flyteidl2` is published to buf.build, PyPI and npm, and `desc` collided there with another published extension claiming the same tag. That collision fails extension registration **at import time** for anyone importing both schemas.

The SDK reads this annotation in `_resolve_description` (`src/flyte/remote/_settings.py`) to render the per-setting `##` documentation lines in `flyte get settings` output.

## No source change needed

`_resolve_description` goes through the generated `settings_definition_pb2.desc` symbol, never the literal number, and the annotation lives in `FieldOptions` so it is never serialized into an RPC payload. The renumber is descriptor-level only — the pin is the whole change.

The existing `test_every_leaf_has_a_description` is the regression guard: it fails loudly if the annotation ever stops resolving.

## Verification

- `desc.number` is `1364` on 2.0.37, and all 18 leaves still resolve a description (none blank).
- `pytest tests/flyte/remote/test_settings.py tests/flyte/cli/test_edit_settings.py` → 110 passed.
- Full suite → 4711 passed, 9 skipped. Two failures (`test_retrieve_degrades_when_keyring_not_installed`, `test_real_build`) are local-environment ones that reproduce identically on 2.0.31.
- `cargo check` in `rs_controller/` clean against the 2.0.37 crate.
- The `check-flyteidl2-versions` lint condition holds across all four pins.

The jump spans six releases, so it also picks up unrelated proto changes to `app`, `artifact`, `core/literals`, `core/tasks`, `task/common` and `workflow/run_definition`.

### DIFF
--- a/plugins/ray/pyproject.toml
+++ b/plugins/ray/pyproject.toml
@@ -8,7 +8,7 @@ requires-python = ">=3.10,<3.13"
 dependencies = [
     "ray[default]",
     "flyte",
-    "flyteidl2==2.0.31",
+    "flyteidl2==2.0.37",
 ]
 
 [dependency-groups]
--- a/plugins/ray/uv.lock
+++ b/plugins/ray/uv.lock
@@ -536,7 +536,7 @@ requires-dist = [
     { name = "docstring-parser", specifier = ">=0.16" },
     { name = "fastapi", marker = "extra == 'examples-test'" },
     { name = "flyte-controller-base", marker = "extra == 'rust-controller'", directory = "../../rs_controller" },
-    { name = "flyteidl2", specifier = "==2.0.31" },
+    { name = "flyteidl2", specifier = "==2.0.37" },
     { name = "fsspec", specifier = ">=2025.3.0" },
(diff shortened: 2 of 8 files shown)


pingsutw:
Closed it since we already bumped it on main

PR #1378, opened 2026-08-05
test(functional): add an installable environment-validation smoke suite

mhotan:
## What

A self-contained, **installable** pytest suite (`flyte-functional-tests`, under `tests/functional/`) that runs the core flyte v2 scenarios **end-to-end against a real backend** to prove a deployment is healthy: it can run tasks, return I/O and logs, build and cache images, run reusable actors, serve apps, and register triggers.

It's packaged as a real distribution so another project's CI consumes it by *installing* it and running `pytest --pyargs flyte_functional_tests` — no vendoring or path-hacking:

- `flyte_functional_tests/` package with a `pyproject` (version + build backend)
- `--skip` / markers / lightest-first ordering shipped as a **`pytest11` plugin**, so they survive `pytest --pyargs`
- extras: `[app]` = fastapi/httpx (the app scenario); `[harness]` = flyteplugins-union (only if you also drive cluster pool/queue/routing ops from the same env)

## Why

The same smoke scenarios currently live copied across several repositories and have drifted apart. This lands one consolidated, CI-agnostic, installable home so those copies can be replaced by a dependency on this suite.

## Scenarios

| Test | Checks |
| --- | --- |
| `test_simple` | Submit the basic task and wait for a successful terminal state |
| `test_image_builder` | Build a custom image (requests) and run a task on it |
| `test_image_cache` | Run the same stable-image task twice; the second hits the image cache |
| `test_io` | A completed run's outputs are retrievable |
| `test_logs` | The run's logs are retrievable through the SDK (best-effort) |
| `test_trigger` | Deploy a scheduled trigger, verify its automation spec, toggle it |
| `test_reusable` | Fan out work over a `ReusePolicy` (actor) environment |
| `test_app` | Deploy a FastAPI app, hit `/` and `/health`, then tear it down |

Every scenario is **self-contained** — it submits its own run and waits — so there is no shared fixture coupling and the suite parallelises cleanly.

## How to run

Marked `integration`, so the unit run (`-k "not integration"`) skips them. Point at a backend via your standard flyte config, or explicitly with `FLYTE_FUNCTIONAL_*` env (endpoint / api-key / org / project):

```bash
# In this repo (dev):
uv run pytest tests/functional/flyte_functional_tests -m integration -v

# From another repo: install the package (editable, from a checkout under the working dir)
# and collect it with --pyargs. See tests/functional/README.md -> "Consuming from another repo".
uv pip install --prerelease=allow -e "tests/functional[app,harness]"
cd tests/functional && pytest --pyargs flyte_functional_tests -m integration -v
```

Independent scenarios parallelise with **pytest-xdist** (`-n auto`; cap on a small/single-node backend so concurrent image builds don't overwhelm the builder). `--skip <scenario>` skips any a backend doesn't support (by marker, test name, or short name — e.g. `--skip app,trigger,reusable` on an OSS backend).

## App-serving robustness

`test_app` exercises `flyte.serve()`, which has a couple of rough edges this suite defends against so the scenario is reliable across runs:

- `serve()` blocks in `watch(wait_for="activated")` with no timeout, so a never-ready revision hangs silently at "Deployed App". Bounded (360s) with a server-side deployment-status dump on stall.
- The app registration persists across runs; a merely-*deactivated* (or serve-timeout-orphaned) app makes the next `serve()` re-activate a "pending deletion" registration that never reaches `is_active`. The suite purges the app (stop -> retry-delete) **before serve, on the serve-timeout path, and at teardown**, so every run starts and ends from a clean slate.

## Validation

Consumed by unionai/helm-charts#521 and validated end-to-end on its full integration matrix ([Integration Checks run](https://github.com/unionai/helm-charts/actions/runs/31756985710)) — k3d, managed dataplane (aws/gcp/azure), and self-hosted (aws/gcp) legs all green (self-hosted skips the app scenario, which its control plane doesn't serve).

## Notes

- `flyte` + `flyteplugins-union` are temporarily pinned to a known-good pair; unpin once the consuming CI tracks a current SDK line. `flyte` 2.x is a pre-release on PyPI, so installs need `--prerelease=allow`.
- A fully-decoupled, no-checkout consumption mode (install by version + `FLYTE_FUNCTIONAL_SUITE_SPEC`, so each scenario image installs the suite) is implemented but **experimental** — blocked on two SDK-core gaps: a run with no in-tree source raises `CodeBundleError`, and the remote image builder can't `pip install` a git+pre-release spec. Until those land, use editable-from-checkout.


tests/functional/flyte_ops.py
```diff
@@ -0,0 +1,240 @@
...
+            domain=domain,
+            image_builder="remote",
+        )
+
+
+def _phase_name(run) -> str:  # type: ignore[no-untyped-def]
+    return str(run.phase).rsplit(".", 1)[-1].lower()
+
+
+_ASSERT_TIMEOUT = 600  # seconds — bound each wait so a stuck run fails, not hangs
```
mhotan: make configurable with this default

tests/functional/flyte_ops.py
```diff
@@ -0,0 +1,240 @@
...
+            try:
+                act = run.pb2.action
+                if act.HasField("error_info"):
+                    detail = f": {act.error_info.kind}: {act.error_info.message}"
+            except Exception:
+                pass
+        raise RuntimeError(f"{label}: run {run.name} ended in phase={run.phase}{detail}")
+
+
+_SUBMIT_MAX_ATTEMPTS = 40
```
mhotan: make configurable with this default

tests/functional/flyte_ops.py
```diff
@@ -0,0 +1,240 @@
...
+_SUBMIT_RETRY_DELAY = 30
```
mhotan: make configurable with this default

tests/functional/flyte_ops.py
```diff
@@ -0,0 +1,240 @@
...
+# Submission-time errors that are transient backend/propagation states, not real
+# failures — retry them. Two families:
+#   * routing/capability propagation on a fresh data plane ("no clusters found",
+#     'cluster "<name>" not found' — the enabled-clusters cache is empty or lagging)
+#   * a briefly-unhealthy data plane: node consolidation/churn can evict the
+#     singleton operator/proxy (the DP<->CP tunnel) between the health gate and
+#     submission, so the CP reports the pool unhealthy for ~30-60s until the pod
+#     reschedules ("unhealthy", "could not select a cluster", "failed to get
+#     proxy"). Riding these out avoids false negatives from infra churn.
+_TRANSIENT_SUBMIT_MARKERS = (
```
mhotan: This list actually differs depending on flyte OSS vs Union. We probably should make this configurable or a switch statement between the two.

tests/functional/flyte_ops.py
mhotan: Done — the wait ceiling is now `FLYTE_FUNCTIONAL_WAIT_TIMEOUT` (default 600) via `_int_env`.

tests/functional/flyte_ops.py
mhotan: Done — `FLYTE_FUNCTIONAL_SUBMIT_ATTEMPTS` (default 40).

tests/functional/flyte_ops.py
mhotan: Done — `FLYTE_FUNCTIONAL_SUBMIT_RETRY_DELAY` (default 30).

tests/functional/flyte_ops.py
mhotan: Done — added a backend switch: `FLYTE_FUNCTIONAL_BACKEND=union|oss` (default union) selects a common base + per-backend transient markers (Union keeps the tunnel/proxy/cluster-pool set; OSS uses common-only). Extra substrings can be appended via `FLYTE_FUNCTIONAL_TRANSIENT_MARKERS`. See `_transient_markers`.

mhotan:
artifacts was released recently, we should add tests for that.

PR #1395, opened 2026-08-07
Add ArtifactLineageAppEnvironment example: track artifact producer/consumer lineage

cosmicBboy:
## Summary

- Adds `ArtifactLineageAppEnvironment` (`examples/artifacts/lineage/`), an example `AppEnvironment` construct that renders, for any published artifact, the full lineage graph around it: every producing run traced back to the original run/artifact in the chain, and every downstream run/artifact/app that consumed it (directly or transitively). List view at `GET /lineage`, per-artifact graph view at `GET /lineage/artifact/{name}`. Modeled on the `RunLineageDashboard` pattern from [unionai-oss/experimental-flyte-plugins](https://github.com/unionai-oss/experimental-flyte-plugins/tree/main/sensor) (label-driven, no stored lineage state — everything is re-derived from the backend on each request).
- Lineage is derived from two signals:
  - **Automatic**: artifacts bound as typed run inputs (manual `flyte.run(task, model=artifact)` or `OnArtifact`-triggered runs via `flyte.TriggeredArtifact`) already stamp `Literal.artifact_id` — no extra code needed.
  - **Fallback**: a private `__upstream_artifact__` label (set to the artifact's `tracker` string) for cases the backend can't see on its own — a task that resolves an artifact's value without binding it as a typed input, or an app that consumes an artifact at all. Runs pick this up via the existing `flyte.with_runcontext(labels=...)`; apps needed a new `labels` field (see below).
- `examples/artifacts/artifact_lineage_example.py`: an end-to-end runnable script exercising both signals — a multi-hop task chain (`produce_raw_data -> train_model`, automatic detection), a task that only takes a plain string input (`audit_model`, label fallback), and an app that "serves" the trained model (`model_server`, label fallback via the new `AppEnvironment.labels`).

## SDK change

- `flyte.app.AppEnvironment` gains a `labels: Mapping[str, str] | None` field, wired into `translate_app_env_to_idl` (`app_definition_pb2.Meta(labels=...)`). This mirrors the `labels` already accepted by `flyte.with_runcontext()` for runs — apps had no declarative way to carry labels before this (only a post-hoc `flyte.remote.App.replace(..., labels=...)`), which is what the lineage dashboard needs to detect artifact-consuming apps.

## Test plan

- [x] `uv run pytest tests/flyte/app/ tests/user_api_apps/ -q` — 387 passed (labels field doesn't break existing app serde/dataclass tests)
- [x] `ruff check` / `ruff format --check` clean on all new/changed files
- [x] Unit-level smoke test of `build_artifact_lineage()`'s graph-walking logic against mocked `Run`/`Artifact`/`App` objects — verifies producer chaining, label-based consumer detection, and downstream artifact chaining all produce the correct nodes/edges
- [x] Verified `translate_app_env_to_idl` round-trips `AppEnvironment(labels=...)` into `App.metadata.labels` on the proto
- [x] Imported `artifact_lineage_example.py` and confirmed the `ArtifactLineageAppEnvironment`'s FastAPI routes (`/lineage`, `/lineage/graph/{name}`, `/lineage/artifact/{name}`) wire up correctly
- [ ] Not tested against a live Flyte backend (no cluster available in this environment) — the example's `__main__` deploy/run/serve flow has not been exercised end-to-end

### DIFF
--- a/examples/artifacts/artifact_lineage_example.py
+++ b/examples/artifacts/artifact_lineage_example.py
@@ -1,0 +1,128 @@
+"""Artifact lineage: track producers and consumers of artifacts across runs and apps.
+
+`ArtifactLineageAppEnvironment` (see `lineage/`) renders, for any published
+artifact, the chain of everything connected to it: every producing run back to
+the original run/artifact in the chain, and every consuming run, artifact, or
+app downstream of it. Two signals feed the graph:
+
+1. **Automatic** — when an artifact is bound as a typed task input (a plain
+   `flyte.run(task, model=artifact)` call, or an `OnArtifact` trigger binding
+   `flyte.TriggeredArtifact`), the platform stamps the artifact's identity on
+   the input literal itself. No extra code needed; see `train_model` below,
+   which consumes `raw-data` this way.
+2. **Fallback label** — a task that resolves an artifact's value without
+   binding it as a typed input (e.g. it only takes a plain string), or an app
+   that reads an artifact at startup, leaves nothing for the platform to see.
+   For those, the caller stamps a private `__upstream_artifact__` label —
+   `flyte.with_runcontext(labels=...)` for a run, `AppEnvironment(labels=...)`
+   for an app — that the dashboard scans for. `audit_model` and
+   `model_server` below both need this.
+
+Try it:
+
+    python examples/artifacts/artifact_lineage_example.py
+
+then open the printed dashboard URL, or `<endpoint>/lineage` if you deploy it
+separately.
+"""
+
+import tempfile
+
+from lineage import LABEL_UPSTREAM_ARTIFACT, ArtifactLineageAppEnvironment
+
+import flyte
+import flyte.artifacts as artifacts
+from flyte.app import AppEnvironment
+from flyte.io import File
+
+env = flyte.TaskEnvironment(
+    name="artifact_lineage_example",
+    image=flyte.Image.from_debian_base().with_pip_packages("fastapi", "uvicorn"),
+)
+
+
+@env.task(produces_artifacts=True)
+async def produce_raw_data() -> File:
+    with tempfile.NamedTemporaryFile("w", suffix=".csv", delete=False) as f:
+        f.write("feature,label\n1,0\n2,1\n")
+    file = await File.from_local(f.name)
+    return artifacts.new(file, artifacts.Metadata(name="raw-data", description="Raw ingested training data"))
+
+
+# Automatic detection: `data` binds the artifact as a typed `File` input, so the
+# platform stamps `Literal.artifact_id` on it — the dashboard finds this edge by
+# scanning `train_model`'s runs for artifact-bound inputs, no label required.
+@env.task(produces_artifacts=True)
+async def train_model(data: File) -> File:
+    async with data.open("rb") as fh:
+        rows = bytes(await fh.read()).decode().splitlines()
+    with tempfile.NamedTemporaryFile("w", suffix=".bin", delete=False) as f:
+        f.write(f"weights trained on {len(rows) - 1} rows")
+    weights = await File.from_local(f.name)
+    return artifacts.new(weights, artifacts.Metadata(name="trained-model", description="Trained model weights"))
+
+
+# Fallback label: `model_ref` is a plain string, not a bound `Artifact`/`File` — the
+# platform has no way to know this run consumed anything. The caller (see `__main__`)
+# stamps the `__upstream_artifact__` label explicitly so the dashboard can still find it.
+@env.task
+async def audit_model(model_ref: str) -> str:
+    return f"audited {model_ref}"
+
+
+# Apps can't bind artifacts as typed inputs at all, so the label is the *only* way to
+# declare "this app serves artifact X". The label value is only known once the model
+# exists, so it's attached via `clone_with(labels=...)` + redeploy in `__main__`,
+# not at module-definition time.
+model_server = AppEnvironment(
+    name="artifact-lineage-model-server",
+    image=env.image,
+    command="python -m http.server 8080",
+    requires_auth=False,
+)
+
+lineage_dashboard = ArtifactLineageAppEnvironment(
+    name="artifact-lineage-dashboard",
+    image=flyte.Image.from_debian_base().with_pip_packages("fastapi", "uvicorn"),
+    include=("lineage/*.py",),
+    requires_auth=False,
+    # `train_model` is discovered via the automatic bound-input scan; `audit_model` is
+    # discovered via the label (listed here too so the bound-input scan also covers it).
+    watched_tasks=["artifact_lineage_example.train_model", "artifact_lineage_example.audit_model"],
+    watched_apps=[model_server.name],
+)
+
+
+if __name__ == "__main__":
+    flyte.init_from_config()
+
+    from flyte.remote import Artifact
+
+    data_run = flyte.run(produce_raw_data)
+    print(data_run.url)
+    data_run.wait()
+    data = Artifact.get("raw-data")
+
+    model_run = flyte.run(train_model, data=data)
+    print(model_run.url)
+    model_run.wait()
+    model = Artifact.get("trained-model")
+
+    # Fallback path #1 (run): stamp the label before launching, since `audit_model`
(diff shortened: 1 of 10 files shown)


cosmicBboy:
Deployed \`ArtifactLineageAppEnvironment\` (the \`artifact-lineage-dashboard\` app) live against the playground cluster (\`.flyte/config-playground.yaml\`) to sanity-check it beyond unit tests. It came up ACTIVE with no crash-loop, confirming the code bundle, FastAPI routes, and `Artifact`/`Run`/`App` remote calls all wire together correctly outside a mocked environment.

Two real bugs surfaced and are fixed in the latest commit:
- `include=("lineage/*.py",)` resolved against the wrong directory (`Environment._get_declaring_file()` anchors relative `include` paths to wherever the concrete subclass's own `__post_init__` first runs, not the file that calls the constructor) — turned out to be unnecessary anyway, since the `lineage` package is eagerly imported at module scope and already covered by the default `loaded_modules` bundling.
- `requires_auth=False` is rejected outright by orgs that disallow anonymous apps (`app.disallow_anonymous`) — dropped in favor of the default `requires_auth=True`.

Not yet exercised: the full `__main__` demo flow (producing artifacts, running the task chain, deploying the labeled consumer app) — only the dashboard app itself was deployed, so `/lineage` will show an empty artifact list until that flow is run against the same project/domain.

cosmicBboy:
Expanded scope, all verified against the live playground cluster:

**Three new example pipelines** under \`examples/artifacts/lineage_examples/\`, each exercising both lineage-detection signals with a different graph shape:
- \`ml_pipeline.py\` — dataset -> train -> evaluate (merge point, two upstream artifacts feed one run) -> FastAPI model-serving app. Small closed-form linear regression, no ML framework.
- \`etl_pipeline.py\` — two independent sources (diamond shape) -> transform -> join -> load -> FastAPI monitor app.
- \`bi_pipeline.py\` — ingest -> daily/weekly aggregates -> report (fan-in: one run consumes artifacts at two different chain depths) -> FastAPI report app.
- \`dashboard.py\` — one consolidated \`ArtifactLineageAppEnvironment\` watching all of the above plus the original \`artifact_lineage_example.py\`.

**Two real bugs** surfaced only by actually running these against a live cluster (not visible from unit tests against mocks):
1. Kubernetes label values reject \`/\` and \`@\` (which \`artifact.tracker\` contains), and label *keys* can't start/end with \`_\`. Both silently killed the run/app with no useful client-side error. Replaced the single \`__upstream_artifact__\` label with two plain keys, \`upstream-artifact-name\`/\`upstream-artifact-version\`.
2. \`AppEnvironment.clone_with(labels=...)\` called inside a function returns a local-variable copy the app resolver can't find by name in the module's global namespace, so the deployed container doesn't know what to serve (deployment "succeeds" but the app never activates). Fixed by mutating \`.labels\` in place on the original module-level object instead.

**Dashboard UX redesign** (the other ask from this thread): \`examples/artifacts/lineage/dashboard.py\` is now a single-page app — a searchable sidebar lists every artifact, and clicking one renders its lineage graph inline via a client-side fetch, replacing the old two-page list-then-navigate flow.

**Live verification**: ran all four pipelines and deployed all five apps (4 consumers + the dashboard) against \`.flyte/config-playground.yaml\`, using an image built from \`make dist\` so the container picked up the \`AppEnvironment.labels\` SDK change. All reached \`ACTIVE\`. Spot-checked \`build_artifact_lineage()\` directly against the live artifacts/runs — confirmed a genuine multi-parent merge node (ml-eval-report's two upstream artifacts) and a label-only app-consumer edge (ml-model -> ml-model-server), both wired correctly.

cosmicBboy:
Fixed the dashboard failing to load (\"Dashboard failed to load (CDN unreachable?)...\") — reported after the last update.

Root cause: the single-page redesign dropped the \`<script type="importmap">\` block entirely, so every dynamic \`import("react")\` etc. threw \`Failed to resolve module specifier\`, caught by the top-level try/catch and replaced with the generic fallback message. A second, independently fatal bug was hiding behind it: the \`Legend\` component passed \`style="background:var(--artifact)"\` as a plain string, which React rejects outright (style props must be an object) — fixed by moving those to static CSS classes.

Reproduced and verified the fix with headless Chromium (playwright) rather than just re-reading the code:
- Confirmed the exact failure locally against the actual rendered HTML (both bugs, one after the other, each with the real browser error).
- After the fix, ran the full interactive flow against a mocked backend: sidebar populates with artifacts, clicking one renders its lineage graph inline (artifact node → producing run → consuming app, wired with the correct edges), zero console errors.
- Redeployed to the live playground dashboard app, which reached \`ACTIVE\`.

Also cleaned up a couple of React dev-mode warnings (missing \`key\` props on sibling ternary branches) surfaced by the same repro.

cosmicBboy:
Fixed both issues:

**Malformed node links** — \`Run\`/\`Artifact\`/\`App\`\`.url\` properties are already fully-qualified absolute URLs (confirmed against \`ControlPlaneClient._resource_url\`: \`f"{scheme}://{domain}/v2/..."\`). The dashboard was prepending a derived console base on top of that, producing e.g. \`https://playground.canary.unionai.cloudhttps://playground.canary.unionai.cloud/v2/.../runs/...\`. Removed that base-prepending entirely — node clicks and the "Open in console" link now use \`.url\` directly.

**Slow loading** — \`build_artifact_lineage()\` walked the *entire* upstream chain and *entire* downstream tree unconditionally; the downstream side especially can cost several RPCs per hop (a label-filtered run scan plus, per watched task, a full recent-run scan with an \`input_literals\` fetch per run). Replaced the flat \`max_depth\` safety cap with \`upstream_depth\`/\`downstream_depth\` (both default 1 — one step in each direction), and added \`has_more_upstream\`/\`has_more_downstream\` to the graph response. The dashboard now shows "← Load more upstream" / "Load more downstream →" buttons on the canvas that bump the depth and refetch, so the graph grows incrementally instead of paying for the whole lineage up front.

Verified with headless Chromium against a mocked backend: node clicks and the console link resolve to the correct URL, and clicking "Load more upstream" grows the graph (5 cards → adds 2 more) and the button correctly flips to disabled "Origin reached" once there's nothing left upstream. Also spot-checked directly against the live playground cluster — depth=1 now returns 4 nodes for a real artifact vs. 6 with the old unbounded walk, with accurate \`has_more\` flags. Redeployed; the dashboard app is ACTIVE.

cosmicBboy:
## Why the lineage graph was loading slowly

Traced this with real RPC-level profiling against the live playground cluster rather than guessing. \`GET /lineage/graph/{name}\` does two walks from the selected artifact:

- **Upstream** (cheap): the artifact already carries its producing run's identity as a protobuf field, so this is just a \`Run.get\` + inspecting that run's inputs.
- **Downstream** (expensive): there's no "who consumed me" pointer on an artifact, so it has to *search* — for every task name in \`watched_tasks\`, list that task's recent runs and, for any not already found via the label shortcut, fetch its full details to check whether it consumed this artifact. Same idea for \`watched_apps\`.

Profiled a depth=1 request for \`ml-eval-report\` against the consolidated dashboard's watch list (10 tasks, 4 apps):

\`\`\`
TOTAL WALL CLOCK: 4.68s
Total RPCs: 34
  app_service.get:        4 calls, sum=0.19s
  artifact_service.get_artifact: 8 calls, sum=0.37s
  run_service.get_run_details:  11 calls, sum=0.51s
  run_service.list_runs:  11 calls, sum=0.76s
\`\`\`

Two things going on:
1. It scans *every* watched task, not just the ones relevant to this particular artifact (only 2 of the 10 actually mattered here) — inherent to how discovery works without a stored index, not something I fixed here.
2. The scan loops (\`_consumer_runs_of_artifact\`, \`_consumer_apps_of_artifact\`) were plain sequential \`for\` loops — every task, and every unmatched run within each task, awaited one at a time despite being fully independent of each other. **This is what I fixed**: restructured both to fan out with \`asyncio.gather\` (bounded by a small semaphore so a dashboard watching many tasks doesn't fire an unbounded RPC burst).

Re-profiled after the change: same artifact, same 34 RPCs, **1.61s** (down from 4.68s) — sum-of-latencies became max-of-latencies. Verified the lineage smoke test (graph shape, depth-truncation flags) still passes unchanged, and redeployed — the dashboard app is ACTIVE.

PR #1397, opened 2026-08-08
Flush a leaf task's report at task end

pingsutw:
## What

`Report.has_content()`, and the end-of-task flush guard in `taskrunner.convert_and_run` now
reads the report off `tctx` instead of a stale `ctx`.

```diff
 if task.report:
-    if ctx.get_report():
+    if tctx.report is not None and tctx.report.has_content():
         await flyte.report.flush.aio()
```

## Why

The guard added in #344 is wrong twice over:

1. `ctx` is bound at the top of `convert_and_run`, *before*
   `with ctx.replace_task_context(tctx)`. `replace_task_context` returns a **new** `Context`;
   the local `ctx` still points at the old one, so `ctx.get_report()` reads the *parent's*
   task context — `None` for a leaf task.
2. Even with the right context, `get_report()` returns a `Report` dataclass, which is truthy
   whether or not anything was logged to it. The "has content" semantics the comment
   describes were never actually implemented.

Net effect: the final flush is skipped for every leaf task. A task that logs after its last
explicit `flyte.report.flush()` loses that content, and a task that never flushes explicitly
produces **no report at all**.

Every existing example under `examples/` calls `flush()` explicitly, which is why this went
unnoticed.

#344's actual goal is preserved: it was fixing the elastic plugin's main process clobbering
valid worker reports with an empty one. `elastic_launch` spawns worker *processes*, so the
main process's own report has no content, `has_content()` is `False`, and the flush is still
skipped. `test_empty_report_is_not_flushed` locks that in so it can't silently regress again.

## Reproduction

Plain v2, no plugins. Save as `repro.py` and run it — it is deliberately not added to
`examples/`, since every task there flushes explicitly on purpose.

```python
import flyte
import flyte.remote
import flyte.report

env = flyte.TaskEnvironment("report_without_explicit_flush")

NEVER_FLUSHED = "<p>logged, never flushed</p>"
BEFORE_FLUSH = "<p>logged before the explicit flush</p>"
AFTER_FLUSH = "<p>logged after the last explicit flush</p>"


@env.task(report=True)
async def never_flushes() -> None:
    """The whole report depends on the end-of-task flush."""
    flyte.report.get_tab("main").log(NEVER_FLUSHED)


@env.task(report=True)
async def logs_after_last_flush() -> None:
    """Only the trailing content depends on the end-of-task flush."""
    flyte.report.get_tab("main").log(BEFORE_FLUSH)
    await flyte.report.flush.aio()
    flyte.report.get_tab("main").log(AFTER_FLUSH)


@env.task
async def main() -> None:
    await never_flushes()
    await logs_after_last_flush()


async def check(run: flyte.remote.Run) -> None:
    expectations = {
        "never_flushes": [NEVER_FLUSHED],
        "logs_after_last_flush": [BEFORE_FLUSH, AFTER_FLUSH],
    }
    failures = []
    async for action in flyte.remote.Action.listall.aio(for_run_name=run.name):
        # task_name is "<env>.<task>"; the action's own name is an opaque id.
        short_name = (action.task_name or "").split(".")[-1]
        expected = expectations.get(short_name)
        if expected is None:
            continue
        try:
            report = await action.get_report.aio()
        except Exception as e:
            # No flush means no report.html at all, so the signed-url request 404s
            # rather than returning empty HTML.
            failures.append(f"{short_name}: no report artifact was written ({type(e).__name__})")
            continue
        for fragment in expected:
            if fragment not in report:
                failures.append(f"{short_name}: missing {fragment!r}")
    if failures:
        raise AssertionError("report content dropped:\n  " + "\n  ".join(failures))
    print("OK: every logged fragment reached the report")


if __name__ == "__main__":
    import asyncio

    flyte.init_from_config()
    run = flyte.run(main)
    print(run.url)
    run.wait()
    asyncio.run(check(run))
```

Against a cluster whose flyte predates this fix:

```
$ python repro.py
...
AssertionError: report content dropped:
  never_flushes: no report artifact was written (ConnectError)
  logs_after_last_flush: missing '<p>logged after the last explicit flush</p>'
```

`never_flushes` writes no `report.html` at all, so fetching it 404s — the report is not
merely empty, it does not exist. Or drop the `check()` call and just open the printed run URL
and click Reports on each task.

Two caveats:

- **Remote only.** Local execution goes through `Run._run_local` and never reaches
  `convert_and_run`.
- The task container runs whatever flyte release is in its image, so this script cannot show
  the *fixed* behaviour until a release carrying this PR is out. The cluster-free proof is
  `test_leaf_task_report_is_flushed_at_task_end`, which drives the real `convert_and_run` and
  fails on `main`.

## Test

`make unit_test` — 4812 passed. Three failures in
`test_image_build_engine.py::test_build_hard_errors_without_push_registry`,
`test_keyring_store.py::test_retrieve_degrades_when_keyring_not_installed` and
`test_image.py::test_install_flyte_false_builds_full_image` are pre-existing on `main` —
they fail identically with this change stashed, and pass on CI.

New coverage:
- `tests/flyte/internal/runtime/test_taskrunner.py` — drives the real `convert_and_run`:
  a leaf task that logs is flushed (fails without this change), an empty report is not,
  and `report=False` is not.
- `tests/flyte/report/test_report.py` — `has_content()` across a fresh report, a created-
  but-unlogged tab, the main tab, and a named-only tab.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,7 +24,11 @@ dependencies = [
     "msgpack>=1.1.0",
     "toml>=0.10.2",
     "async-lru>=2.0.5",
-    "mashumaro",
+    # 3.15 is the first release with mashumaro.jsonschema.plugins, imported by
+    # flyte/types/_type_engine.py. Without the floor, installing into an environment that
+    # already has an older mashumaro (e.g. a flytekit v1 image) leaves it in place and
+    # `import flyte` dies with ModuleNotFoundError.
+    "mashumaro>=3.15",
     "aiolimiter>=1.2.1",
     "flyteidl2==2.0.37",
     "packaging",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,10 +24,6 @@ dependencies = [
     "msgpack>=1.1.0",
     "toml>=0.10.2",
     "async-lru>=2.0.5",
-    # 3.15 is the first release with mashumaro.jsonschema.plugins, imported by
-    # flyte/types/_type_engine.py. Without the floor, installing into an environment that
-    # already has an older mashumaro (e.g. a flytekit v1 image) leaves it in place and
-    # `import flyte` dies with ModuleNotFoundError.
     "mashumaro>=3.15",
     "aiolimiter>=1.2.1",
     "flyteidl2==2.0.37",
--- a/src/flyte/_internal/runtime/taskrunner.py
+++ b/src/flyte/_internal/runtime/taskrunner.py
@@ -210,7 +210,7 @@ async def convert_and_run(
         if task.report:
             # Check if report has content before flushing to avoid overwriting
             # worker reports (from Elastic/distributed tasks) with empty main process report.
-            if ctx.report is not None and ctx.report.has_content():
+            if ctx.get_report() is not None and ctx.get_report().has_content():
                 await flyte.report.flush.aio()
 
         sw = Stopwatch("convert_outputs_from_native")
--- a/src/flyte/_internal/runtime/taskrunner.py
+++ b/src/flyte/_internal/runtime/taskrunner.py
@@ -194,7 +194,7 @@ async def convert_and_run(
         tctx_kwargs["run_start_time"] = run_start_time
     tctx = TaskContext(**tctx_kwargs)
 
-    with ctx.replace_task_context(tctx):
+    with ctx.replace_task_context(tctx) as ctx:
         sw = Stopwatch("convert_inputs_to_native")
         sw.start()
         inputs_kwargs = await convert_inputs_to_native(inputs, task.native_interface)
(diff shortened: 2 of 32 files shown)


pyproject.toml
```diff
@@ -24,7 +24,7 @@ dependencies = [
     "msgpack>=1.1.0",
     "toml>=0.10.2",
     "async-lru>=2.0.5",
-    "mashumaro",
+    "mashumaro>=3.15",
```
pingsutw, line 27: 3.15 is the first release with mashumaro.jsonschema.plugins, imported by flyte/types/_type_engine.py. Without the floor, installing into an environment that already has an older mashumaro (e.g. a flytekit v1 image) leaves it in place and `import flyte` dies with ModuleNotFoundError.

src/flyte/_internal/runtime/taskrunner.py
```diff
@@ -209,8 +209,8 @@ async def convert_and_run(
             return None, convert_from_native_to_error(err)
         if task.report:
             # Check if report has content before flushing to avoid overwriting
-            # worker reports (from Elastic/distributed tasks) with empty main process report
-            if ctx.get_report():
+            # worker reports (from Elastic/distributed tasks) with empty main process report.
+            if tctx.report is not None and tctx.report.has_content():
```
kumare3: can you just override the variable in the contextmanager

EngHabu (dismissed):

kumare3 (approved):

PR #1411, opened 2026-08-11
fix: don't crash-report a non-Connect endpoint answering a Connect RPC (FLYTE-SDK-77, FLYTE-SDK-7A)

EngHabu:
## What

`flyte run` and `flyte deploy` from a Windows host failed with `RuntimeSystemError: Upload failed for C:\Users\...\fast….tar.gz`, whose cause was a bare `ConnectError: No Content` — raised out of both `CreateUploadLocation` (`_data.py`) and `SelectCluster` (`controlplane.py`). Both were reported to Sentry as SDK crashes.

A Connect endpoint answers a unary POST with **200 + a Connect body**, or with an **error status + a Connect JSON body**. HTTP **204 is neither**. A *success* status that isn't 200 means the request never reached a Connect handler at all — a proxy, VPN appliance, captive portal, corporate TLS interceptor or misrouted ingress absorbed it and answered on the backend's behalf. That's endpoint/network configuration, not a Python-SDK logic bug, and the SDK can't recover from it.

This is the same rationale already applied to `UNIMPLEMENTED` in `_USER_ACTIONABLE_CONNECT_CODES` ("a raw HTTP 404 from the ingress … the endpoint config is wrong").

## Why the check looks the way it does

connectrpc maps only a handful of HTTP statuses onto Connect codes (401/403/404/429/502/503/504). Everything else falls through to `ConnectWireError.from_http_status`, which produces `Code.UNKNOWN`, **no details**, and the stdlib reason phrase as the *entire* message. The numeric status is dropped on the floor, so matching the phrase is the only way back to the status class.

## Deliberately narrow — and why that matters

HTTP 500 produces an **identically shaped** error: `Code.UNKNOWN`, no details, message `"Internal Server Error"`. Filtering on "message is a bare HTTP phrase" would therefore have silenced real backend signal, including **FLYTE-SDK-64 (276 events)**, 6S and 43. Only the 2xx class is unambiguously "an intermediary answered instead of the backend", so only it is filtered.

I checked this against the live corpus before writing the fix — every `ConnectError` across all 67 unresolved issues, and the only ones carrying a *2xx* phrase are 77 and 78. Genuine backend errors all carry verbose Go-style messages (`rpc error: code = … desc = …`), and the backend-collapses-code-into-UNKNOWN family (67/4T) is untouched.

Verified behaviour:

| response | code | message | filtered |
|---|---|---|---|
| 204 / 202 | UNKNOWN | `No Content` / `Accepted` | ✅ yes |
| 500 | UNKNOWN | `Internal Server Error` | ❌ no (real signal) |
| 302 | UNKNOWN | `Found` | ❌ no |
| 404 / 401 / 502 | UNIMPLEMENTED / UNAUTHENTICATED / UNAVAILABLE | — | ❌ no (already handled) |
| backend UNKNOWN + verbose msg | UNKNOWN | `rpc error: code = Unavailable desc = …` | ❌ no |

## Tests

10 new tests. 7 fail without the `_sentry.py` change; the two **regression guards** (500 and backend-UNKNOWN still reach Sentry) pass both before and after — that's the point of them. Full file: 55 passed.

## Follow-up not included here

Filtering stops the false crash report but the user still sees `ConnectError: No Content`, which is opaque. A clearer message ("the endpoint returned HTTP 204 with no Connect payload — something between this machine and `<endpoint>` intercepted the request") would need a chokepoint at the RPC boundary; kept out of this PR to keep it reviewable.

---

## Update: the same root cause via connectrpc's other rejection path (FLYTE-SDK-7A)

Added in a follow-up commit on this branch. connectrpc has a second way of saying "this is not a Connect response" — when the body's content-type cannot be decoded by the codec:

```
ConnectError(Code.UNKNOWN, "invalid content-type: 'text/html'; expecting 'application/proto'")
```

An HTML page where a protobuf body belongs means the same thing a 204 does: a proxy, login page or misrouted ingress answered instead of the backend. `_client_config.py` already draws exactly this conclusion for this exact signature when it comes back from the auth metadata fetch (#1235) — [FLYTE-SDK-7A](https://unionai.sentry.io/issues/?query=FLYTE-SDK-7A) and [FLYTE-SDK-6P](https://unionai.sentry.io/issues/?query=FLYTE-SDK-6P) are the same thing arriving on a data-plane call.

Kept as narrow as the status-class check: only `text/*` matches, so an `application/*` mismatch — which would point at a codec bug on our side — keeps reporting.

**Corpus check.** Following the same discipline as the 2xx discriminator, the predicate was tested against `events/latest` for *every* unresolved issue in the project. Exactly two match — 7A and 6P — and no backend-owned issue does. Worth calling out explicitly: FLYTE-SDK-3A and FLYTE-SDK-4K embed an **nginx HTML error page inside the message** and do *not* match, because the filter keys on connectrpc's own message shape rather than on any mention of HTML. There is a regression test pinning that.

Note 6P is nominally tracked by #1340; its latest event has since drifted to this content-type shape, so this is what would actually stop it firing. Left the `fixes` line to #1340 rather than claiming it from two PRs.

fixes FLYTE-SDK-77
fixes FLYTE-SDK-78
fixes FLYTE-SDK-7A

### DIFF
--- a/src/flyte/_sentry.py
+++ b/src/flyte/_sentry.py
@@ -9,6 +9,7 @@ import atexit
 import errno
 import logging
 import os
+import re
 import socket
 from contextlib import contextmanager
 from http import HTTPStatus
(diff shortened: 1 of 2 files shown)


EngHabu (commented):
LGTM

AdilFayyaz (approved):

PR #1416, opened 2026-08-11
Add flyte.load_interactive_ctx() to restore task context in a debugger

EngHabu:
## What

Adds a public `flyte.load_interactive_ctx()` for restoring the task execution context when attached to a live debug-mode task pod (debugger, REPL, or notebook).

## How

**Write side** — `flyte/_interactive_run_context.py:write_interactive_run_context()` persists the a0 runtime parameters (inputs path, outputs path, raw data prefix, checkpoint/prev-checkpoint paths, run/action names, project/domain/org, version, run base dir, image cache, code bundle tgz/pkl/dest, resolver + args) to a well-known location: `<cwd>/.flyte/run_context.json`. It is called from `prepare_launch_json()` — the same place and moment `.vscode/launch.json` is written — so both the vscode and ssh debug flows get it, with the `{{...}}` run/action name templates already resolved.

**Read side** — `flyte.load_interactive_ctx(path=None)`:
- Looks for `<cwd>/.flyte/run_context.json` (or an explicit `path`); raises `FileNotFoundError` with a pointed message when absent, `ValueError` when required fields are missing.
- Calls `init_in_cluster(org, project, domain)` to initialize the SDK/storage against the cluster.
- Rebuilds `ActionID`, `RawDataPath` (including the `_F_PATH_REWRITE` accelerated-datasets check, mirroring the a0 entrypoint), `CheckpointPaths`, `CodeBundle`, and `ImageCache`, assembles a `TaskContext` (`mode="remote"`, `interactive_mode=True`), and installs it permanently on the root context var — no context-manager scoping, since a REPL/debugger has no natural exit point.
- Returns the `TaskContext`; from then on `flyte.ctx()`, File/Dir IO, and checkpoints behave as if inside the task.

## Notes

The config is only written in debug mode (vscode/ssh), matching the launch.json behavior. If we want `load_interactive_ctx()` to work when attaching to *any* task pod, the write call can move into `_run_action` in `runtime.py` so every a0 start drops it.

## Testing

`tests/flyte/test_interactive_run_context.py`: write-side content, missing-file and missing-field errors, full restore round-trip (asserting `flyte.ctx() is tctx`), env-var template resolution, and that `prepare_launch_json` drops both `launch.json` and `.flyte/run_context.json`. Existing debug tests unaffected; ruff clean.

## Example file:

```json
{
    "config_version": 1,
    "inputs": "s3://<bucket>/metadata/v2/demo/enghabu/development/i47DEQpj8HBSa-_TImW-/a0/inputs.pb",
    "outputs_path": "s3://<bucket>/metadata/v2/demo/enghabu/development/ur4kmd248cw7kcsckzk4/a0/0",
    "version": "4690294904163e827d0048d5d05350cb",
    "run_base_dir": "s3://<bucket>/metadata/v2/demo/enghabu/development/ur4kmd248cw7kcsckzk4",
    "raw_data_path": "s3://<bucket>/ba/demo/enghabu/development/ur4kmd248cw7kcsckzk4/a0/ur4kmd248cw7kcsckzk4-a0-0",
    "checkpoint_path": "s3://<bucket>/ba/demo/enghabu/development/ur4kmd248cw7kcsckzk4/a0/ur4kmd248cw7kcsckzk4-a0-0/_flytecheckpoints",
    "prev_checkpoint": "\"\"",
    "name": "a0",
    "run_name": "ur4kmd248cw7kcsckzk4",
    "run_start_time": "{{.runStartTime}}",
    "project": "enghabu",
    "domain": "development",
    "org": "demo",
    "image_cache": "H4sIAAAAAAAC/03OQZKDIBAF0Lv0OhIjgo6XsVoghiC0AUlSsbz7MFnN9tXv/3sH63E240Lk8grDDmnF6GAALqTkvJZNLXumXWRGRZZTZTBtVcPQ44cCvhJT5M85WApnbTwN3/tKCd1eOqFEW7edwAbr6087XbgRPe+wv8JxginbRY8xh9Hq9G96B4pz+eCvDk6wRrobtRUwYb7hlItp8mjDN/M0C63ehK1wQG8KZvm4LZty77tr/PPxke8+wXEcv4JuTHntAAAA",
    "tgz": "s3://<bucket>demo/enghabu/development/I2ICSSIECY7IE7IAJDK5AU2QZM======/fast31c598f01b6ecf7e7560d172c5c12110.tar.gz",
    "pkl": null,
    "dest": ".",
    "resolver": "flyte._internal.resolvers.default.DefaultTaskResolver",
    "resolver_args": [
        "mod",
        "examples.basics.debug_spark",
        "instance",
        "spark_notebook"
    ]
}
```

### DIFF
--- a/examples/plugins/spark_example.py
+++ b/examples/plugins/spark_example.py
@@ -1,4 +1,3 @@
-# # Spark Example
 import random
 from copy import deepcopy
 from operator import add
@@ -16,11 +15,13 @@ image = (
         extendable=True,
         platform=("linux/amd64", "linux/arm64"),
     )
-    .with_pip_packages("flyteplugins-spark>=2.5.0")
+    .with_pip_packages("flyteplugins-spark>=2.5.18")
 )
 
 task_env = flyte.TaskEnvironment(
-    name="get_pi", resources=flyte.Resources(cpu=(1, 2), memory=("400Mi", "1000Mi")), image=image
+    name="get_pi",
+    resources=flyte.Resources(cpu=(1, 2), memory=("400Mi", "1000Mi")),
+    image=image,
 )
 
 spark_conf = Spark(
@@ -29,7 +30,7 @@ spark_conf = Spark(
         "spark.executor.memory": "1000M",
         "spark.executor.cores": "1",
         "spark.executor.instances": "2",
-        "spark.driver.cores": "1",
+        "spark.driver.cores": "3",
         "spark.kubernetes.file.upload.path": "/opt/spark/work-dir",
         "spark.jars": "https://storage.googleapis.com/hadoop-lib/gcs/gcs-connector-hadoop3-latest.jar,https://repo1.maven.org/maven2/org/apache/hadoop/hadoop-aws/3.2.2/hadoop-aws-3.2.2.jar,https://repo1.maven.org/maven2/com/amazonaws/aws-java-sdk-bundle/1.12.262/aws-java-sdk-bundle-1.12.262.jar",
     },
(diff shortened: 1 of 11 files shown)


cosmicBboy (commented):
nit: can we rename the module and related functions to capture the fact that this context is for interactive debugging/sessions?

e.g. `_interactive_run_context.py`

And the public method can be `load_interactive_ctx` or something like it. `load_context` can be confused with the existing `ctx` method

pingsutw (approved):

PR #1421, opened 2026-08-12
fix: make flyte start devbox work on Windows (FLYTE-SDK-79, FLYTE-SDK-7B)

EngHabu:
## What

`flyte start devbox` carried three POSIX-only assumptions. On Windows each one crashes the CLI with an exception that reads like an SDK bug, and both Sentry issues below came from the same Windows user in the same session.

### 1. `/tmp` is not a path on Windows — [FLYTE-SDK-79](https://unionai.sentry.io/issues/?query=FLYTE-SDK-79)

`_KUBE_DIR = Path("/tmp/.kube")` resolves to the drive-relative `\tmp\.kube`, and `launch_devbox` fails at `mkdir`:

```
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: '\tmp\\.kube'
```

Derived from `tempfile.gettempdir()` instead. Side benefit on POSIX: each user gets their own directory on shared Linux hosts, where a single root-or-other-user-owned `/tmp/.kube` is exactly the situation the chown fallback below exists to work around.

### 2. `KUBECONFIG` is a path *list*

kubectl splits `KUBECONFIG` on the platform separator — `;` on Windows. Joining with a hardcoded `:` makes kubectl split `C:\Users\...` on its drive letter, so neither file loads and `kubectl config view --flatten` exits 1. Now uses `os.pathsep`.

### 3. `os.getuid` does not exist on Windows — [FLYTE-SDK-7B](https://unionai.sentry.io/issues/?query=FLYTE-SDK-7B)

Recovering from that kubectl failure reached straight for `os.getuid()`/`os.getgid()`, so the recovery attempt itself raised:

```
AttributeError: module 'os' has no attribute 'getuid'
```

...which is what got reported, burying the real kubectl error underneath it. The chown only means something where POSIX uids exist; elsewhere we now surface the original kubectl failure.

While here: the two `subprocess.CalledProcessError`s that could escape `_merge_kubeconfig` — from `docker exec ... chown` and from the retried flatten — are now `click.ClickException`s carrying kubectl's own stderr, so users get the reason instead of a traceback (and these stop reaching Sentry as SDK crashes).

## Testing

`tests/cli/test_devbox.py`: 26 passed. 5 of them fail on `main` (verified by stashing the source change):

- `test_kube_dir_is_under_the_platform_temp_dir`
- `test_kubeconfig_list_uses_platform_path_separator` (monkeypatches `os.pathsep` to `;` so it discriminates on any host, not only Windows)
- `test_missing_getuid_reports_original_kubectl_failure` (deletes `os.getuid` via monkeypatch)
- `test_second_flatten_failure_surfaces_as_click_exception`
- `test_chown_failure_surfaces_as_click_exception`

fixes FLYTE-SDK-79
fixes FLYTE-SDK-7B

### DIFF
--- a/src/flyte/cli/_devbox.py
+++ b/src/flyte/cli/_devbox.py
@@ -5,6 +5,7 @@ import json
 import os
 import shutil
 import subprocess
+import tempfile
 import time
 import urllib.error
 import urllib.request
@@ -22,9 +23,10 @@ from flyte import _sentry
 
 _CONTAINER_NAME = "flyte-devbox"
 _VOLUME_NAME = "flyte-devbox"
-_KUBE_DIR = Path(
-    "/tmp/.kube"
-)  # This path is used to store k3s kubeconfig file, we later merge it with the default kubeconfig
+# This path is used to store the k3s kubeconfig file, we later merge it with the default kubeconfig.
+# Use the platform temp directory rather than a hardcoded "/tmp": on Windows there is no /tmp, and
+# Path("/tmp/.kube") becomes the drive-relative "\tmp\.kube", which mkdir rejects outright.
+_KUBE_DIR = Path(tempfile.gettempdir()) / ".kube"
 _KUBECONFIG_PATH = _KUBE_DIR / "kubeconfig"
 _FLYTE_DEVBOX_CONFIG_DIR = Path.home() / ".flyte" / "devbox"
 _PORTS = ["6443:6443", "30000:30000", "30001:30001", "30002:30002", "30003:30003", "30080:30080", "30081:30081"]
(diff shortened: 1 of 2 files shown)


EngHabu (commented):
LGTM

EngHabu:
Rebased onto `main` (was CONFLICTING against #1449, `implements flyte get devbox`, which landed in `cli/_devbox.py`).

`src/flyte/cli/_devbox.py` **auto-merged cleanly** — all three platform fixes survived and the new `flyte get devbox` code shares the same constants, so it picks them up for free:
- `_KUBE_DIR = Path(tempfile.gettempdir()) / ".kube"`
- `env["KUBECONFIG"] = os.pathsep.join(...)`
- `hasattr(os, "getuid")` guard on the chown fallback

The only conflict was the **module docstring** in `tests/cli/test_devbox.py` — both sides appended a clause describing their own coverage; kept both. 57/57 tests pass (this PR's plus #1449's). No behavioral change from the rebase.

AdilFayyaz (approved):

PR #1422, opened 2026-08-12
fix: clear error when the PKCE redirect URI has no host:port (FLYTE-SDK-7A)

EngHabu:
## What

[FLYTE-SDK-7A](https://unionai.sentry.io/issues/?query=FLYTE-SDK-7A) — `_create_callback_server` parsed the redirect URI and handed the result straight to `asyncio.start_server`:

```python
server_url = _urlparse.urlparse(typing.cast(str, self._redirect_uri))
server_address = (server_url.hostname, server_url.port)
server = await asyncio.start_server(handler.handle, server_address[0], server_address[1])
```

When the URI is missing, empty, or has no host and port, both components come back as `None` and asyncio raises:

```
ValueError: Neither host/port nor sock were specified
```

`urlparse` is easy to trip here — `urlparse("localhost:8080/callback")` reads `localhost` as the *scheme*, so a redirect URI that merely forgot `http://` also yields `(None, None)`.

The reported event is a good illustration of how badly this reads: a `flyte run` upload got a non-protobuf response from the endpoint, the auth interceptor treated it as retriable and kicked off a browser login, and the login died on this `ValueError` — so the crash the user saw named neither the redirect URI nor the fact that any of it was configuration.

## Fix

The redirect URI comes from the deployment's public client config, so an endpoint that isn't serving the auth metadata service leaves it empty. Validate before binding and raise `InitializationError("InvalidRedirectURI", "user")` naming the offending value.

This mirrors #1235, which gave the sibling case — that same config fetch returning HTML — exactly this treatment. Being a `user`-kind error it is also filtered out of Sentry by the existing `_is_user_error` check.

Note this only ever *replaces* a broken outcome: with no port, `start_server` binds a random one while the browser is redirected to port 80, so the flow hung forever waiting for a callback that could never arrive.

## Testing

New `tests/flyte/remote/test_pkce_callback_server.py`, 7 tests. The 5 validation cases (missing / empty / no-scheme / no-port / path-only) fail on `main` with the original `ValueError`, verified by stashing the source change. The 2 happy-path tests assert the parsed host and port are what actually get bound, and that a non-loopback redirect URI is still accepted — we validate presence, not policy.

fixes FLYTE-SDK-7A


EngHabu (commented):
LGTM

PR #1423, opened 2026-08-12
Make action watch resilient to transient stream interruptions

EngHabu:
## What

Fixes run/action watching (`run.wait`, `Action.wait`, the remote image builder's build wait, CLI monitoring) crashing or silently mis-terminating when the underlying watch stream is interrupted — e.g. `ConnectError: Error reading content` (FLYTE-SDK-6H).

## Why

`ActionDetails.watch` rode a single server stream and treated any end of that stream as authoritative:

- `ConnectError(CANCELED)` — what an intermediary's RST_STREAM surfaces as — was swallowed, ending the generator silently.
- A clean stream close before a terminal phase was indistinguishable from completion.
- Transport-level stream resets weren't caught at all and killed the wait.

Proxies and LBs reset long-lived HTTP/2 streams routinely (idle timeouts, connection churn), so waiting on a slow action — a remote image build, a long task — could crash mid-wait, or worse: `run.wait` returned early, the caller read a non-terminal phase, and the image builder reported `❌ Build failed` for a build that was still running (and often later succeeded).

## How

`ActionDetails.watch` now re-subscribes with exponential backoff (0.5s → 10s cap) on transient failures — `CANCELED`/`UNAVAILABLE`/`DEADLINE_EXCEEDED`, OS-level connection drops/timeouts, and HTTP/2 stream resets — and on clean closes before a terminal phase. Key property: only **consecutive** failed subscriptions count against the reconnect budget (5); any delivered update resets it. So a watch lasting hours survives periodic proxy resets indefinitely, while a genuinely unreachable backend still fails fast.

Classifying the stream reset is the subtle part. You never see the transport's `pyqwest.StreamError` — connectrpc catches it and re-raises it as a `ConnectError` (`_client_async._send_request_bidi_stream`, `raise rst_err from e`), and `maybe_map_stream_reset` sends the `NO_ERROR` / `INTERNAL_ERROR` / `PROTOCOL_ERROR` reset codes onto `Code.INTERNAL`. Retrying every `INTERNAL` would paper over real server bugs, so `_is_transient_watch_error` checks `__cause__` instead: a `ConnectError` is transient if its code is in the transient set **or** a `StreamError` sits underneath it. A genuine server-side `INTERNAL` is built from the response body and carries no such cause, so it stays fatal — as do `PERMISSION_DENIED` and friends.

Concretely, this is what an AWS ALB at its default 60s idle timeout does to a quiet watch stream; before the `__cause__` check it surfaced as a fatal `ConnectError: Error reading content` and the reconnect loop never ran.

The generator now only ends at a terminal phase or raises, which fixes every caller of `run.wait`/`Action.wait` without touching them.

## Testing

New `tests/flyte/remote/test_action_watch_reconnect.py`: reconnect after transient error, reconnect after clean close, non-transient raises, reconnect budget exhaustion, budget reset on delivered updates, the transient-error classifier, and the wrapped-`StreamError` shape connectrpc actually raises (plus its negative case, a causeless server-side `INTERNAL`). Full `tests/flyte/remote` suite passes (575 tests).

### DIFF
--- a/src/flyte/remote/_action.py
+++ b/src/flyte/remote/_action.py
@@ -176,24 +176,36 @@ _WATCH_RECONNECT_INITIAL_BACKOFF_SECS = 0.5
 _WATCH_RECONNECT_MAX_BACKOFF_SECS = 10.0
 
 # CANCELED is what an intermediary's RST_STREAM surfaces as; UNAVAILABLE/DEADLINE_EXCEEDED are
-# the standard transient transport codes. INTERNAL/UNKNOWN stay fatal — they can be real bugs.
+# the standard transient transport codes. INTERNAL/UNKNOWN stay fatal here — they can be real
+# bugs — and are rescued only when _is_stream_reset proves a transport reset underneath.
 _TRANSIENT_CONNECT_CODES = (Code.UNAVAILABLE, Code.DEADLINE_EXCEEDED, Code.CANCELED)
+
+
+def _is_stream_reset(exc: BaseException) -> bool:
+    """Whether exc is (or wraps) a pyqwest HTTP/2 stream reset — "Error reading content".
+
+    connectrpc catches the transport's StreamError and re-raises it as a ConnectError
+    (`raise rst_err from e`), mapping most RST_STREAM codes — NO_ERROR, INTERNAL_ERROR,
+    PROTOCOL_ERROR, ... — onto Code.INTERNAL. Only the __cause__ distinguishes an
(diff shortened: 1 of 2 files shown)


pingsutw (dismissed):

pingsutw:
Wait don't merge it for now

pingsutw:
Pushed a change and tested it locally, and confirm it works

pingsutw (approved):

PR #1431, opened 2026-08-13
fix: remove trailing space from workflow filename (breaks git clone on Windows)

piyush182004:
.github/workflows/validate-codecov-config.yml had a trailing space in its filename. NTFS (Windows) rejects filenames with trailing spaces/dots at the Win32 API level, so git clone/checkout fails entirely on Windows with:

  error: invalid path .github/workflows/validate-codecov-config.yml
  fatal: unable to checkout working tree

This blocks every Windows user from cloning the repo. Renaming the file to drop the trailing space fixes it with no behavior change and everyone can use it.


cosmicBboy:
@piyush182004 looks like `.github/workflows/validate-codecov-config.yml` is a brand new file... can you add an explanation in the PR description for what this new workflow file is for? i.e. what problem it's solving?

piyush182004:
hello @cosmicBboy this file used to exist with a trailing space in the name which broke git clone on Windows, which is what I was fixing. Looks like it's been deleted from main since then, so my fix isn't needed anymore. Closing this PR, thanks for reviewing it.

PR #1435, opened 2026-08-14
Add `flyte proxy app`: authenticated localhost proxy into a Flyte App

mhotan:
## Overview

Adds a `flyte proxy app NAME` command — an authenticated localhost reverse proxy into a no-auth-inside Flyte App. It reuses the CLI's own authenticator to inject a fresh bearer on every request, so a local HTTP client (an MCP server, `curl`, a browser) can reach an app-edge-gated app with no token handling. Think `kubectl port-forward` for Flyte Apps.

## Details

- Resolves the target by app name (`App.get`) or an explicit `--url`.
- Binds loopback on **both `127.0.0.1` and `::1`** so `localhost` — which resolves to `::1` first on macOS — can't be silently shadowed by a wildcard listener from another process (e.g. a container runtime binding `*:PORT` with `SO_REUSEPORT`). Previously an IPv4-only bind still "succeeded" while every request landed on the other listener and reset; now a genuine conflict fails loudly with `EADDRINUSE`. A non-loopback `--address` binds a single family and warns (the proxy is ambient authority for the caller's identity).
- Streams responses unbuffered and preserves `Content-Encoding` — SSE-safe for MCP streamable-HTTP.
- Drops inbound `Authorization`/hop-by-hop headers, injects the fresh bearer on the standard `authorization` header (the SDK auth flow may emit it under a gRPC-style metadata key the HTTP app edge never reads), and refreshes + retries once when the edge bounces an expired token to `/login`.
- `--emit-mcp-config` prints a generic HTTP-MCP client block pointed at the local port.
- WebSocket upgrade proxying is deferred to a follow-up.

## Also included

- **App-serde: materialize delayed parameters on the pod_template path.** When an app parameter is a delayed value (e.g. `RunOutput`), it is materialized to a resolved `File`/`Dir` URI at deploy time. The container path already did this; the `pod_template` serialization path did not — so an app that set a `pod_template` serialized the raw, unresolved getter and the serve entrypoint couldn't download it. Threads `parameter_overrides` through `_get_k8s_pod` / `_serialized_pod_spec` so the pod command carries the resolved value. Independent of the proxy — can be split into its own PR on request.

## Testing

- Unit tests (`tests/cli/test_proxy.py`): inbound-auth/hop-by-hop stripping, `Content-Encoding` preservation, JWT identity display, and the MCP-config block.
- Validated end-to-end against a running app with an upstream `mcp-grafana` server driven by the official MCP Python client SDK (`ClientSession`): `initialize` -> 65 tools -> `list_datasources` and `query_prometheus` returned live data through the proxy with no real credential in the MCP config.
- Dual-stack bind verified: with a wildcard listener holding the port, `curl localhost:PORT` reaches the proxy (not the other listener); a conflicting specific bind fails with `EADDRINUSE`.


src/flyte/app/_runtime/app_serde.py
```diff
@@ -144,7 +145,12 @@ def _serialized_pod_spec(
 
         if container.name == pod_template.primary_container_name:
             container.args = app_env.container_args(serialization_context)
-            container.command = app_env.container_cmd(serialization_context)
+            # Pass the materialized parameters (delayed values like RunOutput already
+            # resolved to their File/Dir URI) so the serve container command carries the
+            # resolved value — matching the container path (get_proto_container). Without
+            # this the pod path serializes the raw, unresolved parameter and the serve
+            # entrypoint can't download it.
+            container.command = app_env.container_cmd(serialization_context, parameter_overrides)
```
mhotan, line 153: This is actually a bug fix I think I found.

src/flyte/cli/_proxy.py
```diff
@@ -0,0 +1,257 @@
+import asyncio
+import errno
+import json
+import sys
+
+import rich_click as click
+
+import flyte.remote as remote
```
kumare3: can we import lazily?

mhotan: Done in 39fd2182b — moved `import flyte.remote as remote` into the `app` command's name-resolution branch, so the heavy remote client stack only loads when resolving an app by name (the `--url` path skips it entirely). Module top now imports only stdlib + `rich_click` (needed for the decorators).

cosmicbboy (approved):

PR #1442, opened 2026-08-15
feat: add flyte.EarlyStopping for training loops

piyush182004:
Adds flyte.EarlyStopping, a small framework-agnostic helper that tracks a metric across training steps/epochs and reports when patience consecutive steps have passed without improvement, so long training runs can stop as soon as they plateau instead of always running to max_epochs.

- src/flyte/_early_stopping.py: EarlyStopping class (step/reset, min/max mode, min_delta threshold)
- tests/flyte/test_early_stopping.py: unit tests covering improvement tracking, patience exhaustion, both modes, min_delta, and reset
- examples/ml/early_stopping.py: runnable example (flyte run --local) demonstrating early stopping on a simulated decaying loss
- FEATURES.md: add Early Stopping to the feature reference table

Verified locally: flyte run --local examples/ml/early_stopping.py train stops at epoch 15 (best loss at epoch 10, patience=5) instead of running all 50 epochs.


cosmicBboy:
hi @piyush182004, can we move the early stopping functionality to a new subpackage `flyte.ml.EarlyStopping`?

piyush182004:
hii @cosmicBboy ,done the functionality to new subpackage in latest commit please check

kumare3:
It seems all frameworks add it, so maybe we should not add it. cc @cosmicBboy

PR #1443, opened 2026-08-16
feat: add flyte.ml.MetricTracker for running metric statistics

piyush182004:
Adds `flyte.ml.MetricTracker`, a small framework-agnostic helper that records a metric across training steps/epochs and reports running statistics (mean, min, max, best-so-far) without deciding whether to stop. Pairs naturally with `flyte.ml.EarlyStopping`: the tracker reports what happened, `EarlyStopping` decides when to act on it.

- `src/flyte/ml/_metric_tracker.py`: `MetricTracker` class (`update`/`reset`, min/max mode, `mean`/`min`/`max`/`best`/`best_step`/`last`/`count`)
- `tests/flyte/ml/test_metric_tracker.py`: unit tests covering running stats, both modes, unchanged-best-on-no-improvement, and reset
- `examples/ml/metric_tracker.py`: runnable example (`flyte run --local`) combining `MetricTracker` with `EarlyStopping` on a simulated decaying loss
- `FEATURES.md`: add Metric Tracker to the feature reference table

## Verification
- 17/17 tests pass (`tests/flyte/ml/`)
- `ruff check` / `ruff format --check` / `mypy` / `check_docstring_style.py` — all clean
- `flyte run --local examples/ml/metric_tracker.py train` — verified locally, reports correct running mean/best each epoch and stops early once the loss plateaus


piyush182004:
@cosmicBboy please review it

kumare3:
Thanks for the contribution @piyush182004 — the code here is clean, well-tested, and well-documented. But after discussion we've decided not to add a `flyte.ml` namespace to the SDK, so I'm going to close this one. Let me explain the reasoning.

**Every training ecosystem already ships these helpers.** Early stopping with patience/min_delta/mode is built into `tf.keras.callbacks.EarlyStopping`, `transformers.EarlyStoppingCallback` (with `load_best_model_at_end`), PyTorch Lightning's and Ignite's `EarlyStopping`, and `flax.training.early_stopping.EarlyStopping` for JAX. Likewise, running-metric tracking exists as `torchmetrics.wrappers.MetricTracker` (same name, plus `best_metric()`), Ignite's `RunningAverage`, and in practice via experiment trackers (W&B/MLflow/TensorBoard) which also persist and visualize the history. Users writing raw training loops who opt out of all of those typically inline the ~15-line snippet directly.

**Flyte's value-add lives at the orchestration layer, not inside the training loop.** These classes don't touch any Flyte machinery — no checkpointing, no reports, no run metrics, no control-plane visibility — so they'd be permanent public API surface without differentiated value, and would invite a parallel reimplementation of training-framework utilities (`flyte.ml.LRScheduler`, `flyte.ml.Checkpointer`, …) that we don't want to maintain.

That said, there IS a Flyte-shaped version of this idea we'd be excited about: **cross-task early stopping** — e.g. pruning sibling trials in a `flyte.map` hyperparameter sweep (Optuna-pruner / ASHA style), driven by metrics the control plane can see and the run-abort machinery that recently landed (#1446). No training library can do that, because it spans tasks. If that interests you, we'd ask for a short design doc / issue first so we can align on the API before code — happy to discuss there.

---
It seems all frameworks add it, so maybe we should not add it. cc @cosmicBboy

PR #1447, opened 2026-08-17
feat: add flyte.ml.GradientAccumulator for gradient accumulation

piyush182004:
> **Note:** base is `feature/metric-tracker` (stacked on `feature/early-stopping`), not `main` — this depends on those two PRs merging first, since `flyte.ml` doesn't exist on `main` yet.

## What
Adds `flyte.ml.GradientAccumulator`, a small framework-agnostic helper that tracks micro-batch counts and tells you exactly when to call `optimizer.step()`, so training loops don't hand-roll the modulo counter (and its off-by-one bugs) that shows up in nearly every large-model training script.

Confirmed genuinely absent from PyTorch, PyTorch Lightning, and HuggingFace Trainer — those frameworks give you the mechanics (`backward()`/`step()`/`zero_grad()`) but not the accumulation bookkeeping itself, which is why this pattern gets copy-pasted so often.

- `src/flyte/ml/_gradient_accumulator.py`: `GradientAccumulator` class (`step`/`reset`, `is_accumulating` property)
- `tests/flyte/ml/test_gradient_accumulator.py`: unit tests covering `accumulation_steps=1`, firing on the Nth step, counter reset after firing, `is_accumulating` state, and correctness across many cycles
- `examples/ml/gradient_accumulator.py`: runnable example (`flyte run --local`) showing exactly which micro-batches trigger an optimizer step
- `FEATURES.md`: add Gradient Accumulator to the feature reference table


piyush182004:
hello @cosmicBboy can you please review it

cosmicBboy:
hey @piyush182004, would you mind adding an example each of how this would work, realistically, with pytorch, huggingface trainer, and pytorch lightning? I can see its utility, but just wanted to see what users would typically do to implement this functionality if they were to write this from scratch vs. using `flyte.ml.GradientAccumulator`

kumare3:
It seems all frameworks have this, so maybe we should not add it. cc @cosmicBboy

PR #1448, opened 2026-08-17
fix(io): fix 0-byte directory placeholder download

danielsalvador:
## Problem

When downloading a directory, let's say `bucket`, of type `flyte.io.Dir` from GCS using `bucket.download_sync()`, the operation fails with:
```
OSError: [Errno 39] Directory not empty: '/tmp/.../run_dir/.flyte-download-u3nhf4r3' -> '/tmp/.../run_dir'
```

This occurs when the directory contains 0-byte directory placeholder objects (folder markers), which GCS uses to represent empty directories.

## Root Cause

GCS and other cloud storage systems store 0-byte directory placeholder objects to explicitly represent empty directories. When `obstore.list()` is called to enumerate objects under a prefix, it returns these placeholders alongside actual data files.

In `flyte/storage/_parallel_reader.py`, the `_gen()` function processes all objects from `obstore.list()` without filtering:

1. For a directory placeholder at the root level (e.g., `prefix/`), computing `rel_path = path.relative_to(src_prefix)` yields `Path(".")`
2. This causes the download target to be calculated as `tmp_dir / Path(".")`, which resolves to the temp directory itself
3. A file buffer is created at the temp directory location instead of a child path
4. Later, when the transformer attempts to move the staging directory with `aiofiles.os.replace(staging_dir, target_dir)`, it fails because:
   - `staging_dir` is still the temp directory (non-empty)
   - `target_dir` is the parent directory
   - POSIX `os.replace()` requires the destination to be empty

This issue was introduced in #1366, which added support for preserving zero-byte files in Dir/File downloads.

## Fix

The fix uses metadata-based directory detection to identify and skip directory placeholders:

1. **Materialize the object list** to enable analysis of parent-child relationships:
   ```python
   objs = [obj async for obj in _list_downloadable()]
   ```

2. **Build a set of parent directories** by extracting all parent paths from the objects:
   ```python
   dir_paths = {str(parent) for o in objs for parent in pathlib.Path(o["path"]).parents}
   ```

3. **Skip directory placeholders** identified by two conditions:
   - `rel_path == pathlib.Path(".")` — root-level entry representing the prefix itself
   - `size == 0 and str(path) in dir_paths` — 0-byte object that is a parent of other files

```python
   if rel_path == pathlib.Path(".") or (size == 0 and str(path) in dir_paths):
       continue
```

This approach uses actual metadata relationships (parent-child file structure) rather than relying on naming conventions. It correctly identifies directory markers across different cloud storage systems regardless of how they represent directories.

## Testing

Added a comprehensive regression test `test_download_files_skips_directory_placeholders()` that verifies:
- Root-level directory placeholders are correctly skipped using metadata (parent detection)
- Subdirectory placeholders (0-byte objects that are parents of other files) are correctly skipped
- Real data files in the directory are downloaded correctly
- Directory structure is preserved in the output

All existing unit tests pass without modification, confirming backward compatibility:
- `test_download_preserves_zero_byte_files` — verifies 0-byte files are still materialized
- `test_download_files_single_file` — verifies single file downloads work
- `test_download_files_multipart` — verifies multipart downloads work
- `test_download_files_no_cross_device_error` — verifies cross-device move safety
- Plus 6 additional unit tests (10/10 passing, 3 sandbox tests skipped)

### DIFF
--- a/src/flyte/storage/_parallel_reader.py
+++ b/src/flyte/storage/_parallel_reader.py
@@ -326,9 +326,15 @@ class ObstoreParallelReader:
             async for obj in _list_downloadable():
                 path = pathlib.Path(obj["path"])  # e.g. Path(prefix/file.txt), needs to be changed to str.
                 size = obj["size"]
-                source = Source(id=path, path=path, length=size)
-                # Strip src_prefix from path for destination
+
+                # Skip directory placeholder objects (0-byte folder markers)
+                # These are typically used by cloud storage (e.g., GCS) to represent empty directories
+                # and end with "/" or have a relative path of "."
                 rel_path = path.relative_to(src_prefix)  # doesn't work on windows
+                if rel_path == pathlib.Path(".") or obj["path"].endswith("/"):
+                    continue
+
+                source = Source(id=path, path=path, length=size)
                 # Emit a single empty chunk for zero-byte objects so the file is still materialized locally
                 chunk_ranges = self._chunks(size) if size else [(0, 0)]
                 for offset, length in chunk_ranges:
--- a/src/flyte/storage/_parallel_reader.py
+++ b/src/flyte/storage/_parallel_reader.py
@@ -323,15 +323,23 @@ class ObstoreParallelReader:
                     yield obj
 
         async def _gen(tmp_dir: str) -> typing.AsyncGenerator[DownloadTask, None]:
-            async for obj in _list_downloadable():
+            # Materialize all objects to detect directory placeholders via parent relationships
+            objs = [obj async for obj in _list_downloadable()]
+
+            # Build a set of all parent directory paths from the objects
+            # This identifies which paths are parent directories of actual files
+            dir_paths = {str(parent) for o in objs for parent in pathlib.Path(o["path"]).parents}
+
+            for obj in objs:
                 path = pathlib.Path(obj["path"])  # e.g. Path(prefix/file.txt), needs to be changed to str.
                 size = obj["size"]
-
-                # Skip directory placeholder objects (0-byte folder markers)
-                # These are typically used by cloud storage (e.g., GCS) to represent empty directories
-                # and end with "/" or have a relative path of "."
+                # Strip src_prefix from path for destination
                 rel_path = path.relative_to(src_prefix)  # doesn't work on windows
-                if rel_path == pathlib.Path(".") or obj["path"].endswith("/"):
+
+                # Skip directory placeholder objects. These are identified by:
+                # 1. Root-level entry (rel_path == ".") — represents the prefix itself
+                # 2. 0-byte object that is a parent of other files in the listing
+                if rel_path == pathlib.Path(".") or (size == 0 and str(path) in dir_paths):
                     continue
 
                 source = Source(id=path, path=path, length=size)
--- a/tests/flyte/storage/test_parallel_reader.py
+++ b/tests/flyte/storage/test_parallel_reader.py
@@ -494,10 +494,13 @@ async def test_download_files_skips_directory_placeholders(tmp_path):
     Regression test for directory placeholder objects causing download failures.
 
     When downloading a directory from GCS/S3, obstore.list() may return 0-byte
-    directory placeholder objects (e.g., "prefix/folder/"). These should be
-    skipped during download to avoid:
-    1. Creating files in the temp directory that would collide
-    2. Failing to replace the staging temp directory onto its parent
+    directory placeholder objects. These are detected as directory markers by checking:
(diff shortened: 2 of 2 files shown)


src/flyte/storage/_parallel_reader.py
```diff
@@ -326,9 +326,15 @@ async def _gen(tmp_dir: str) -> typing.AsyncGenerator[DownloadTask, None]:
...
                 path = pathlib.Path(obj["path"])  # e.g. Path(prefix/file.txt), needs to be changed to str.
                 size = obj["size"]
-                source = Source(id=path, path=path, length=size)
-                # Strip src_prefix from path for destination
+
+                # Skip directory placeholder objects (0-byte folder markers)
+                # These are typically used by cloud storage (e.g., GCS) to represent empty directories
+                # and end with "/" or have a relative path of "."
                 rel_path = path.relative_to(src_prefix)  # doesn't work on windows
+                if rel_path == pathlib.Path(".") or obj["path"].endswith("/"):
```
pingsutw: I think it's better to check if the path is parent dir or not by checking the metadata, like 
```python
objs = [obj async for obj in _list_downloadable()]                                                                                                                                                
dir_paths = {str(parent) for o in objs for parent in pathlib.Path(o["path"]).parents}                                                                                                             
for obj in objs:                                                                                                                                                                                  
    path = pathlib.Path(obj["path"])  # e.g. Path(prefix/file.txt), needs to be changed to str.                                                                                                   
    size = obj["size"]                                                                                                                                                                            
    # Strip src_prefix from path for destination                                                                                                                                                  
    rel_path = path.relative_to(src_prefix)  # doesn't work on windows                                                                                                                            
    if rel_path == pathlib.Path(".") or (size == 0 and str(path) in dir_paths):                                                                                                                   
        continue                                                                                                                                                                                  
    source = Source(id=path, path=path, length=size)
```

danielsalvador: ## Analysis of the Suggested Approach

Thanks for the feedback! I analyzed the suggested approach and found several aspects that, in my opinion, make the current implementation more robust:

#### 1. **Memory Overhead** (Critical)
```python
objs = [obj async for obj in _list_downloadable()]  # Materializes entire list
```
- For large datasets with millions of files, this loads everything into memory upfront
- The current async generator approach is **streaming-friendly** and memory-efficient
- Could cause OOM errors on large downloads

#### 2. **Flawed Logic for Empty Files** (Critical)
```python
if rel_path == pathlib.Path(".") or (size == 0 and str(path) in dir_paths):
```
The `size == 0 and str(path) in dir_paths` check has a critical flaw:
- **It might incorrectly skip legitimate 0-byte files** that should be materialized
- This repo has an existing test: `test_download_preserves_zero_byte_files` that explicitly ensures 0-byte files ARE kept
- The suggested approach could break this guarantee and cause test failures

#### 3. **Doesn't Handle All Cases**
- Empty directories with no files inside — the placeholder won't appear in `dir_paths` if there are no enumerated children
- When downloading specific file paths (via the `paths` parameter), directory markers might be listed but have no children to detect them

#### 4. **Performance Cost**
```python
dir_paths = {str(parent) for o in objs for parent in pathlib.Path(o["path"]).parents}
```
- Computes parents for every object: O(n × depth) operations
- Two passes over the data instead of one
- Additional latency before any download starts (must read entire listing before downloading)

#### 5. **Violates the Streaming Model**
The current implementation fits the async generator pattern:
- Downloads start immediately without waiting for full listing
- Memory usage is bounded by `max_concurrency`
- The suggested approach blocks until all listings complete before processing

### Summary

```python
if rel_path == pathlib.Path(".") or obj["path"].endswith("/"):
    continue
```

✅ **Reliable**: Cloud storage systems (GCS, S3, Azure) consistently use trailing `/` for directory markers  
✅ **Efficient**: Streaming-compatible, O(1) per object  
✅ **Correct**: Preserves 0-byte file behavior (tested)  
✅ **Simple**: Clear intent, easy to understand 

Of course, if you still think your approach is a better choice, I will update my fix.

pingsutw: looks good, could you make this change?

danielsalvador: Sorry, it's late and I'm a bit confused. Which option are you referring to when you say "this change"? 😄

danielsalvador: I have updated the PR with your suggestion.

cosmicBboy:
hi @danielsalvador please run the `prek` linter to fix the ruff format linter issues

danielsalvador:
Done!

pingsutw (approved):
LGTM, thank you

PR #1460, opened 2026-08-21
feat(vllm,sglang): speculative decoding, cache-aware routing, and current engine pins

cosmicBboy:
## Motivation

This PR fills in current gaps in the plugin implementation:

- A draft model is a *second* set of weights, and both environments have room for exactly one — they own a single `model_path` and raise on both `parameters` and `args`.
- vLLM's `--speculative-config` is JSON and `fserve` runs the command through a shell, so the blob has to be quoted or the shell eats its inner quotes.
- Mounts have to sit under `/tmp`, not `/root` — the released Flyte base image runs non-root.
- The pinned engines (vLLM 0.11.0, SGLang 0.5.2) predate the algorithms the recipes are about, and both engines have since moved to CUDA 13 wheels that the images' hand-pinned flashinfer layers cannot match.
- Cache-aware routing has to live inside the pod, because Flyte apps have no session affinity or per-replica address — and the plugin had no way to express it.

This PR closes those gaps so the recipes' configurations are expressible natively.

## Summary

**Speculative decoding (both plugins).** New `speculative_config`, plus optional `draft_model_path` / `draft_model_hf_path`. Draft weights are mounted as a second `Parameter` next to the target and wired into the engine — vLLM's single `--speculative-config` JSON blob, SGLang's flat `--speculative-*` flags (`{"algorithm": "EAGLE3"}` → `--speculative-algorithm EAGLE3`). Draft-model-free methods (ngram) work with no mounted weights. The Flyte streaming loader is process-wide and describes a single set of weights, so configuring a draft model falls back to downloading both models and warns rather than silently feeding the target's tensors to the drafter.

**Shell-safe args (both plugins).** `fserve` runs the app with `Popen(" ".join(args), shell=True)` (`flyte/_bin/serve.py`), so any token with spaces or quotes reached the engine mangled — `--speculative-config` being the obvious casualty. Args are now `shlex.quote`d, except `$VAR` tokens, which `fserve` expands *before* joining and which quoting would turn into literals.

**Writable mount path (both plugins).** Model mounts move from `/root/flyte` to `/tmp/flyte/...`. `_ensure_dest_writable` needs the image's user to create the parent directory, and the released base image runs non-root, so `/root` failed with `Permission denied` before the engine ever started.

**SGLang cache-aware routing.** `router=True` swaps the entrypoint to `sglang_router.launch_server`; worker count and policy stay ordinary server args (`--dp-size`, `--router-policy`). Streaming is unavailable in this mode — the router launches its workers as separate processes that never load the patched loader — so it falls back to download and warns.

**SGLang host binding.** The server now binds `0.0.0.0` unless the caller passes their own `--host`. SGLang's own default is `127.0.0.1`, which is unreachable from outside the container.

**Engine pins and images.** vLLM `0.11.0` → `0.26.0` (DFlash needs 0.20.1+) and SGLang `0.5.2` → `0.5.16` (DFLASH landed 2026-04, on the Spec V2 engine that became default in the 2026-06 line). Both now pull CUDA 13 wheels, so:

- **vLLM image**: `flashinfer-jit-cache` moves to the `cu130` index, pinned to vLLM's own flashinfer version (`0.6.14`), in a layer *after* vLLM — layers resolve independently, so an earlier unpinned layer is just overwritten by vLLM's exact pin and leaves the cache mismatched. `flashinfer-cubin` is dropped: it publishes no `0.6.14`. Without a matching jit-cache, the FlashInfer sampler JIT-compiles during warmup and dies on missing `nvcc`, long after the weights load.
- **SGLang image**: CUDA toolkit `12.8` → `13.0`; the hand-pinned flashinfer layers are dropped (SGLang pins `flashinfer_python[cu13]` itself, and a `cu128` jit-cache against a `cu13` flashinfer is actively wrong); `pre=True` added, which is required rather than cosmetic — 0.5.16 depends on the pre-release `flash-attn-4>=4.0.0b18`, and without it uv reports the package as simply unsatisfiable. Also ships `sglang-router` for `router=True`.

Version-tracking constants (`FLASHINFER_VERSION`, `FLASHINFER_JIT_CACHE_INDEX_URL`, `SGLANG_ROUTER_VERSION`, `CUDA_TOOLKIT_PACKAGE`) are pulled into `_constants.py` with notes on what has to be re-checked together on the next bump.

**Examples.** Every example using these plugins hand-rolled an image that was a near-copy of the plugin default, and each had drifted to a version *below* the new minimum (vllm 0.11.0 / 0.19.0 / 0.20.0, sglang 0.5.7) — so the shims' version check would have rejected them at container startup, and their `cu128`/`cu129` flashinfer layers could not match engines pinning CUDA 13 wheels. Rather than re-pin nine copies, they now build on `DEFAULT_VLLM_IMAGE` / `DEFAULT_SGLANG_IMAGE`: six use it as-is, three `.clone(name=...)` it and add only what they actually need (TriAttention, librosa/soundfile, a separate build cache). The engine version, its matching flashinfer build and the CUDA toolkit now live in exactly one place and cannot drift again.

One casualty: the `transformers==4.57.6` pins are gone, because vLLM 0.26 requires `transformers>=5.5.3` — no vLLM new enough for the plugin can satisfy the old pin. In `vllm_app_artifact.py` that pin carried a comment claiming newer transformers break the `fserve` tokenizer load path; there was no way to keep both, so that claim needs re-validating on a real deploy.

**Prefetch sharding pin.** `flyte.prefetch.hf_model(shard_config=...)` shards a checkpoint with vLLM's `save_sharded_state`, and this plugin's `FlyteModelLoader._load_sharded_model` is the only thing that reads those files back — rebuilding the expected `state_dict` from *its own* vLLM's model implementation and raising `Missing keys ... in loaded state!` on anything it cannot fill. The two are halves of one format, but nothing tied their versions together: `VLLM_SHARDING_IMAGE_PACKAGES` asked for `vllm>=0.11.0`, a floor rather than a pin, so it installed whatever was newest at image-build time — two prefetch runs weeks apart could shard with different vLLMs, and neither was pinned to the one doing the serving. It is now pinned to `0.26.0`, and the sharding image's CUDA toolkit moves 12.9 → 13.0 to match the wheels vLLM 0.26 pins (the toolkit is there so startup JIT compiles have an `nvcc`; one from the wrong CUDA major fails inside a warmup compile). A test in the vLLM plugin's suite asserts the two constants stay equal, so a one-sided bump fails in CI rather than at model-load time on a real deploy.

## Test Plan

- 13 new vLLM tests and 19 new SGLang tests covering: draft-model mounting, HF draft models, draft-model-free (ngram) configs, forced download when a draft model or the router is configured, the `speculative_config` validation errors, flag rendering (including `speculative_`-prefixed keys and booleans), shell quoting of JSON args, `$VAR` args left unquoted, router entrypoint selection, host defaulting, and `clone_with` round-trips that add or drop a draft model.
- 2 new tests guarding the prefetch/plugin version contract, verified to actually fail (with an actionable message) when the two constants are desynced.
- Existing suites pass unchanged apart from the two intentional mount-path assertions: **39 passed** (vLLM), **43 passed** (SGLang), **64 passed** (`tests/flyte/prefetch`).
- All nine example modules import and their app envs build with the inherited images; `make fmt`, `make mypy`, `make ty` and `make check-docstrings` clean across the repo.
- `ruff format --check` and `ruff check` clean on both plugins; `mypy` clean on the changed files (the 2 remaining errors are pre-existing, in the untouched SGLang shim).

⚠️ **Not exercised:** the container image builds themselves, which need a GPU-class builder and a registry (`maint_tools/build_plugin_image.py`). The pins are verified against PyPI and the flashinfer indexes — `flashinfer-jit-cache==0.6.14` exists on `cu130`, and the previously pinned `cu128`/`cu129` indexes top out at `0.6.9`, so they could never have matched the new engines. Worth a real build + serve before merge.

**Out of scope, adjacent:** `examples/ml/batch_inference_saturate.py` and `batch_inference_saturate_app.py` install an unpinned `vllm` alongside `flashinfer-python==0.6.4` / a `cu129` jit-cache. Unpinned resolves to 0.26.x, which pins flashinfer exactly at 0.6.14 — so those layers are both overwritten and CUDA-mismatched. Same bug class as this PR fixes, but a different subsystem: they use vLLM directly and touch neither plugin nor the sharded-state format, so they are left alone here.

### DIFF
--- a/examples/genai/nemotron_omni_voice/README.md
+++ b/examples/genai/nemotron_omni_voice/README.md
@@ -104,8 +104,10 @@ print(r.json()["choices"][0]["message"]["content"])
 ## Notes / knobs
 
 - **Version pins are deliberate.** The Omni checkpoint needs **vLLM ≥ 0.20** and
-  `--trust-remote-code`; both are set in `serve.py`. Bump `VLLM_VERSION` there if
-  NVIDIA's recipe moves.
+  `--trust-remote-code`. The vLLM pin comes from the plugin's `DEFAULT_VLLM_IMAGE`
+  (which is well past that floor) and `--trust-remote-code` is set in `serve.py`.
+  If NVIDIA's recipe moves to a newer vLLM, append a `.with_pip_packages("vllm==<version>")`
+  layer to the image there.
 - **Want lower cost over lower latency?** Set `Scaling(replicas=(0, 1), scaledown_after=300)`
   to scale to zero when idle — at the cost of a cold start (multi-GB load) on the next request.
 - **BF16 instead of FP8** needs ~64 GB → switch `MODEL_REPO` to the BF16 variant
--- a/examples/genai/nemotron_omni_voice/serve.py
+++ b/examples/genai/nemotron_omni_voice/serve.py
@@ -40,7 +40,7 @@ Deploy
        python examples/genai/nemotron_omni_voice/voice_client.py --endpoint <app-url>
 """
 
-from flyteplugins.vllm import VLLMAppEnvironment
+from flyteplugins.vllm import DEFAULT_VLLM_IMAGE, VLLMAppEnvironment
 
 import flyte
 import flyte.app
@@ -60,21 +60,17 @@ MODEL_ID = "nemotron-omni"  # the name clients pass as `model=...`
 # ---------------------------------------------------------------------------
 # Image
 #
-# The Omni model needs vLLM >= 0.20 and trust-remote-code. We mirror the
-# plugin's default image (flashinfer for fast FP8 attention on Ada/L40S) but
-# pin a newer vLLM and add the audio decoding libraries vLLM uses to read the
-# incoming wav/mp3 payloads.
+# The Omni model needs vLLM >= 0.20 and trust-remote-code. The plugin's default
+# image already satisfies the version floor (and carries a flashinfer build
+# matching that vLLM, for fast FP8 attention on Ada/L40S), so this only adds the
+# audio decoding libraries vLLM uses to read the incoming wav/mp3 payloads.
+#
+# To follow a newer vLLM than the plugin pins, append your own layer:
+#   .with_pip_packages("vllm==<version>")
 # ---------------------------------------------------------------------------
 
-VLLM_VERSION = "0.20.0"
-
 image = (
-    flyte.Image.from_debian_base(name="nemotron-omni-vllm", install_flyte=False)
-    .with_pip_packages("flashinfer-python", "flashinfer-cubin")
-    .with_pip_packages("flashinfer-jit-cache", index_url="https://flashinfer.ai/whl/cu129")
-    .with_pip_packages("flyteplugins-vllm", pre=True)
-    # vLLM goes in its own layer (dependency conflict with flyte on protovalidate).
-    .with_pip_packages(f"vllm=={VLLM_VERSION}", "transformers>=4.57.0")
+    DEFAULT_VLLM_IMAGE.clone(name="nemotron-omni-vllm")
     # Audio decoding for the multimodal input pipeline.
     .with_pip_packages("librosa", "soundfile")
 )
--- a/examples/genai/sglang/sglang_app.py
+++ b/examples/genai/sglang/sglang_app.py
@@ -39,29 +39,15 @@ print(response.choices[0].message.content)
 ```
 """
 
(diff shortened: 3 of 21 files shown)


cosmicBboy:
cc @kumare3 please review

kumare3 (approved):

PR #1463, opened 2026-08-21
docsy: detect plugin-provided CLI groups that have no callback

ppiegaze:
`get_plugin_info` returns early when `cmd.callback` is `None`, so a `click.Group` that only dispatches to subcommands is reported as **core** regardless of which package defined it. The module path — the thing that actually decides — is never reached.

```python
if not cmd or not cmd.callback:
    return False, None
module = cmd.callback.__module__   # never gets here for a dispatch-only group
```

A callback-less Group is an ordinary construct, and having no callback says nothing about provenance.

## Why it matters

Three consequences in the generated CLI reference, all from that one branch:

1. the plugin marker (`⁺`) is dropped from the command matrix
2. the *"provided by the `<plugin>` plugin"* note is dropped from the section
3. **the command is emitted into the OSS Flyte variant** — the variant filter keys off the same `is_plugin` flag

So a reader of the open-source docs gets a full section, examples and options for a command that requires a plugin install and a Union backend.

## The case that surfaced it

`flyteplugins-union` 0.8.0 registers `flyte fork` as a Group subclass instantiated directly:

```python
fork = ForkFiles(name="fork", help=...)      # cli/fork.py — no callback
```

`flyte debug`, from the same package, is unaffected — but only because it is written as `@click.command("debug")` on a function and so carries a callback. **The two differ in construction style alone**, which is not a fact about where they came from. `debug` gets its note and is correctly excluded from the OSS variant; `fork` gets neither.

Observed live in the 2.6.2 → 2.6.3 regen of the docs site (unionai/unionai-docs#1483), where `flyte fork` appears inside the `{{< variant flyte >}}` block.

## The change

Prefer the callback's module; fall back to `type(cmd).__module__`. For a plugin-supplied group that is the plugin's own module, which the existing checks already classify correctly.

One guard: an un-subclassed `click.Group()` resolves to click's own module, which is equally uninformative — without handling it the function reports a plugin named `"click.core"`. That case returns core.

## Tests

`tests/cli/test_gen_plugin_detection.py`, 6 cases: the regression, the decorated-plugin control that must not break, a core command, a core `Group` subclass, the bare `click.Group`, and `None`.

Verified adversarially — with the fix reverted, `test_plugin_group_without_callback_is_detected` fails and the other five pass:

```
1 failed, 5 passed      (without the fix)
6 passed                (with it)
```

Full `tests/cli` suite: **390 passed**.

## Scope note

Verified against a faithful reproduction of how `flyteplugins-union` constructs the group (same module path, same instantiation), **not** against the shipped 0.8.0 wheel — the local resolver's `exclude-newer` cutoff predates that release. The logic under test is package-agnostic, but worth stating.

A plugin-side workaround exists (give the group a no-op callback), but it papers over the defect and the next plugin registering a dispatch-only group hits the same wall.

---
— docsy · automated docs agent · [DOC-1478](https://linear.app/unionai/issue/DOC-1478)

### DIFF
--- a/src/flyte/cli/_gen.py
+++ b/src/flyte/cli/_gen.py
@@ -274,12 +274,15 @@ def _build_index_table(
 
             # Filter entries based on include_plugins
             filtered = [(n, ip, pm) for n, ip, pm in entries if include_plugins or not ip]
-            if not filtered and not key_is_plugin:
-                # Verb has no non-plugin nouns — still show it if it's a core verb
+            if not filtered:
+                # The verb is listed in this table but has no subcommands to
+                # show -- either it takes none, or it builds them dynamically.
+                # Still list it: dropping it makes a documented command
+                # unreachable from the index. Detecting `fork` as plugin-provided
+                # (above) puts it in exactly this shape, and `debug` is already
+                # in it today.
                 verb_link = f"[`{key_display}`](#flyte-{key})"
                 output.append(f"| {verb_link} | - |")
-            elif not filtered:
-                continue
             else:
                 noun_links = []
                 for noun, noun_is_plugin, _ in filtered:
--- a/src/flyte/cli/_gen.py
+++ b/src/flyte/cli/_gen.py
@@ -112,13 +112,13 @@ def get_plugin_info(cmd: click.Command) -> tuple[bool, str | None]:
     The module a command was *defined* in is what decides this. Prefer the callback's
     module; fall back to the command object's own class when there is no callback.
 
-    A ``click.Group`` that only dispatches to subcommands has ``callback is None`` —
-    an ordinary construct, not a signal about provenance. Returning early on it
+    A `click.Group` that only dispatches to subcommands has `callback is None`, an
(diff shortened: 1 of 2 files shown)


ppiegaze:
Follow-up commit: the detection fix alone had a side effect I missed.

The index lists a verb with no filterable subcommands only when it is **not** plugin-provided. `flyte fork` is a plugin group that builds its subcommands dynamically, so correcting its detection moved it from *listed in the OSS index* (the wrong table, the bug being fixed) to *listed in no index at all* — a documented command with nothing linking to it. Verified by regenerating against `flyteplugins-union` 0.8.0.

That branch is now unconditional: by the time control reaches it, the verb is already known to belong in the table being built.

Side effect, and a second real fix: `flyte debug` is in the same state on `main` today, and is restored.

---
— docsy · automated docs agent · [DOC-1478](https://linear.app/unionai/issue/DOC-1478)

cosmicBboy (approved):

PR #1464, opened 2026-08-21
docsy: resolve CLI doc variants per plugin instead of globally

ppiegaze:
The CLI doc generator takes a single `--plugin-variants` value and applies it to **every** plugin. That can only describe a world where all plugins ship to the same audience, and they do not: `flyteplugins-union` is Union-only, while `flyteplugins-hydra` ships inside flyte-sdk and is available to open-source users.

The failure is silent and points the wrong way. Install an OSS CLI plugin in the docs regen environment and its commands get wrapped in a `union` variant block, hiding them from the Flyte docs entirely.

Reproduced against the real generator with `flyteplugins-hydra` installed:

| | invocation | `flyte hydra` renders in |
|---|---|---|
| **A** | `main`, `--plugin-variants union` | `{{< variant union >}}` — **wrong, hidden from OSS docs** |
| **B** | this branch, same flags | `{{< variant union >}}` — unchanged, back-compat |
| **C** | this branch + `--plugin-variant-map 'flyteplugins.hydra=flyte union'` | both variants |

In C the Union commands stay Union-only, so the two are now independent.

## What changed

`--plugin-variant-map` (repeatable) maps a module prefix to the variants that plugin ships to:

```
--plugin-variant-map 'flyteplugins.hydra=flyte union'
```

`--plugin-variants` remains the default for anything unmapped, so existing invocations are unaffected. `--variants` supplies the page's full variant list, which `_non_plugin_variants` had hardcoded to `{flyte, union}`.

Two smaller fixes fall out of the same conflation:

- Index tables are built **per variant** instead of as a hardcoded core/full pair, so a plugin visible in several variants appears in each.
- The Union-specific legend is emitted only when a `flyteplugins.union` command is actually present, and a second note names any other contributing distribution. Previously every `⁺` was attributed to `flyteplugins-union` whatever its origin.

Malformed `--plugin-variant-map` entries raise instead of being skipped — silently dropping one would mis-gate the very plugin it named.

## Verification

**Near-behavior-preserving for the current pipeline.** Regenerating with the docs invocation (`--plugin-variants union`, `flyteplugins-union` 0.8.0) gives output identical to `main` across all 2744 lines except (a) the three rows documenting the new options and (b) one restored index entry:

```
> | [`debug`](#flyte-debug) | - |
```

`flyte debug` is **missing from the index on `main` today**. The index listed a verb with no filterable subcommands only when it was *not* plugin-provided, so a plugin group that builds its subcommands dynamically kept its section but had nothing linking to it. That branch is now unconditional, since by the time control reaches it the verb is already known to belong in the table being built.

This also matters for #1463: correcting `fork`'s detection without this would have moved it from *listed in the wrong index* to *listed in no index at all*. #1463 carries the same fix so it is correct standing alone.

16 new tests in `tests/cli/test_gen_variant_mapping.py`, covering prefix-boundary matching (`flyteplugins.union` must not match `flyteplugins.unionx`), longest-prefix precedence, malformed entries, and per-variant index filtering. `make fmt` / `mypy` / `ty` / `check-docstrings` clean; `tests/cli` 400 passed.

## Relationship to #1463

Independent and non-overlapping. #1463 fixes *which* commands are detected as plugin-provided; this fixes *where* a detected plugin command is published. Either can merge first.

---
— docsy · automated docs agent

### DIFF
--- a/src/flyte/cli/_gen.py
+++ b/src/flyte/cli/_gen.py
@@ -8,6 +8,9 @@ import rich_click as click
 
 import flyte.cli._common as common
 
+# The distribution whose commands the Union-specific legend below describes.
+UNION_PLUGIN_DIST = "flyteplugins.union"
+
 
 @click.group(name="gen")
 def gen():
@@ -25,13 +28,35 @@ def gen():
     default=None,
     help="Hugo variant names for plugin commands (e.g., 'union'). "
     "When set, plugin command sections and index entries are wrapped in "
-    "{{< variant >}} shortcodes. Core commands appear unconditionally.",
+    "{{< variant >}} shortcodes. Core commands appear unconditionally. "
+    "Applies to any plugin not named in --plugin-variant-map.",
+)
+@click.option(
+    "--plugin-variant-map",
+    "plugin_variant_map",
+    type=str,
+    multiple=True,
+    default=None,
+    help="Per-plugin variant override, as 'module_prefix=variants' (repeatable). "
+    "Example: --plugin-variant-map 'flyteplugins.hydra=flyte union'. "
+    "Plugins are not all shipped to the same audience, so a single "
+    "--plugin-variants value cannot describe them all.",
+)
+@click.option(
+    "--variants",
+    "all_variants",
+    type=str,
+    default="flyte union",
+    help="Every Hugo variant the generated page declares (default: 'flyte union'). "
+    "A command visible in all of them is emitted unwrapped.",
 )
 @click.pass_obj
 def docs(
     cfg: common.CLIConfig,
     doc_type: str,
     plugin_variants: str | None,
+    plugin_variant_map: tuple[str, ...] | None = None,
+    all_variants: str = "flyte union",
     project: str | None = None,
     domain: str | None = None,
 ):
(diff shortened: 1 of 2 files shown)


ppiegaze:
Closing in favour of [DOC-1481](https://linear.app/unionai/issue/DOC-1481), which moves the Hugo-rendering half of this generator into `unionai-docs-infra`.

Every line of this PR is on the *renderer* side of that split — `_build_index_table`, `_non_plugin_variants`, the variant shortcode emission, the plugin legend. Merging it would mean asking you to review Hugo shortcode logic and variant policy in the SDK, and then deleting it from the SDK a few weeks later. The variant mapping belongs in docs config, not in a CLI flag.

The work is not lost: it lands in the infra renderer instead, where the third-audience problem is actually solvable. Once the split has a JSON boundary carrying a `distribution` field per command, the module-prefix guess this PR makes (`flyteplugins-union` → `flyteplugins.union`) is unnecessary — entry-point ownership is exact.

**One fix from this branch has moved to #1463**, because it is needed there and stands alone: a verb with no listable subcommands was only indexed `if not key_is_plugin`, so correcting `fork`'s detection would have moved it from *listed in the wrong index* to *listed in no index at all*. `flyte debug` is in that state on `main` today and is restored by it.

#1463 is unaffected and still worth merging on its own.

---
— docsy · automated docs agent · [DOC-1480](https://linear.app/unionai/issue/DOC-1480)

PR #1466, opened 2026-08-21
feat(rerun): support new inputs alongside --recover, in the CLI and the API

cosmicBboy:
## What

`flyte rerun` had no way to change inputs at all, and `rerun(..., recover=True)` rejected changed inputs outright (pointing at fork). Both now work, and they compose.

- **Inputs are options on `flyte rerun`**, built from the source run's interface with the same literal-type-to-click-type conversion `flyte run` uses — `flyte rerun rxyz --n 10` reads like `flyte run ... --n 10`. `flyte rerun rxyz --help` lists them.
- **`flyte.rerun(..., x=2)`** keeps its existing `**inputs` varkwargs; nothing changed about how the Python API takes inputs.
- **`recover` + changed inputs is allowed.** The new run starts from the changed inputs while still reusing the source run's succeeded actions. A warning points out that recovered actions keep the outputs they produced under the *original* inputs, and that `force_rerun_actions` / `--force-rerun-action` is how to re-execute them.
- **Partial input changes merge** onto the source run's inputs instead of failing with "missing required inputs": only the changed inputs are converted and overlaid on the prior literals. A full set of new inputs still skips the source-input fetch entirely, which keeps that path working as the escape hatch when the source inputs are gone from storage. Unknown input names are rejected up front rather than silently dropped.

Partial merging wasn't strictly required by the ask, but without it `flyte rerun r1 --recover --n 10` would fail on any task with more than one required input, so the flag would have been a footgun.

Code substitution stays fork's job — only inputs moved.

## How the CLI options get built

The source task's interface is only discoverable once RUN_NAME has been read off the command line, so `RerunCommand` parses in two passes: rerun's own options first (a resilient parse that tolerates the not-yet-built input options), then the interface fetch, then the real parse with both sets. Nothing left over from the first pass means no input was passed, so a plain `flyte rerun r1` costs no extra round trip.

Two deliberate differences from `flyte run`, both following from "an input left out keeps the source run's value":

- No input option is required and none carries the task's default. Only what the user actually typed is forwarded, determined from click's parameter source.
- Booleans get `--x/--no-x`, since "not passed" already means "keep the prior value" and False has to stay expressible.

An input whose name would shadow one of rerun's own options (e.g. a task input called `name`) is skipped rather than silently taking it over; it keeps the prior run's value.

## Test plan

- `tests/cli` + `tests/flyte/test_rerun.py` + `tests/flyte/test_run_runspec_chars.py` — 439 passed
- `tests/flyte` top-level files — 498 passed (3 failures in `test_deploy_image_cache.py` / `test_int_image_builder.py` reproduce identically on the base commit: no local container registry)
- `tests/flyte/{remote,ai,internal,cli}` + `tests/internal/controllers` — 1516 passed
- `ruff check` / `ruff format`, `mypy`, `ty` all clean (repo pre-commit hooks passed)

New coverage: an option per input with click-native conversion, omitted inputs not being sent, `--x/--no-x` both ways, inputs composing with `--recover`, the no-input path skipping the interface fetch, `--action-name` sourcing the right interface, rerun's own options not being shadowed, unknown options rejected, `--help` with and without a run name; plus partial-input merge (plain and with recover), full-input-set skipping the fetch, the unmergeable `--allow-missing-outputs` case, unknown-input rejection, and the recover-with-inputs warning.

### DIFF
--- a/examples/rerun/README.md
+++ b/examples/rerun/README.md
@@ -10,14 +10,17 @@ One verb, two behaviours:
 | ---- | ------------ |
 | `flyte.rerun(run)` / `flyte rerun <run>` | A whole new run with the same inputs. Every action executes again, subject to global caching. |
 | `flyte.rerun(run, recover=True)` / `flyte rerun <run> --recover` | A whole new run with the same inputs, but actions that already succeeded are reused as-is. Only what failed or never ran executes. |
+| `flyte.rerun(run, x=2)` / `flyte rerun <run> --input x=2` | Same code, changed parameters. Every input left out keeps the prior run's value. Composes with `recover=True` / `--recover`. |
 
 Reused actions land in the `RECOVERED` phase — terminal and success-equivalent — so
 `flyte get action <run>` tells you exactly what recovery skipped.
 
-Recovery is durability against *intermittent* failures, not a way to patch a run. It replays
-the source run's code and inputs as-is; the run environment (`-e KEY=VALUE`) is the only lever
-you get. `recover=True` combined with changed inputs raises — change inputs on a plain rerun
-instead.
+Recovery is durability against *intermittent* failures. It always replays the source run's
+*code* as-is — substituting local code is `flyte fork`, reserved for `flyteplugins-union`.
+Inputs and the run environment (`-e KEY=VALUE`) are the levers you get: `recover=True` combined
+with changed inputs starts the new run from those inputs, while every recovered action keeps the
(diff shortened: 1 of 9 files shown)


SVilgelm (approved):
Tested on demo: https://demo.hosted.unionai.cloud/v2/domain/development/project/sergey/runs/uqf5jl8f9vj9ghp7mgn5

PR #1467, opened 2026-08-21
adding changes to make UV related environment configuable

debasisdwivedy:
Hi,

Issue described below:

Set python env as 3.12/3.13 etc and use `apt get python python-lib` the linker does not point to the right file error `libpython3.12.so.1.0: cannot open shared object file`

This PR changes the `DOCKER_FILE_UV_BASE_TEMPLATE` to add the linker based on the python version.

Regards


debasisdwivedy:
DO NOT merge it. It was based on Rust-sdk testing. We might not need it. I will keep the team posted

PR #1475, opened 2026-08-24
docsy: `--type markdown` emits plain Markdown, and six RST literals the style check could not see

ppiegaze:
The docs pipeline no longer asks this repo for a finished page — it asks for `--type json` and renders the Hugo itself. So the Hugo in here has no consumer, and this removes it.

**The hold this PR carried is satisfied.** It was opened as a draft on 24 August because merging it before the pipeline switched would have removed a flag that pipeline still passed, failing every regen. Since then:

| | |
| -- | -- |
| `--type json` released | `flyte 2.6.6`, and now 2.6.9 |
| infra switched to it | unionai-docs-infra#276, live |
| **first production regen through the new renderer** | unionai-docs#1503 — the CLI page diff was **exactly** the 39 plugin-name corrections plus the version bump, and nothing else |

That last row is the one that mattered: the port was predicted byte-identical from a scratch repo root, and production reproduced it line for line.

## What goes

`{{< grid >}}` / `{{< markdown >}}` wrappers, `{{< variant >}}` gating, `--plugin-variants`, the escaping of `{{<` and `{{%` in **user** help text, the option-alias layout chosen to avoid raw HTML under goldmark `unsafe=false`, and the rewrite of indented help blocks into fenced code — which existed only because indented code did not survive one site's `RenderString`.

None of that is a fact about this CLI. Those workaround comments existed because the knowledge lived in another repository, and an SDK maintainer had no way to verify any of them.

## What stays

**Where a command came from is a fact about the command.** A plugin-provided command keeps its `⁺` marker and names its distribution:

```
### flyte fork

> **Note:** This command is provided by the `flyteplugins-union` plugin.
```

**Who should see it is a decision, and that goes** — including the hardcoded `{"flyte", "union"}` audience set, which could not express a third audience without an SDK release.

## `get_plugin_info` is deleted, not bypassed

It inferred provenance from `__module__` and returned early when a command had no callback, which is how `flyte fork` — a dispatch-only `click.Group` — reached the open-source CLI reference. Referenced nowhere but `_gen.py` and its own tests, so the inference is now **absent from the codebase** rather than merely unused. Both output modes report what `_plugins.py` recorded at registration.

One visible consequence: the per-command note and the summary block now agree on the plugin's name. They disagreed in the same document — `flyteplugins.union` against `flyteplugins-union` — because one was derived from `__module__` and the other hardcoded.

## Verification

- `_gen.py`: **610 → 371 lines**
- generated output: **zero** Hugo constructs
- `tests/cli`: **433 passed**; `make cli-docs-gen`, `check-docstrings`, ruff and format clean
- `--type json` unaffected: 92 commands, `declares_options` on all of them
- rebased onto current `main` and re-verified

`test_gen_plugin_detection.py` is deleted with the function it tested; its two index-rule tests are carried into `test_gen_markdown.py`, whose first test is parametrised over `{{<`, `{{%`, `variant` and `grid` and asserts none reach the output.

---


ppiegaze:
Added a second, related commit: **six RST inline literals that `check-docstrings` structurally cannot see**, one of which is live on the published reference.

`DOUBLE_BACKTICK_RE` is `` ``[^`\n]+`` `` and the scan is line-at-a-time. Docstrings wrap at the line limit, so a literal starting near the margin has its closing delimiter on the next line — invisible. Six were live and every file reported clean.

**One reaches readers.** `flyte.PodTemplate`'s page renders:

```
container.apparmor.security.beta.kubernetes.io/&lt;primary&gt;: unconfined
```

The `` ``…`` `` is not a code span to the API generator, so its angle brackets were escaped as prose — and then it formed a code span anyway at render time, trapping the entities where nothing decodes them. A single-backtick span **is** recognised and its angle brackets are left alone: the generated reference already carries `` `[mcp_servers.<name>]` `` and `` `--from-file <path>` `` correctly, which is how I confirmed the fix rather than assuming it.

So all six become single-backtick spans, and the check gains a pass over the whole docstring instead of a line at a time — with fenced regions blanked first, so an example containing a wrapped literal is still left alone. Verified both directions: a reintroduced wrapped literal is caught, one inside a fence is not.

Three of the six needed their surrounding prose rewrapped rather than merely unwrapped, to stay inside the 120-char limit.

Files: `src/flyte/_pod.py`, `src/flyte/ai/chat/_html.py`, `src/flyte/types/_type_engine.py`, and two agent plugins.

**492 files clean · 433 tests · ruff and format clean on every file touched.** No infra change is needed — the generated page corrects itself at the next regen.

---

ppiegaze:
**`rust lint` is red and it is not from this branch.** Flagging rather than leaving it for a reviewer to work out.

- this branch changes **zero** Rust files
- its merge-base is `f8d3f030`, which is **exactly** the commit where `rust lint` ran **green on `main`** at 10:16 — same Rust code, opposite result
- re-running the job reproduced the failure, so it is not a one-off

The lints are `clippy::result_large_err` on `flyte_core`'s tonic streaming signatures — pre-existing code untouched here.

The likeliest explanation is the Cargo cache: `rs-lint` restores a `target/` dir, and clippy does not re-emit diagnostics for crates it considers unchanged. A warm cache on `main` therefore reports clean, while a cache miss on a PR rebuilds and surfaces what was always there. Consistent with `main`'s own history, which has an intermittent failure in it (`79107259`).

If that is right, the lint is real and `main` is green by caching rather than by cleanliness — worth someone with Rust context confirming, since it means any PR that misses the cache fails on it.

**It does not block this PR.** `rust lint` is not among the 7 required checks, and all 7 are green.

---

ppiegaze:
Follow-up on the previous commit: **my first pass was too clever and missed three more.**

It still looked for a *pair* of delimiters. The shapes actually present were:

| shape | example | caught by |
| -- | -- | -- |
| literal on one line | `` ``dict`` `` | the original check |
| literal wrapped at the line limit | `` ``dict of``<br>`` things`` `` | the previous commit |
| **stray backtick, delimiters mismatched** | `` `x`` `` / `` ``x` `` | **nothing** |

The third pairs with nothing, so no pattern hunting for a pair could see it — and there were two, both rendering as visible garbage wherever they appear (`flyte/_debug/ssh.py`, and the deepagents memory module).

The house rule is that a docstring never contains a double backtick, so the check now **says that**, rather than recognising shapes. Simpler, and it subsumes all three. Fenced blocks stay exempt — an example may legitimately show RST.

That immediately turned up a fourth kind in `flyte/remote/_settings.py`: `` `` `` used as an **empty code span**, meaning "replaced with nothing". CommonMark has no empty code span, so it was never going to render as intended. The sentence now says it in words.

## Where it stands

Verified with an independent scan that looks for `` `` `` **anywhere** outside a fence, not with the checker's own patterns:

```
src/               2,390 docstrings   0 occurrences
plugins/*/src      2,240 docstrings   0 occurrences
```

And the check discriminates correctly — all three bad shapes flagged, a fenced example and a clean docstring both pass.

## Deliberately not in scope

`examples/` still holds roughly 2,900. It is not scanned by this tool and does not reach the API reference, so it is a house-style inconsistency rather than a rendering defect. Worth its own sweep; folding ~2,900 edits into this PR would bury the change it is actually about.

492 files clean · 433 tests · ruff and format clean.

---

ppiegaze:
**Correcting my earlier comment about `rust lint`.** It is now **green** (51s), on the same untouched Rust code.

I said re-running had reproduced the failure and that it was therefore not a one-off. Two runs failed and this one passed, so "reproducible" was too strong — it is **intermittent**, which fits the cache explanation better than it contradicts it: clippy does not re-emit diagnostics for crates it considers unchanged, so whether the lints surface depends on what the runner restored.

The substance stands and is still worth a look from someone with Rust context: the `clippy::result_large_err` findings on `flyte_core`'s tonic signatures are real and pre-existing, and `main`'s own history has an intermittent failure in it. If that reading is right, `main` is green by caching rather than by cleanliness, and any run that misses the cache fails on it.

Nothing here changes for this PR — it touches no Rust, and all 58 checks now pass.

---

cosmicBboy (approved):

kumare3 (approved):

PR #1477, opened 2026-08-24
Raise instead of blocking when a sync task is called from an async task

dansola:


cosmicBboy:
@dansola mind resolving the conflicts and also adding a PR description?

kumare3:
#1472 shipped the same guard and restructured sync-task execution, so the `SYNC_EXECUTION_LOOP_ATTR` detection here targets code that no longer exists on main. Closing as superseded.

PR #1478, opened 2026-08-25
add GPU metrics examples for the run details metrics tab

samhita-alla:
gpu_metrics.py walks a GPU through distinct load regimes (tensor, bandwidth, PCIe, NVLink, idle), each held long enough to be visible at the console's polling resolution, so every panel on the Metrics tab gets a recognizable signature. gpu_metrics_multinode.py is the ClusteredTaskEnvironment variant, one pod per node with an all-reduce phase so the cross-node interconnect shows up too.

Both can end with a deliberate Xid 31 from an out-of-bounds Triton store, which is how the GPU fault path gets exercised end to end: after_xid picks whether the task then returns, raises, or exits outright, matching the three shapes a fault takes by the time it reaches the console.


cosmicbboy (approved):
🚀

PR #1481, opened 2026-08-25
fix: preserve image build run link on image cache hits

AdilFayyaz:
## Motivation                                                                         

  When an image already exists in the registry, no build runs, so TaskMetadata.image_build_run was never stamped and the console lost its link from a task's image to the build that produced it. The image service now reports the originating build run on a GetImage hit, so the SDK can stamp it without a build.

 ## Summary

  - Add ImageBuild.build_run, carrying the run that originally produced an image when no build was launched in this process.
  - RemoteImageChecker.image_exists records Image.build_run from the GetImage response into a process-local side channel (_image_build_runs), which ImageBuildEngine.build attaches to cache-hit results. Needed because the public ImageChecker protocol (flyte.extend) returns only a URI.
  - _build_image_bg falls back to build_run when remote_run is absent; a build launched in-process still takes precedence.
  - _build_images now propagates build_run_ids from seed_cache, so nested runs don't drop the link.

##  Test Plan

  - uv run pytest tests/flyte/imagebuild/ tests/flyte/test_deploy.py
  - New tests/flyte/imagebuild/test_remote_image_checker.py covers: build run recorded on hit, unset build_run, partial identifier rejected, bare {id, fqin} response, and recording failures swallowed. The three tests needing the new field skip on flyteidl2==2.0.42.
  - New engine/deploy tests cover cache-hit attachment, remote_run precedence over a recorded run, and seed-cache propagation.
  - Manual: deploy an env whose image already exists, confirm the task's image links to the original build run in the
    console.


src/flyte/_internal/imagebuild/remote_builder.py
```diff
@@ -61,6 +61,40 @@
...
+    Image.build_run only exists on flyteidl2 >= 2.0.43 and is only populated by backends
+    that track it, so missing schema or unset field is a silent no-op. Isolated from the
+    caller's control flow: a failure here must never turn a successful existence check
+    into a rebuild.
+    """
+    try:
+        from flyte._internal.imagebuild.image_builder import RunIdentifierData, record_image_build_run
+
+        # Schema check must come first: HasField raises ValueError for fields the
+        # installed flyteidl2 doesn't know about. Both the descriptor and the field are
```
pingsutw: It shouldn't happen, right? we will pin the idl version in the sdk

src/flyte/_build.py
```diff
@@ -21,10 +22,16 @@ class ImageBuild:
...
+        build_run: Identifier of the remote build run that originally produced the image,
+            when known without a build having been launched (the registry existence check
+            can learn it from the image service on a cache hit). None when a build ran in
+            this process (use remote_run), for locally built images, and for backends that
+            don't report it.
     """
 
     uri: str | None
     remote_run: Optional["remote.Run"]
+    build_run: Optional["RunIdentifierData"] = None
```
pingsutw, line 35: why don't we just override remote_run if the image exists?

src/flyte/_internal/imagebuild/remote_builder.py
AdilFayyaz: yes, but have to cut a release first

src/flyte/_build.py
AdilFayyaz: because `remote_run` is a live Run object. on a cache hit we'd only get back the run id from it unless we add a separate API call. Its better to have another field `build_run` which has the already completed run that built the image

pingsutw: but we only use run Identifier at the end here https://github.com/flyteorg/flyte-sdk/blob/530b7b0cf8d586312adb3baded15103bd88a2e95/src/flyte/_deploy.py#L417-L418

pingsutw: or should we only need build_run?

pingsutw: i feel like we only need one to make it cleaner

src/flyte/_internal/imagebuild/remote_builder.py
pingsutw: yes, let's cut a release first. people don't usually upgrade idl

src/flyte/_build.py
AdilFayyaz: yes two fields can be confusing. Ill make `build_run` the one field that always says "this run built the image" 
the remote builder will set it when it launches a build and the cache hit already sets it from the image service. 
Then `_deploy.py` only reads the `build_run` and the if/else goes away. 

I would still keep `remote_run` tho, because its the live Run object and the `wait=False` path returns so callers can wait on the build. We cant do that with just an id.

pingsutw: I see that makes sense

pingsutw (dismissed):

pingsutw (approved):

PR #1487, opened 2026-08-26
docs: backtick literals that were rendering as prose, and check for it

ppiegaze:
Flag names, values and file names in help text and docstrings were written three ways — `'local'`, `--recursive`, and `` `local` `` — and only the third renders as monospace.

The CLI reference shows the effect inside a single table row: the **Option** column is monospaced and the **Description** column beside it is not, for the same flag.

## This is drift, not a distinction

```
CLI help strings:  31 backticked spans   vs   52 single-quoted tokens
help strings using both styles:  0
```

Zero mixing means no author ever drew a considered line — each picked a style and kept to it. Backticks were already the majority.

## What changed

**51 literals across 17 files.** Quoted tokens and phrases become code spans, and so do bare `--flags`:

```diff
- help="MCP transport. 'stdio' speaks JSON-RPC on stdin/stdout and ignores --port/--mcp-mount-path."
+ help="MCP transport. `stdio` speaks JSON-RPC on stdin/stdout and ignores `--port`/`--mcp-mount-path`."

- help="Image builder to use for building images. Defaults to 'local'."
+ help="Image builder to use for building images. Defaults to `local`."
```

Both halves of `--port`/`--mcp-mount-path`, and **every** item of *"Accepts ISO dates, `now`, `today`, or `now - 1 day`"* — a rule that converted only single words would have left that sentence inconsistent **with itself**, which is worse than the uniform inconsistency it replaces.

## Deliberately out of scope

Each measured rather than assumed:

| left alone | why |
| -- | -- |
| SCREAMING_CASE identifiers | `RANK` and `SCRIPT` have no underscore, `WORLD_SIZE` does — a mechanical rule backticks one word of a list and not its neighbour, and widening it catches `API`/`JSON`/`HTTP` in prose. Needs judgement. |
| 48 concatenated multi-line `help=` literals | rebuilding collapses the source onto one line; an earlier attempt produced **28** line-length violations |
| ~98 quoted tokens in indented docstring examples | the quotes are Python syntax and must stay |
| `examples/`, `tests/` | not published |

Docstring delimiters are preserved — an earlier attempt rewrote `"""x"""` as `"x"`, which the formatter does not catch.

## Verification

427 tests · `check-docstrings` 492 files clean · `make cli-docs-gen` green · ruff clean on every file touched. The diff is a 1:1 line swap, **51 insertions / 51 deletions**, no structural change.

Independent of #1475 — different files, no overlap.

---

### DIFF
--- a/maint_tools/check_docstring_style.py
+++ b/maint_tools/check_docstring_style.py
@@ -117,6 +117,15 @@ WRAPPED_RE = {
 # `text <url>`_ and the anonymous `text <url>`__ form.
 RST_HYPERLINK_RE = re.compile(r"`[^`\n]*<[^>\n]+>`__?")
 RST_GRID_TABLE_RE = re.compile(r"^\s*\+[-=+]{3,}\+")
+# A literal in quotes, or a bare flag, renders as prose. The CLI reference shows
+# the effect inside one table row: the Option column is monospaced and the
+# Description column beside it is not, for the same flag. Backticks were already
+# the majority style before this was enforced.
+QUOTED_LITERAL_RE = re.compile(r"(?<![`\w\[])'([A-Za-z0-9_][A-Za-z0-9_.:/@=-]*|\S[^'\n]*\s[^'\n]*\S)'(?!\w|\])")
+BARE_FLAG_RE = re.compile(r"(?<![`\w-])(--[a-z][a-z0-9-]{2,})(?![\w-])")
+# Lines that are example code, not prose: the quotes there are Python syntax and
+# the flags are being demonstrated, so both must be left alone.
+CODEISH_RE = re.compile(r"^(\s{4,}|\s*(\$|>>>|flyte |python |uv |pip |make ))")
 RST_FOOTNOTE_RE = re.compile(r"\[\d+\]_")
 FENCE_RE = re.compile(r"^\s*```")
 
@@ -127,6 +136,8 @@ FIXES = {
     "rst-directive": "use plain prose, or a fenced code block",
     "sphinx-field": "use Google style: 'Args:' / 'Returns:' / 'Raises:'",
     "double-backtick": "use a single-backtick Markdown code span: `value`",
+    "quoted-literal": "wrap the literal in backticks so it renders as monospace: `value`",
+    "bare-flag": "wrap the flag in backticks so it renders as monospace: `--flag`",
     "rst-hyperlink": "use a Markdown link: [text](url)",
     "rst-grid-table": "use a Markdown table",
     "rst-footnote": "inline the reference, or use a Markdown link",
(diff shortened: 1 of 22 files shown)


ppiegaze:
Rewritten since the first push. Two commits now, and the second one **caught a bug in the first**.

## 1. The sweep, in two passes

**87 rewrites across 23 files.** The second pass exists because rebuilding an implicitly-concatenated `help=` literal from its parsed value collapses the source onto one line — 28 line-length violations on my first attempt. Those are now rewritten **token by token, where they sit**, so line structure is untouched.

## 2. The check — which found the sweep corrupting text

Two rules on docstrings **and** `help=` strings: a quoted literal, and a bare `--flag`. `help=` was never scanned by this tool at all, which is precisely why a flag could be monospaced in one column of a table row and plain prose in the next.

Running it over the swept tree immediately flagged one line, and it was my own damage:

```diff
  original   (replacing '.' with '-')
- swept      (replacing '.` with `-')
```

The phrase rule had matched the `' with '` **between** two separate literals. Fixed by requiring the content to begin and end with a non-space, then the sweep was re-run from a clean base. Worth stating plainly: an automated pass over prose needs an independent check, because it will find a shape you did not think of.

## Verified in both directions

| case | result |
| -- | -- |
| quoted value in `help=` | flagged |
| bare flag in `help=` | flagged |
| quoted value in a docstring | flagged |
| already backticked | clean |
| indented example | clean |
| fenced example | clean |
| ordinary apostrophes (`file's`, `doesn't`) | clean |
| two adjacent literals (`'.'` with `'-'`) | clean |

## On `now - 1 day`

Called out because it drove the design. `DateTimeType` splits on `(.+)\s+([-+])\s+(.+)`, so it is genuinely a value you pass — confirmed against the shipped package:

```
'now'             → 2026-08-26 13:54
'now - 1 day'     → 2026-08-25 13:54
'today + 2 hours' → 2026-08-26 02:00     (the + form is undocumented)
'now - 1 fortnight' → rejected
```

A single-token rule would have backticked `now` and `today` and left `now - 1 day` in quotes — the same sentence, two renderings.

**427 tests · check-docstrings 492 files clean · cli-docs-gen green · ruff clean.**

---

ppiegaze:
`rust lint` is red and is not from this branch — **0 Rust files changed**. Same intermittent, cache-dependent failure already raised on #1475: `clippy::result_large_err` on `flyte_core`'s tonic signatures, pre-existing, and green or red depending on what the runner restored.

All **7 required checks pass**: `make fmt`, `make mypy`, `unit-tests (3.11)`, `unit-tests (3.12)`, `check flyteidl2 versions`, `check uv.lock`, `DCO`.

---

cosmicBboy (dismissed):

ppiegaze:
Rebased onto `main` after #1475 merged. Thanks for the review.

**One conflict, and it resolved itself.** The sweep had backticked `'union'` inside the `--plugin-variants` help text; #1475 deleted that option outright, so the deletion wins and the edit is simply moot. That is the only change from what you approved — the diff is otherwise identical, now **86 insertions / 86 deletions** rather than 87, for exactly that reason.

Re-verified against the CI gates rather than assumed: `check-docstrings` 492 files clean · 433 tests · `make cli-docs-gen` green · `make fmt` idempotent.

---

ppiegaze:
Heads-up: the force-push dismissed the approval automatically — this repo drops stale reviews on new commits, so it now shows 0 approvals and the merge is blocked on that alone. Nothing else changed and CI is green (57/57).

@cosmicBboy sorry for the second ask; the rebase was unavoidable once #1475 landed, since both touched the same option.

---

cosmicBboy (approved):

PR #1500, opened 2026-08-29
Handle backoff with new error

wild-endeavor:
## Summary

Handle `RESOURCE_EXHAUSTED` in the remote controller instead of the generic RPC retry interceptor.

- Introduce `ResourceExhaustedError` as a specialized `SlowDownError`.
- Retry with capped exponential backoff while the controller is running.
- Track resource-exhaustion attempts separately without incrementing `action.retries`.
- Log a warning for each retry.

## Why

Resource exhaustion is not a terminal action error and may persist for an arbitrary amount of time. Consuming the normal bounded retry budget would eventually fail runs that should continue waiting.

## Testing

- Added coverage confirming `RESOURCE_EXHAUSTED` bypasses the generic retry interceptor.
- Added coverage confirming repeated resource exhaustion does not consume the action retry budget.
- Tested in staging with:
  - `run_concurrency: 1`
  - `action_concurrency: 2`
  - `depth: 5`
- Observed retries backing off from 0.5s to 1s, 2s, 4s, and 8s.


src/flyte/_internal/controllers/remote/_core.py
```diff
@@ -501,6 +518,8 @@ async def _bg_launch(self, action: Action):
                     if e.code == Code.ALREADY_EXISTS:
                         logger.info(f"Action {action.name} already exists, continuing to monitor.")
                         return
+                    if e.code == Code.RESOURCE_EXHAUSTED:
+                        return await self._bg_handle_resource_exhausted(action, e)
```
kumare3: why do this separately, all other backoff is in this function itself right

src/flyte/_internal/controllers/remote/_core.py
```diff
@@ -401,6 +401,23 @@ async def _bg_cancel_action(self, action: Action):
...
             await informer.fire_completion_event(action.name)
 
+    async def _bg_handle_resource_exhausted(self, action: Action, error: ConnectError) -> None:
+        action.resource_exhausted_retries += 1
+        backoff = min(
+            self._min_backoff_on_err * (2 ** min(action.resource_exhausted_retries - 1, 20)),
+            self._max_backoff_on_err,
+        )
+        logger.warning(
+            "Resource exhausted for action %s; retrying in %.1fs (attempt %d): %s",
```
kumare3: I think a better way to handle this is --                     raise flyte.errors.SlowDownError(f"Failed to launch action: {e.message}") from e
Or similar error. This then simply yields and lets the queue handle it again

src/flyte/_internal/controllers/remote/_core.py
```diff
@@ -401,6 +401,23 @@ async def _bg_cancel_action(self, action: Action):
...
+            action.name,
+            backoff,
+            action.resource_exhausted_retries,
+            error.message,
+        )
+        await asyncio.sleep(backoff)
```
kumare3: check how we handle --                 except flyte.errors.SlowDownError as e:
-- keeping all this logic in multiple places doesnt feel great. how about create a new error called flyte.errors.QueueFull etc

src/flyte/_internal/controllers/remote/_core.py
```diff
@@ -401,6 +401,23 @@ async def _bg_cancel_action(self, action: Action):
         if informer:
             await informer.fire_completion_event(action.name)
 
+    async def _bg_handle_resource_exhausted(self, action: Action, error: ConnectError) -> None:
+        action.resource_exhausted_retries += 1
```
pingsutw: Should we add a max_resource_exhausted_retries?

pingsutw: or do we want to retry forever?

src/flyte/remote/_client/auth/_interceptors/retry.py
```diff
@@ -6,7 +6,7 @@
 from connectrpc.code import Code
 from connectrpc.errors import ConnectError
 
-RETRYABLE_CODES = frozenset({Code.UNAVAILABLE, Code.RESOURCE_EXHAUSTED, Code.INTERNAL})
```
pingsutw, line 11: we don't want to use RetryUnaryInterceptor? why?

wild-endeavor: maybe we do... my reasoning is that unavailable is a network level thing.  resource_exhausted we think is not a thing that's sent by nginx/envoy/nlbs, so basically the network request successfully went through - but the server said hold off.

honestly i don't know why we retry on code.internal either - we should not, that would just add to the storm... but i suppose that does give us time to live patch something.  we should probably add a sentry on this to see how often we hit this/recover.

src/flyte/_internal/controllers/remote/_core.py
wild-endeavor: my vote is retry forever.  upcoming observability should expose long running tasks to the user to manually abort.

pingsutw (dismissed):

pingsutw (approved):

PR #1509, opened 2026-08-31
introduce service-account on task environment

fiedlerNr9:
## Summary

- Add `service_account` as a task-environment/task-level setting for deployed tasks.
- Serialize the configured service account into `TaskTemplate.security_context.run_as.k8s_service_account` alongside existing secret handling.
- Preserve service-account and secret fields when overriding remote task security context.

## Why

`flyte.run(..., service_account=...)` already supports setting the Kubernetes service account for an individual run, but deployed tasks did not have an equivalent default. This is needed for `flyte deploy` so tasks can be registered with the intended pod identity.

This matters especially for trigger-invoked runs: those runs are launched by the platform from the deployed task definition, so there is no caller-provided `flyte.run(..., service_account=...)` at invocation time. Storing the service account on the deployed task template gives trigger-created executions the correct Kubernetes identity.

## Validation
- I tested this internally on demo
``` python
# hello.py
import flyte

env = flyte.TaskEnvironment(
    name="my_map_workflow",
    service_account="default",
    image=flyte.Image.from_debian_base(install_flyte=False)
    .with_apt_packages("git")
    .with_commands(
        "pip install git+https://github.com/flyteorg/flyte-sdk.git@91d1133de945fbdef91039641110585420358e34"
    ),
)

every_minute = flyte.Trigger(name="service-account-test", automation=flyte.FixedRate(1))


# use TaskEnvironments to define tasks, which are regular Python functions.
@env.task
def fn(x: int) -> int:  # type annotations are recommended.
    slope, intercept = 2, 5
    print(slope * x + intercept)
    return slope * x + intercept


# tasks can also call other tasks, which will be manifested in different containers.
@env.task(triggers=every_minute)
def main(x_list: list[int] = list(range(10))) -> float:
    x_len = len(x_list)
    if x_len < 10:
        raise ValueError(f"x_list doesn't have a larger enough sample size, found: {x_len}")

    y_list = list(flyte.map(fn, x_list))  # flyte.map is like Python map, but runs in parallel.
    y_mean = sum(y_list) / len(y_list)
    return y_mean


if __name__ == "__main__":
    flyte.init_from_config("config.yaml")
    deployed_env = flyte.deploy(env)

```
- `ruff format --check src/flyte/_internal/runtime/task_serde.py`
- `ruff check src/flyte/_environment.py src/flyte/_task_environment.py src/flyte/_internal/runtime/task_serde.py`
- `mypy --config-file pyproject.toml src/flyte/_environment.py src/flyte/_task_environment.py`
- `ty check src/flyte/_environment.py src/flyte/_task_environment.py`

### DIFF
--- a/src/flyte/_environment.py
+++ b/src/flyte/_environment.py
@@ -49,6 +49,7 @@ class Environment:
             `Resources` object.
         env_vars: Environment variables as `dict[str, str]`.
         secrets: Secrets to inject into the environment.
+        service_account: Kubernetes service account to run task pods as.
         pod_template: Kubernetes pod template as a string reference to a
             named template or a `PodTemplate` object. To set a termination grace
             period without depending on the `kubernetes` package, use
@@ -75,6 +76,7 @@ class Environment:
     interruptible: bool = False
     image: Union[str, Image, Literal["auto"], None] = "auto"
     include: Tuple[str, ...] = field(default_factory=tuple)
+    service_account: Optional[str] = None
 
     # Absolute path of the user file where this environment was instantiated.
     # Populated in __post_init__. Used to anchor relative `include` paths.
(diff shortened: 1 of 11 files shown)


EngHabu (commented):
I see you did put it in the Environment, which is good. Can we make sure APpEnvironment also gets it? I'm not sure if there is an Override in the base Environment class? so both tasks and apps can override the service account?

fiedlerNr9:
> I see you did put it in the Environment, which is good. Can we make sure APpEnvironment also gets it? I'm not sure if there is an Override in the base Environment class? so both tasks and apps can override the service account?

Patched AppEnvironment as well! There is no override for service-account in the base Environment class. But actually im unsure if im utilising the base class correctly here?

pingsutw (approved):

EngHabu (approved):

PR #1511, opened 2026-09-01
Consolidate the SaaS integrations: one webhook plugin, idempotent_run in core, recipes over vendor SDKs

cosmicbboy:
Replaces the six per-product integration plugins (#1502–#1507) with the parts that are genuinely Flyte's, and hands the rest back to the vendors' own maintained clients.

**This supersedes #1502, #1503, #1504, #1505, #1506, and #1507.** They are still open; closing them is your call.

## Why

Reviewing those six surfaced a shape problem. The valuable idea — idempotent run launching — was product-agnostic and shipped six times, byte-identical. The client layer was ~2,200 lines of 1:1 REST wrappers competing with `PyGithub` and `slack_sdk`. The MCP servers competed with first-party ones from GitHub, Notion, Linear, and Atlassian. And ~840 lines of setup instructions were compiled into Python, needing a package release to fix a typo.

The cost was concrete: every follow-up in this series — the async rework, the sign-off, the event constants — had to be applied six times, nine commits per branch.

## What ships

**`flyte.extras.idempotent_run`** — launch a run once per event key. Idempotency lives on a run *label*, never a name: names race against concurrent launches and cap how many runs one key can ever have. Async-first behind `@syncify`, so a handler awaits it rather than stalling its event loop.

**`plugins/webhooks`** — one app for all five providers. GitHub, Slack, Linear, ClickUp, and Jira land on one receiver at `/webhook/{provider}`, each verified with its own scheme and normalized into a single `WebhookEvent`. One dashboard instead of six, one dedupe implementation instead of six copies, and one place to fix a verification bug. Typed event constants per provider (`events.github.PullRequest.OPENED`).

**`examples/external_saas_integrations`** — recipes, not wrappers. `PyGithub`, `slack_sdk`, `jira`, and `gql` are maintained by people with live API access; a task calling them directly needs nothing in between. `review_pr` moves here as an example, since `flyte.new_condition` is the only part of it that needed inventing.

## What is dropped

- **The six API clients.** Use the vendors' own.
- **The six MCP servers.** Every one of these vendors ships a better-maintained MCP server.
- **Notion entirely.** It has no webhooks — a polling app never belonged beside five receivers. The polling recipe is a scheduled task, not a plugin.

## Numbers

| | before | after |
| --- | --- | --- |
| plugin source | ~11,400 lines | ~1,500 |
| dashboards | 6 | 1 |
| API clients | 6 | 0 |
| MCP servers | 6 | 0 |
| packages to release | 6 | 1 |

## Correctness carried forward

Every fix from the review of the six is preserved: bytes-based `compare_digest` (a non-ASCII header turned a clean 401 into a 500), Slack's raw-bytes HMAC basestring, the timestamp in the dedupe key so `update` events do not collapse onto one key, distinct comments on one issue not colliding, Linear's team id nested on comment payloads, ClickUp's nested list id, fail-closed scope allowlists, and dashboards reading the newest events rather than the oldest.

267 tests: 204 in core, 63 in the plugin. `make fmt`, `make mypy` (856 files), `make ty`, ruff, and codespell all pass.

## Two things worth knowing

Task names in the examples are **environment-qualified** (`github-triage.triage_pr`). Bare names never resolve — a latent bug in the original six examples — and both the Linear and Jira recipes define a `triage_issue`, so the qualifier is what keeps them apart. There is a check in the test pass that every name the receiver launches is actually defined.

The caveat from the review stands: **the webhook parsers still encode assumptions about payload shapes I cannot verify without live accounts.** That surface is much smaller now — the API clients and their ~213 field assumptions are gone — but the parsers and event constants remain unverified. The testing guide in `plugins/webhooks/README.md` is written so one real pass per provider surfaces anything wrong.

### DIFF
--- a/examples/external_saas_integrations/README.md
+++ b/examples/external_saas_integrations/README.md
@@ -1,0 +1,63 @@
+# External SaaS integrations
+
+Recipes for driving GitHub, Slack, Linear, ClickUp, and Jira from Flyte.
+
+There is deliberately **no Flyte client plugin** for these products. Each vendor
+already ships (or the community maintains) a Python client that is tested
+against the live API by people who get deprecation notices first, and a task is
+just a function — so calling `PyGithub` or `slack_sdk` from a task needs nothing
+in between. A wrapper here would only add a surface to keep in sync with someone
+else's release calendar.
+
+What *is* Flyte's job, and what these examples use from it:
+
+- **`flyte.extras.idempotent_run`** — launch a run once per event key. Webhook
+  senders retry on any non-2xx, pollers overlap, people click retry; this makes
+  all of that safe by keying on a run label.
+- **`flyteplugins-webhooks`** — authenticate an inbound delivery with the
+  provider's own scheme and normalize it into one event model.
+- **`flyte.new_condition`** — park a run on a human decision, with a typed
+  payload coming back.
+
+## The examples
+
+| File | What it shows | Client |
+| --- | --- | --- |
+| `webhook_receiver.py` | One app receiving from all five providers, launching a task per event | `flyteplugins-webhooks` |
+| `github_pr_review_gate.py` | Human-gated merge: a condition carrying JSON, parsed into a typed decision | `PyGithub` |
+| `github_triage_pr.py` | Label, comment, and report a check run | `PyGithub` |
+| `slack_notify.py` | Post, thread, react, answer a mention | `slack_sdk` |
+| `linear_triage_issue.py` | Query a backlog and comment, over GraphQL | `gql` |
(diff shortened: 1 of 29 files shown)


cosmicBboy:
Superseded by a stack that restructures this into one package per product over a shared core, matching `plugins/agents`:

- #1512 `flyteplugins-webhooks-core` — the contract, the app, `idempotent_run`, the conformance harness
- #1513 github · #1514 slack · #1515 linear · #1516 clickup · #1517 jira
- #1518 the SaaS integration recipes

Closing in favour of those.

PR #1524, opened 2026-09-03
GPU faults are catchable typed exceptions

samhita-alla:
# Why

The backend now classifies GPU faults into the failure a task reports: the error code names the fault (GpuXidError, GpuFallenOffBus, GpuEccUncorrectable, GpuRowRemapPending, GpuNvlinkError, GpuGspError), the message leads with the driver's own sentence, and ExecutionError carries a typed gpu_fault with the Xid, severity, GPU UUID and node. Until now the SDK converted all of that into a generic runtime error, so a parent task could only string-match on the code. This makes GPU faults catchable as typed exceptions.

# What

- flyte.errors gains GPUFaultError, carrying fault_kind, fault_code, fault_name, severity, gpu_uuid, gpu_index, node, pci_bus_id and process, plus xid and sxid properties that answer only for the matching kind, since Xid and SXid share a numbering space but not a meaning. Two leaves, GPUFaultUserError and GPUFaultSystemError, keep the error kind honest: the backend files a user-severity Xid (an out-of-bounds access in the workload's own kernel) as USER and a critical hardware fault as SYSTEM, and a single class would have to lie about kind for one of them. except flyte.errors.GPUFaultError catches both.
- The error conversion selects the GPU classes on an exact match of the six backend codes, ahead of the existing substring branches, and tolerates server-prefixed codes. Fields fill only from the typed gpu_fault; a failure carrying none leaves every fault attribute at None, with the raw message untouched. The message is prose, not a contract, so nothing parses it — the backend that classifies a fault always attaches the typed data, since the two shipped together.
- The docstring states the semantics plainly: user-severity Xids arrive on the task's own retry budget, while critical hardware faults are retried by the platform without charging that budget, so user code sees those only after platform policy gives up.

```python
try:
    result = await train(shard)
except flyte.errors.GPUFaultError as e:
    if e.severity == "USER":
        raise  # the workload's own kernel is at fault
    result = await train(shard)
```

# Tests

31 tests in the conversion suite: ruff, mypy and ty clean; the touched test modules pass 187 tests total.


src/flyte/_internal/runtime/convert.py
```diff
@@ -473,6 +474,147 @@ async def convert_outputs_to_native(interface: NativeInterface, outputs: Outputs
         return tuple(kwargs[k] for k in interface.outputs.keys())
 
 
+# The sentence the backend prepends to the failure message when it attributes the failure to a GPU fault, for example
+# "[gpu-health] [CRITICAL] Xid 79 (GPU has fallen off the bus) on GPU 3 GPU-1a2b." or
+# "[gpu-health] [CRITICAL] SXid 22 on NVSwitch 0000:3b:00.0.". It is the human half of the GPU health daemon's event
+# message, and it is what the fault attributes are read from when a failure arrives with no typed fault on it, for
+# example from a backend older than the one that added the field. The trailing full stop is matched only where a space
+# or the end of the message follows it, so a PCI bus id keeps its own dots.
+_GPU_FAULT_SENTENCE = re.compile(
```
kumare3: this looks very brittle, i dont think we should create an expectation that we will have this. We can provide this as a helper. Otherwise users will expect we will always come up with this error. but, this cannot be guaranteed

samhita-alla: yeah, removed the regex matching altogether. the code now reads gpu_fault. if it's not available, msg passes through untouched.

samhita-alla:
@kumare3 would you mind reviewing this PR when you have the time?

kumare3 (approved):

PR #1527, opened 2026-09-03
artifacts: parent lineage, duck-typed declaration protocol, content-hash versions

EngHabu:
Core plumbing so plugin-provided offloaded assets (Union Volumes first) can publish as artifacts, and so any artifact can carry git-style parent lineage. Pairs with unionai/flyteplugins-union (volume wiring) and rides flyteidl2 2.0.48's `repeated parent_artifacts` (flyteorg/flyte#7972).

## What changed

- **`Metadata.parents`** — ordered parent edges (`ArtifactParent` or a bare version string = same-name parent), serialized onto `ProducedArtifact.parent_artifacts` (declarative path) via `to_produced_artifact`. A bare entry stays keyless on the wire so the service inherits the child's name/scope; `ArtifactParent` carries a key only when overriding.
- **`Metadata.version_from_content`** — opt-in: an empty version resolves to the literal's content hash (`Literal.hash`) at conversion/create time, so republishing identical content is idempotent. Opt-in on purpose: File/Dir defaults are unchanged.
- **Duck-typed declaration protocol** — output conversion now recognizes any top-level output exposing `get_flyte_metadata() -> Metadata | None`, not just `ArtifactWrapper`. A new `current_output_declares_artifact()` contextvar (set only for the output actually being declared, never for nested values) lets the value's transformer stamp registry knowledge into the serialized value.
- **`Artifact.create(parents=...)`** — imperative path parity: fills `ArtifactSpec.parent_artifacts`; duck metadata seeds name/version/description/attrs/card/parents exactly like the wrapper does; `ensure_artifactable` accepts protocol-carrying values.
- **flyteidl2 pinned to 2.0.48** (incl. `rs_controller` pyproject/Cargo pins + lock checksums).

## Usage

**Declare an artifact on a returned value.** `artifacts.new` wraps an offloaded asset (`File`, `Dir`, `DataFrame`) zero-copy; it still behaves exactly like the underlying object.

```python
import flyte.artifacts as artifacts
from flyte.io import File

@env.task
async def train() -> File:
    f = await File.from_local("model.pt")
    return artifacts.new(f, artifacts.Metadata(name="model"))
```

**Content-hash versions** — republishing unchanged content is idempotent, and an explicit `version` still wins:

```python
artifacts.new(f, artifacts.Metadata(name="model", version_from_content=True))
```

**Multi-parent lineage.** A bare string is shorthand for a same-name parent (an earlier version of this artifact); `ArtifactParent` carries `name`/`project`/`domain`/`org` when the parent lives elsewhere. Order is preserved, so a merge records which side came first:

```python
artifacts.new(
    merged,
    artifacts.Metadata(
        name="dataset",
        parents=("v3", artifacts.ArtifactParent(version="v7", name="dataset-b")),
    ),
)
```

Parents are stored as given and never resolved — a parent that was never published is legal, like a git remote missing commits nobody pushed. Only a direct self-reference is rejected.

**Duck-typed declaration.** Any value exposing `get_flyte_metadata() -> Metadata` participates without importing the wrapper, which is how `flyteplugins-union` volumes publish themselves (unionai/flyteplugins-union#106):

```python
class WarmVenv:
    def get_flyte_metadata(self) -> artifacts.Metadata:
        return artifacts.Metadata(name="warm-venv", version_from_content=True)

@env.task
async def prepare() -> WarmVenv:   # declared on the Outputs envelope; no RPC from the task
    ...
```

**Imperative parity** for code that registers directly:

```python
await Artifact.create(name="dataset", version="v8", parents=("v7",))
```

## Testing

- `tests/flyte/test_produces_artifacts.py`: parent serialization shapes, ordered multi-parents, duck-typed declaration + contextvar behavior, content-hash version precedence (explicit > hash > empty).
- `tests/flyte/remote/test_artifact.py`: `create(parents=...)` wire shapes, duck seeding, hash-defaulted versions.
- Full artifact/convert suites green locally (205 passed).

https://claude.ai/code/session_01N415hnSCNmZ2e7WrwEPJo7


src/flyte/artifacts/_wrapper.py
```diff
@@ -108,17 +108,23 @@ def __contains__(self, item):
...
 
-    if not isinstance(obj, (File, Dir, DataFrame)):
-        raise TypeError(
-            f"values of type {type(obj).__name__!r} cannot be artifacts; artifacts are offloaded "
-            "assets: flyte.io.File, flyte.io.Dir, or flyte.io.DataFrame "
-            "(wrap a raw dataframe with DataFrame.from_df())"
-        )
+    if isinstance(obj, (File, Dir, DataFrame)):
+        return
+    if not isinstance(obj, type) and callable(getattr(obj, "get_flyte_metadata", None)):
```
kumare3: i think it should be get_artifact_metadata

src/flyte/remote/_artifact.py
```diff
@@ -258,16 +265,25 @@ async def create(
         cfg = get_init_config()
 
         obj = value
+        md: Metadata | None = None
         if type(value) is ArtifactWrapper:
-            md: Metadata = value.get_flyte_metadata()
+            md = value.get_flyte_metadata()
             obj = value._obj
+        elif callable(getattr(value, "get_flyte_metadata", None)):
```
kumare3: better to make it a protocol? 

```
from typing import Protocol, runtime_checkable

@runtime_checkable
class ArtifactMetadata(Protocol):
    def get_artifact_metadata(self) -> None: ...

class File:
    def get_artifact_metadata(self) -> None:
        print("closed")

isinstance(File(), ArtifactMetadata)   # True
```

tests/flyte/test_produces_artifacts.py
```diff
@@ -357,3 +358,230 @@ async def test_nested_wrapper_in_container_rejected(self):
...
+        )
+        assert len(decl.parent_artifacts) == 0
+
+
+_duck_flags: list[bool] = []
+
+
+class DuckAsset(BaseModel):
+    """A type outside flyte.io that opts into artifact declaration by exposing
+    the metadata protocol — the seam plugin-provided volumes ride."""
```
kumare3, line 436:
```suggestion
    the metadata protocol."""
```

kumare3 (approved):

PR #1528, opened 2026-09-03
feat(prefetch): add file selection (allow_patterns/ignore_patterns) to hf_model

mhotan:
## Intent
`flyte.prefetch.hf_model` prefetches the **entire** HuggingFace repo. That makes it unusable for **GGUF** repos, which ship many quantizations at one commit (100s of GB) when a caller wants exactly one (~18 GB). This adds file selection so `hf_model` can prefetch just the files you want — the missing piece for using the prefetch → model-artifact pattern with `flyteplugins-llamacpp` (#1521).

## What changed
- **`allow_patterns` / `ignore_patterns`** threaded through `HuggingFaceModelInfo` → `hf_model()` → `store_hf_model_task` into both download paths.
- **Streaming path** (`_stream_to_remote_dir`) now lists the repo **recursively** (`list_repo_files`) and filters with `huggingface_hub.utils.filter_repo_objects`, **preserving each file's repo-relative subpath** — so nested quant layouts are fully captured, not just top-level files. (Previously a flat `ls` silently skipped subdirectories.)
- **Snapshot-download fallback** forwards the patterns; the unfiltered call is byte-for-byte unchanged.
- **Artifact metadata** records the selected patterns (`allow_patterns`/`ignore_patterns` attrs), so a stored subset (e.g. one quant) is identifiable beyond the source commit.
- **CLI**: `flyte prefetch hf-model` gains `--allow-pattern` / `--ignore-pattern` (repeatable).
- **Sharding** intentionally ignores the patterns (it needs the full weights).
- **Examples**: new `examples/genai/llamacpp/` — the GGUF counterpart to `vllm`/`sglang`: prefetch one quant (`allow_patterns=["*q4_k_m*"]`) → model artifact → serve via the llama.cpp plugin, bound by `ArtifactValue`.

**Backward compatible**: default (`None`) prefetches the whole repo exactly as before — existing callers and the exact-kwargs snapshot test are unaffected.

## How it was tested
- `pytest tests/flyte/prefetch/test_hf_model.py` — 65 passed, 1 skipped. New tests cover: pattern fields default `None` + json round-trip; streaming filters to matching files (incl. nested) and preserves subpaths; snapshot fallback forwards patterns (and omits the kwargs when unset); metadata records the patterns; `hf_model` threads patterns into the info.
- `pytest tests/cli/test_prefetch.py` — 16 passed.
- `ruff check` + `ruff format --check` clean on all changed files.

## Notes
First of a two-PR stack; a follow-up adds a lazy object-store FUSE serving mode for GGUF artifacts (the scale-to-zero-clean delivery that download/`stream_model` don't cover). This PR stands alone.


kumare3 (dismissed):

src/flyte/cli/_prefetch.py
```diff
@@ -87,6 +87,24 @@ def prefetch():
     type=str,
     help="Short description of the model.",
 )
+@click.option(
+    "--allow-pattern",
```
kumare3, line 91: How can we make this simpler

cosmicbboy: suggestion?

kumare3: dont know maybe select the name of the model, lets keep it

PR #1531, opened 2026-09-04
fix(llamacpp): buildable default CUDA image (slim toolkit + nvcc on PATH) + optional NCCL

mhotan:
## Intent
`flyteplugins.llamacpp`'s default `build_llama_cpp_image()` (CUDA mode) does not build on resource-constrained image builders. Both failures were found while driving a real GGUF GPU serve end-to-end on a live control plane; with these fixes the default image compiles `llama-server` and the app serves an OpenAI-compatible `/v1` on a GPU node. Also exposes an optional NCCL build for fast multi-GPU serving.

## What changed
- **Slim the CUDA install** (`_constants.py`). `cuda-toolkit-12-8` is a metapackage that drags in `cuda-nsight-systems` — a GUI profiler pulling Java/GTK/X11 (~2 GB) that a headless build never uses. It bloats the image and OOM-kills constrained builders mid-install. Replaced with the compile-only subset: `cuda-nvcc-12-8 cuda-cudart-dev-12-8 cuda-driver-dev-12-8 libcublas-dev-12-8 libcurand-dev-12-8` (~2.8 GB).
- **Put `nvcc` on `PATH` for the build** (`_image.py`). The image's `PATH` env (which includes `$CUDA_HOME/bin`) is applied *after* the cmake configure/build RUN steps, so `nvcc` isn't on `PATH` during the build and cmake fails with `No CMAKE_CUDA_COMPILER could be found`. Set `PATH=$CUDA_HOME/bin:$PATH` inline on the cmake configure and build commands.
- **Optional NCCL build** (`_image.py`). Add a `nccl: bool = False` parameter to `build_llama_cpp_image()`; when enabled it installs `libnccl-dev libnccl2` and compiles with `-DGGML_CUDA_NCCL=ON`, enabling NCCL for the fastest multi-GPU tensor-parallel serving. Off by default so single-GPU images stay lean.

## How it was tested
Built the default image via `build_llama_cpp_image()` on a live dataplane and deployed a `LlamaCppAppEnvironment` (Qwen3.8-27B GGUF, single L40s). Before: the build failed at ~4 min (OOM during CUDA install) or immediately (`No CMAKE_CUDA_COMPILER`). After: the image compiles and the app reaches a live `/v1` endpoint on the GPU node.

## Note
Separately, `flyteplugins-llamacpp` isn't published to PyPI with these symbols yet, so `build_llama_cpp_image`'s `with_pip_packages("flyteplugins-llamacpp", pre=True)` can't resolve off the shelf — a release is needed for the default image to be buildable without a local wheel.

## Stack
Linear stack, merge in order: `main` ← **#1528** (prefetch `allow_patterns`) ← **#1531** (this PR) ← **#1562** (ModelDeliveryMixin + app) ← **#1563** (sidecar example) ← **#1571** (LLM-as-jury eval) ← **#1569** (union_volume example). Based on #1528.


mhotan:
Draft: needs commit `511b822d5` (`feat(llamacpp): add nccl option to build_llama_cpp_image`) folded in here. It's an image-build change that currently sits on #1569 but belongs in this PR.

PR #1534, opened 2026-09-04
feat(llamacpp): factor build_fserve_command for reuse outside the App

mhotan:
Extract the llama-cpp-fserve argv builder out of LlamaCppAppEnvironment.__post_init__
into a public build_fserve_command(), and have the App call it. This makes the exact
command the App runs reusable for other serving shapes that are not a Flyte App --
notably a llama.cpp server running as a native sidecar in a task pod -- so both shapes
stay in lockstep instead of re-deriving the argv. Behavior for the App is unchanged
(the argv is identical); adds unit tests including an equivalence check.

Co-Authored-By: Claude Opus 4.8 <noreply@anthropic.com>

Signed-off-by: Michael Hotan <mike@union.ai>

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/examples/genai/llamacpp/README.md
+++ b/examples/genai/llamacpp/README.md
@@ -77,3 +77,8 @@ The same prefetched model can reach the server two ways, set by `model_delivery`
   `--jinja` for tool-calling, `--flash-attn`). See the
   [llama-server docs](https://github.com/ggml-org/llama.cpp/tree/master/tools/server).
 - **Speculative decoding.** Point `draft_model_hf_path` at a small draft GGUF (see the plugin README).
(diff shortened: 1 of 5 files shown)


mhotan:
Superseded by #1562 (feat(app): ModelDeliveryMixin + llama.cpp fuse serving). The fuse-delivery mode was reworked around a reusable `ModelDeliveryMixin`, and this PR's content is folded into #1562 (validated E2E live on AWS Mountpoint-S3 + GCP gcsfuse). Closing to consolidate.

PR #1550, opened 2026-09-07
Add Slurm connector plugin

davidmirror-ops:
Run Flyte 2 tasks on an existing Slurm cluster, including Soperator-managed clusters, without changing the cluster.

Two task types are served by one connector:

- `slurm`: a Python task submitted as an sbatch job that runs the task's own container image and Flyte entrypoint under Pyxis/Enroot, so typed I/O, caching, retries and error reporting work as they do for a Kubernetes pod. Removing `plugin_config` runs the identical task on Kubernetes.
- `slurm_script`: an existing sbatch script submitted unchanged; scalar inputs are exposed as FLYTE_INPUT_<NAME>. Phase, exit code and logs only.

Transport is SSH to a login node (asyncssh) behind a small SlurmTransport protocol so a slurmrestd transport can be added later. One connection per cluster is reused and status queries are batched (squeue, then sacct for finished jobs). Slurm states map to Flyte phases with PENDING reported as QUEUED rather than RUNNING, PREEMPTED as retryable, and CANCELLED as ABORTED; failures carry Slurm's reason and the tail of stderr.

Connection details may be set per task or cluster-wide on the connector via FLYTE_SLURM_HOST / FLYTE_SLURM_USERNAME / FLYTE_SLURM_SSH_PRIVATE_KEY. The plugin is added to the default connector image build.

Claude-Session: https://claude.ai/code/session_01X4XZW8jVev8ueGeDPfEiWx

### DIFF
--- a/plugins/slurm/README.md
+++ b/plugins/slurm/README.md
@@ -1,0 +1,156 @@
+# Slurm Plugin for Flyte
+
+Run Flyte 2 tasks on an existing Slurm cluster. Jobs are submitted over SSH to a
+login node, so this works with any Slurm installation — including clusters managed
+by [Soperator](https://github.com/nebius/soperator) — without changing the cluster.
+
+## Installation
+
+```bash
+pip install flyteplugins-slurm
+```
+
+The connector must also be installed in the `flyteconnector` image of your Flyte
+dataplane, and the `slurm` and `slurm_script` task types routed to the connector
+service. See the deployment section below.
+
+## Task types
+
+| Task type | What is submitted | Typed I/O | Caching |
+|---|---|---|---|
+| `slurm` | The task's own container image and Flyte entrypoint, via Pyxis/Enroot | Yes | Yes |
+| `slurm_script` | A user-supplied `sbatch` script, as-is | No — phase, exit code and logs | No |
+
+## Python tasks on Slurm
+
+```python
+import flyte
+from flyteplugins.slurm import Slurm
+
+env = flyte.TaskEnvironment(
+    name="train",
+    plugin_config=Slurm(
+        partition="main",
+        nodes=1,
+        gres="gpu:8",
+        time_limit="4:00:00",
+        # Where the Flyte entrypoint finds object storage from inside the job:
+        env={"AWS_ENDPOINT_URL": "https://storage.example", "AWS_REGION": "eu-north1"},
+        # Connection — or set FLYTE_SLURM_HOST / FLYTE_SLURM_USERNAME /
+        # FLYTE_SLURM_SSH_PRIVATE_KEY on the connector and omit these:
+        host="login.slurm.example.com",
+        username="flyte",
+        ssh_private_key="slurm-ssh-key",  # name of a Flyte secret
+    ),
+    image=flyte.Image.from_debian_base().with_pip_packages("flyteplugins-slurm"),
+)
+
+@env.task
+async def train(steps: int) -> float:
+    ...
+```
+
+Delete `plugin_config` and the same task runs as a Kubernetes pod. Nothing else changes.
+
+The job runs `srun --container-image=<task image> a0 ...` inside an `sbatch` allocation
+described by the `Slurm` fields. Anything `sbatch` accepts that is not a first-class
+field goes in `sbatch_options`:
+
+```python
+Slurm(partition="main", sbatch_options={"exclusive": True, "mail-type": "FAIL"})
+```
+
+Do not set `resources` on a Slurm task environment. The allocation is described by the
+`Slurm` config and granted by Slurm, not by Kubernetes.
+
+### Container images
+
+Pyxis pulls the task image from its registry and caches it on the cluster. On clusters
+where images are pre-imported to the shared filesystem, point at the squashfs directly:
+
+```python
+Slurm(container_image="/jail/images/train.sqsh", ...)
+```
+
+### Object storage from inside the job
+
+The Flyte entrypoint inside the job reads inputs and writes outputs to the run's object
+storage. Slurm worker nodes need network access to that storage and credentials for it.
+Pass endpoint and credential *references* through `env`; keep secrets on the cluster
+(for example via instance credentials or a credentials file on the shared filesystem),
+not in the task config.
+
+## Existing sbatch scripts
+
+```python
+from flyteplugins.slurm import Slurm, SlurmScriptTask
+
+legacy_train = SlurmScriptTask(
+    name="legacy_train",
+    script=open("train.sbatch").read(),
+    plugin_config=Slurm(partition="main", host=..., username=..., ssh_private_key=...),
+    inputs={"epochs": int},
+)
+```
+
+The script is submitted unchanged. Scalar inputs are exported as `FLYTE_INPUT_<NAME>`
+environment variables. Our `#SBATCH` directives come first, so the script's own
+directives are kept but overridden where they conflict.
+
+## How states map
+
+| Slurm | Flyte |
+|---|---|
+| `PENDING`, `CONFIGURING`, `REQUEUED`, `SUSPENDED` | `QUEUED` — does not count as running |
+| `RUNNING`, `COMPLETING` | `RUNNING` |
+| `COMPLETED` | `SUCCEEDED` |
+| `FAILED`, `NODE_FAIL`, `OUT_OF_MEMORY`, `TIMEOUT`, `DEADLINE`, `BOOT_FAIL` | `FAILED` — with Slurm's reason and the tail of stderr |
+| `PREEMPTED` | `RETRYABLE_FAILED` |
+| `CANCELLED` | `ABORTED` |
+
+Aborting the Flyte run runs `scancel`.
+
+## Deployment
+
+1. **Connector image.** Add `flyteplugins-slurm` to the `flyteconnector` image
+   (`maint_tools/build_default_image.py` in this repo builds the default one).
+2. **Routing.** In the dataplane values:
+   ```yaml
+   union:
+     configmap:
+       enabled_plugins:
+         tasks:
+           task-plugins:
+             default-for-task-types:
+               slurm: connector-service
+               slurm_script: connector-service
+   ```
+3. **Credentials.** The SSH private key is a Flyte secret named in
+   `Slurm.ssh_private_key`, or set cluster-wide as `FLYTE_SLURM_SSH_PRIVATE_KEY` on
+   the `flyteconnector` deployment. Provide a `known_hosts` file for host-key
+   verification; `skip_host_key_verification=True` exists for development only.
+4. **Network.** The connector needs to reach the login node on its SSH port. Restrict
+   the login node's source ranges to the connector's egress addresses.
+
+The connector keeps one SSH connection per cluster and batches status queries, so it
+opens one login-node session per poll, not one per job.
+
+## Connector-level defaults
+
+| Variable | Purpose |
+|---|---|
+| `FLYTE_SLURM_HOST` | Login node when the task config leaves `host` unset |
+| `FLYTE_SLURM_PORT` | SSH port (default 22) |
+| `FLYTE_SLURM_USERNAME` | SSH user |
+| `FLYTE_SLURM_SSH_PRIVATE_KEY` | Private key contents |
+| `FLYTE_SLURM_KNOWN_HOSTS` | Path to a known_hosts file on the connector |
+| `FLYTE_SLURM_WORKING_DIR` | Directory for scripts and logs (default `.flyte/jobs` under the user's home) |
+
+## Not yet supported
+
+- Multi-node gang jobs with a distributed launcher (`nodes > 1` allocates, but the
+  entrypoint runs as a single process).
+- A slurmrestd transport. The SSH transport is behind a small protocol
+  (`flyteplugins.slurm.transport.SlurmTransport`) so one can be added without touching
+  the connector.
+- Per-user job attribution; jobs run as the configured SSH user.
--- a/plugins/slurm/src/flyteplugins/slurm/connector.py
+++ b/plugins/slurm/src/flyteplugins/slurm/connector.py
@@ -1,0 +1,303 @@
+import hashlib
+import os
+import posixpath
+import re
+import uuid
+from dataclasses import dataclass
+from datetime import datetime, timezone
+from typing import Any, AsyncIterator, Dict, Optional
+
+from flyte import system_logger as logger
+from flyte.connectors import AsyncConnector, ConnectorRegistry, Resource, ResourceMeta
+from flyteidl2.connector.connector_pb2 import GetTaskLogsResponse, GetTaskLogsResponseBody, TaskExecutionMetadata
+from flyteidl2.core.execution_pb2 import TaskExecution, TaskLog
+from flyteidl2.core.tasks_pb2 import TaskTemplate
+from google.protobuf.json_format import MessageToDict
+from google.protobuf.timestamp_pb2 import Timestamp
+
+from flyteplugins.slurm.script import render_container_job, render_script_job
+from flyteplugins.slurm.transport import SlurmJobState, SlurmTransport, SSHTransport
+
+TASK_TYPE_NATIVE = "slurm"
+TASK_TYPE_SCRIPT = "slurm_script"
+
+# Connector-level defaults so a platform team can configure the cluster once on the
+# flyteconnector deployment instead of on every task.
+ENV_HOST = "FLYTE_SLURM_HOST"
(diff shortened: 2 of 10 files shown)


EngHabu:
Do we want to use SSH or the API-based thing? Not sure if the v1 supports both: https://github.com/flyteorg/flytekit/tree/d69b3fbd5ffe2a94d47e4fdba98200ce8b13e49c/plugins/flytekit-slurm/flytekitplugins/slurm

davidmirror-ops:
> Do we want to use SSH or the API-based thing? Not sure if the v1 supports both: https://github.com/flyteorg/flytekit/tree/d69b3fbd5ffe2a94d47e4fdba98200ce8b13e49c/plugins/flytekit-slurm/flytekitplugins/slurm

Seems like the V1 connector was also SSH only. 
Using the REST client (`slurmrestd`) is not supported in this plugin yet. The Nebius soperator cluster I'm using for testing doesn't even ship with that component installed and I haven't tested if it can actually issue auth tokens. But I agree is the scalable path. For now I added a known gaps section to the readme and the goal is to have some initial feedback from customers to keep refining.

plugins/slurm/src/flyteplugins/slurm/connector.py
```diff
@@ -0,0 +1,367 @@
...
+        output_prefix: str,
+        inputs: Optional[Dict[str, Any]] = None,
+        task_execution_metadata: Optional[TaskExecutionMetadata] = None,
+        ssh_private_key: Optional[str] = None,
+        **kwargs,
+    ) -> SlurmJobMetadata:
+        custom = MessageToDict(task_template.custom)
+        connection = custom.get("connection") or {}
+
+        host = connection.get("host") or os.getenv(ENV_HOST)
```
samhita-alla: If the connector has FLYTE_SLURM_HOST set, a task shouldn't be able to point the shared key at a different host, especially combined with `skip_host_key_verification`. Suggest connector env wins when present, and `skip_host_key_verification` comes from env only.

plugins/slurm/src/flyteplugins/slurm/script.py
```diff
@@ -0,0 +1,216 @@
...
+        "output": stdout_path,
+        "error": stderr_path,
+    }
+    for field, flag in _FIELD_TO_SBATCH.items():
+        value = fields.get(field)
+        if value is not None:
+            options[flag] = value
+    if extra:
+        for key, value in extra.items():
+            options[str(key)] = value
```
samhita-alla: This lets `sbatch_options` override `--output` and `--error`. A task that prints an ssh public key and sets `output=~/.ssh/authorized_keys` gets a shell as the shared user. Reserve job-name, output, error, chdir, wrap, uid, gid here, and reject values containing whitespace.

plugins/slurm/src/flyteplugins/slurm/connector.py
```diff
@@ -0,0 +1,367 @@
...
+                command=[*container.command, *container.args],
+                env=env,
+                sbatch_fields=sbatch_fields,
+                sbatch_extra=sbatch_extra,
+                container_mounts=container_cfg.get("mounts") or [],
+                container_workdir=container_cfg.get("workdir"),
+                srun_extra_args=container_cfg.get("srun_args") or [],
+            )
+
+        logger.debug(f"Submitting Slurm job {meta.job_name} to {username}@{host}:\n{script}")
```
samhita-alla: This writes every exported env var (`_U_RUN_BASE`, plus whatever the deployment injects) into connector logs. Log the job name and script path instead of the body.

plugins/slurm/src/flyteplugins/slurm/script.py
```diff
@@ -0,0 +1,216 @@
...
+) -> str:
+    """Render an sbatch script that runs `command` inside `image` via Pyxis.
+
+    This is the native-task path: `command` is the task's own Flyte
+    entrypoint, so the job behaves exactly like the equivalent Kubernetes pod.
+    """
+    if not command:
+        raise ValueError("A container job needs a non-empty command")
+
+    srun = ["srun", f"--container-image={pyxis_image_ref(image)}"]
```
samhita-alla: Two things on this line. With `nodes=2` and no `ntasks`, srun defaults to one task per node, so the entrypoint runs twice against the same output prefix. Pin `--nodes=1 --ntasks=1` on this srun, or reject `nodes>1` for `slurm` tasks until there's a launcher. Second, this is where the Pyxis assumption lives. A `container_runtime` option that emits `apptainer exec docker://<image> ...` instead would make the native task usable on most non-Nebius clusters.

plugins/slurm/src/flyteplugins/slurm/connector.py
```diff
@@ -0,0 +1,367 @@
...
+        logger.info(f"Submitted Slurm job {meta.job_id} ({meta.job_name}) to {host}")
+        return meta
+
+    async def get(self, resource_meta: SlurmJobMetadata, ssh_private_key: Optional[str] = None, **kwargs) -> Resource:
+        transport = self._transport_for_meta(resource_meta, ssh_private_key)
+        states = await transport.status([resource_meta.job_id])
+        state = states.get(resource_meta.job_id)
+        if state is None:
+            raise RuntimeError(
+                f"Slurm job {resource_meta.job_id} ({resource_meta.job_name}) is not known to squeue or sacct "
```
samhita-alla, line 338: If sacct isn't enabled, or the job aged out of squeue before the next poll, a successful job fails here. Fall back to `scontrol show job`, and have the script write `$?` to a `.rc` file next to `.out` as the last resort. That combination works on every cluster I've seen.

plugins/slurm/src/flyteplugins/slurm/transport.py
```diff
@@ -0,0 +1,247 @@
...
+        return parse_sbatch_job_id(stdout)
+
+    async def status(self, job_ids: Iterable[str]) -> Dict[str, SlurmJobState]:
+        ids = sorted({str(j) for j in job_ids})
+        if not ids:
+            return {}
+        joined = ",".join(ids)
+        # squeue is authoritative for jobs still in the system; sacct fills in the ones that finished.
+        # squeue exits non-zero when none of the ids exist, which is not an error for us.
+        stdout, _, _ = await self._run(f"squeue -h -j {joined} -o {_q(_SQUEUE_FORMAT)}", check=False)
```
samhita-alla: slurmctld being down looks the same as "job finished" here, since rc and stderr are dropped. Check rc and raise with stderr.

plugins/slurm/src/flyteplugins/slurm/transport.py
```diff
@@ -0,0 +1,247 @@
...
+                    client_keys=[key],
+                    known_hosts=known_hosts,
+                ),
+                timeout=self._connect_timeout,
+            )
+            return self._conn
+
+    async def _run(self, command: str, check: bool = True) -> Tuple[str, str, int]:
+        conn = await self._connection()
+        result = await conn.run(command, check=False)
```
samhita-alla: No timeout on the remote command. A hung login node blocks the poll forever. `asyncio.wait_for` with something like 60s.

plugins/slurm/src/flyteplugins/slurm/connector.py
```diff
@@ -0,0 +1,367 @@
...
+    if base in _SUCCEEDED:
+        return TaskExecution.SUCCEEDED
+    if base in _RETRYABLE:
+        return TaskExecution.RETRYABLE_FAILED
+    if base in _ABORTED:
+        return TaskExecution.ABORTED
+    if base in _FAILED:
+        return TaskExecution.FAILED
+    logger.warning(f"Unrecognized Slurm job state {state!r}; treating as RUNNING")
+    return TaskExecution.RUNNING
```
samhita-alla: An unknown state polls until the task's own timeout. Safer to fail with the raw state string in the message.

plugins/slurm/src/flyteplugins/slurm/task.py
```diff
@@ -0,0 +1,179 @@
...
+@dataclass
+class Slurm:
+    """Configuration for running a task on a Slurm cluster.
+
+    Scheduling fields map one-to-one onto `sbatch` options; anything not covered
+    goes in `sbatch_options` verbatim. Connection fields may be left unset and
+    supplied cluster-wide on the connector via `FLYTE_SLURM_HOST`,
+    `FLYTE_SLURM_USERNAME` and `FLYTE_SLURM_SSH_PRIVATE_KEY` instead.
+
+    Do not set `resources` on a Slurm task environment: the allocation is
```
samhita-alla, line 39: `Resources` is silently dropped today. Either map cpu/mem/gpu to `--cpus-per-task/--mem/--gres` with the Slurm fields overriding, or raise at serialization. Silent is the one option that'll bite people.

plugins/slurm/src/flyteplugins/slurm/connector.py
```diff
@@ -0,0 +1,367 @@
...
+    for key, value in (inputs or {}).items():
+        if isinstance(value, (str, int, float, bool)):
+            name = "FLYTE_INPUT_" + re.sub(r"[^A-Za-z0-9_]", "_", key).upper()
+            env[name] = str(value).lower() if isinstance(value, bool) else str(value)
+        else:
+            # An environment variable can only carry a scalar. Say so rather than
+            # leaving the author to discover an unset variable inside the script.
+            logger.warning(
+                f"Input {key!r} of type {type(value).__name__} cannot be exported to an sbatch "
+                f"script; only str, int, float and bool become FLYTE_INPUT_* variables. "
```
samhita-alla: This warning only lands in connector logs, so users won't see it. For `File`/`Dir` inputs export the URI; for anything else fail the task with this message.

plugins/slurm/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+The connector must also be installed in the `flyteconnector` image of your Flyte
+dataplane, and the `slurm` and `slurm_script` task types routed to the connector
+service. See the deployment section below.
+
+## Task types
+
+| Task type | What is submitted | Typed I/O | Caching |
+|---|---|---|---|
+| `slurm` | The task's own container image and Flyte entrypoint, via Pyxis/Enroot | Yes | Yes |
```
samhita-alla, line 21: Say up front that `slurm` needs Pyxis/Enroot on the cluster, and what to do if it isn't there. Right now someone on an Apptainer cluster finds out from an srun error.

plugins/slurm/README.md
```diff
@@ -0,0 +1,169 @@
...
+| `FLYTE_SLURM_HOST` | Login node when the task config leaves `host` unset |
+| `FLYTE_SLURM_PORT` | SSH port (default 22) |
+| `FLYTE_SLURM_USERNAME` | SSH user |
+| `FLYTE_SLURM_SSH_PRIVATE_KEY` | Private key contents |
+| `FLYTE_SLURM_KNOWN_HOSTS` | Path to a known_hosts file on the connector |
+| `FLYTE_SLURM_WORKING_DIR` | Directory for scripts and logs (default `.flyte/jobs` under the user's home) |
+
+## Not yet supported
+
+- Multi-node gang jobs with a distributed launcher (`nodes > 1` allocates, but the
```
samhita-alla: Not accurate today, see the srun comment. Also mention that PREEMPTED only re-submits when the task sets `retries > 0`.

plugins/slurm/tests/test_slurm.py
```diff
@@ -0,0 +1,605 @@
...
+
+class TestConnectorRegistration:
+    def test_both_task_types_registered(self):
+        assert isinstance(ConnectorRegistry.get_connector("slurm"), SlurmConnector)
+        assert isinstance(ConnectorRegistry.get_connector("slurm_script"), SlurmConnector)
+        names = {c.name for c in ConnectorRegistry._list_connectors()}
+        assert "Slurm Connector" in names
+
+
+class _FakeTransport:
```
samhita-alla, line 424: Everything below `_FakeTransport` bypasses asyncssh, so submit/status/cancel/tail and the squeue-to-sacct fallback have no coverage. A fake `conn.run` over `SSHTransport` would test the part most likely to break on a real cluster.

plugins/slurm/src/flyteplugins/slurm/connector.py
```diff
@@ -0,0 +1,367 @@
...
+            )
+
+        phase = slurm_state_to_phase(state.base_state)
+
+        # The job's stdout and stderr are files on the login node, not resources behind a
+        # URL. A TaskLog uri renders as a hyperlink, so returning a POSIX path there makes a
+        # dead link in the UI; name the paths in the message instead. Streaming stdout is
+        # `get_logs`' job. Kept ahead of the stderr tail so that block stays last.
+        message = (
+            f"{_describe(state)}\n"
```
samhita-alla, line 349: Since this message already names the host and job files, it's a good place for a debug entry point while the job is RUNNING. Slurm users get interactive access for free, so something like:

```
ssh flyte@login.example.com 'srun --jobid=12345 --overlap --pty bash'
```

gives them a shell in the running allocation straight from the Flyte UI, without us building a tunnel. Flyte's own debug SSH won't work here (the entrypoint would start the server, but nothing routes to a Slurm node), so this is the practical equivalent for the `slurm` task type.

davidmirror-ops:
Thank you @samhita-alla this is great feedback. I'm incorporating it and testing again. I'll add also a known gaps section to the docs. Goal for now is to get early feedback from customers

plugins/slurm/src/flyteplugins/slurm/script.py
```diff
@@ -0,0 +1,297 @@
...
+    if runtime == "pyxis":
+        parts = [f"--container-image={pyxis_image_ref(image)}"]
+        if mounts:
+            parts.append(f"--container-mounts={','.join(mounts)}")
+        if workdir:
+            parts.append(f"--container-workdir={workdir}")
+        return parts
+
+    if runtime == "apptainer":
+        parts = ["apptainer", "exec"]
```
samhita-alla: Apptainer needs `--nv` to expose NVIDIA GPUs inside the container, and nothing adds it here. `srun_args` can't fix it either since they land before `apptainer exec`. So a `gres="gpu:8"` task on an Apptainer cluster gets the allocation and then runs without CUDA.

Suggest adding `--nv` when `gres` or `gpus_per_node` is set, plus an `apptainer_args` list for site-specific flags (`--rocm`, `--cleanenv`, extra binds).

plugins/slurm/src/flyteplugins/slurm/connector.py
```diff
@@ -0,0 +1,399 @@
...
+        return TaskExecution.SUCCEEDED
+    if base in _RETRYABLE:
+        return TaskExecution.RETRYABLE_FAILED
+    if base in _ABORTED:
+        return TaskExecution.ABORTED
+    if base in _FAILED:
+        return TaskExecution.FAILED
+    # Polling on would run until the task's own timeout with nothing explaining why, and
+    # an unknown state is more often terminal than not.
+    raise ValueError(
```
samhita-alla: Raising here turns into a gRPC INTERNAL, and the backend handles that badly. In flyte2's webapi cache (`pluginmachinery/internal/webapi/cache.go`), each failed Get bumps `SyncFailureCount`, and after `MaxSystemFailures` (5, polled every 30s) it stops calling Get. If the job was already seen RUNNING, the monitor keeps reporting that stale RUNNING until the task times out. If it was never seen, it fails with a generic cache error.

Returning `Resource(phase=FAILED, message=f"Unrecognized Slurm job state {state}")` gives a clear failure on the first poll instead.

plugins/slurm/src/flyteplugins/slurm/task.py
```diff
@@ -0,0 +1,207 @@
...
+        self,
+        name: str,
+        script: str,
+        plugin_config: Slurm,
+        inputs: Optional[Dict[str, Type]] = None,
+        **kwargs,
+    ):
+        super().__init__(
+            name=name,
+            interface=NativeInterface({k: (v, None) for k, v in inputs.items()} if inputs else {}, {}),
```
samhita-alla: Script tasks having no outputs is the biggest gap for the "wrap your existing sbatch scripts" story, since a script can't feed the next step. The connector can fill this in without anything running inside the job:

1. Let `SlurmScriptTask` declare outputs, like v1's `output_locs`.
2. In `create`, export `FLYTE_OUTPUT_<NAME>` per output. For `File`/`Dir` that's a URI under `output_prefix`; for scalars, a file in the job dir.
3. On SUCCEEDED, `get` reads the scalar files over SSH and returns them plus the File/Dir URIs as `Resource.outputs`.

The script just writes to the paths it's given. Fine as a fast follow if you'd rather not grow this PR.

PR #1551, opened 2026-09-08
fix: normalize Secret mount path validation and hash stability across…

aayuxsh326:
## Description
This PR addresses a cross-platform bug in `flyte._secret.Secret` when validating secret mount paths and generating deterministic hashes on Windows.

### Issues Fixed
1. **Windows Mount Path Validation Failure**: On Windows, `pathlib.Path("/etc/flyte/secrets")` formats path separators as `\etc\flyte\secrets` when converted using `str()`. Comparing `str(self.mount) != "/etc/flyte/secrets"` raised `ValueError: Only /etc/flyte/secrets is supported as secret mount path today.` on Windows systems even for valid paths.
2. **Non-deterministic `stable_hash` Across Platforms**: `Secret.stable_hash()` used `str(self.mount)`. On Windows, the backslash representation produced different SHA-256 hashes than on Linux/macOS for identical secret configurations.
3. **Mount Type Flexibility**: Extended `mount` typing to `pathlib.Path | str | None = None` and ensured string inputs are automatically parsed and normalized.
4. **Broken `__main__` Example**: Updated line 94 in `_secret.py` where `/path/to/secret` was used (which caused a runtime crash when executing the file directly).

### Changes Made
- Modified `src/flyte/_secret.py`:
  - Normalized `self.mount` using `.as_posix()` for path comparison.
  - Used `self.mount.as_posix()` in `stable_hash()` for platform-independent hash calculation.
  - Allowed `mount` to accept `str` paths in addition to `pathlib.Path`.
  - Fixed the test example in `__main__`.
- Modified `tests/user_api/test_secret.py`:
  - Added tests for `mount` supplied as `str`.
  - Added assertion verifying POSIX normalization on `secret.mount`.
  - Added cross-platform test verifying hash stability between `str` and `Path`.

## How Has This Been Tested?
- Unit tests added to `tests/user_api/test_secret.py`.
- Verified on Windows that `Secret(key="my-secret", mount=pathlib.Path("/etc/flyte/secrets"))` and string inputs pass without error.
- Verified that hash outputs match between POSIX and Windows path representations.


kumare3 (dismissed):

aayuxsh326:
Hi @kumare3, I've updated this PR to resolve the mypy and ty check failures. I replaced the | type hint syntax with typing.Union to ensure backward compatibility with Python 3.8/3.9, and I also added the missing DCO sign-offs to the commits. Everything should be passing now. I'd appreciate it if you could give this another review!

PR #1552, opened 2026-09-08
fix: ensure storage.join uses forward slashes for remote URIs

aayuxsh326:
## Description
This PR resolves two bugs in `flyte.storage.join()` when handling remote cloud storage URIs.

### Problems Fixed
1. **Windows Path Corruption for Cloud URIs**: `storage.join()` previously relied solely on `os.path.join()`. On Windows, `storage.join("s3://bucket", "folder", "file.txt")` inserted backslashes (`s3://bucket\folder\file.txt`), corrupting cloud URIs and causing object storage operations to fail.
2. **Leading Slash Discards Base Prefix**: Under standard Python `os.path.join()`, any argument with a leading slash (e.g., `storage.join("s3://bucket/dir", "/subfile.txt")`) was treated as an absolute filesystem root, silently discarding `"s3://bucket/dir"` and returning `"/subfile.txt"`.
3. **Empty Call**: Calling `storage.join()` without arguments now cleanly returns `""`.

### Changes Made
- Modified `src/flyte/storage/_storage.py`:
  - Updated `join(*paths: str)`: when `paths[0]` is a remote URI (checked via `is_remote(first)` or contains `"://"`), path components are joined using forward slashes (`/`), stripping redundant slashes from inner segments to prevent prefix dropping.
  - Preserved standard OS filesystem joining for local paths.
- Modified `tests/internal/storage/test_storage.py`:
  - Added `test_storage_join()` covering remote cloud schemes (`s3://`, `gs://`, `flyte://`), leading-slash subpaths, empty calls, and local paths.

## How Has This Been Tested?
- Unit tests added to `tests/internal/storage/test_storage.py`.
- Verified that remote URIs join cleanly with `/` across platforms without dropping prefix segments.


kumare3 (dismissed):

aayuxsh326:
Hi @kumare3, I've pushed a new commit to fix the make fmt check failures (split a long assertion line to adhere to ruff's line length limits) and I've also rebased to ensure all commits have the proper DCO sign-offs. The CI checks should all be green now. Could you please take another look when you have a moment? Thank you!

kumare3:
@aayuxsh326 can you please fix the linter

PR #1556, opened 2026-09-09
examples: queue resource cap example

SVilgelm:
Adds `examples/queues/resource_cap.py`, covering a queue's `max_resources` cap the way the other examples here cover the integer limits.

`main` submits six 200Mi actions to a queue capped at 600Mi: three run at a time, in two waves. `too_big` submits one 200Mi action to a queue capped at 64Mi, which cannot fit even an idle queue, so it is failed as UNSCHEDULABLE before dispatch rather than held for room that can never appear.

The docstring carries the expected timing for each half, the commands to watch it from the scheduler while it runs, and how to create the two queues.

### DIFF
--- a/examples/queues/resource_cap.py
+++ b/examples/queues/resource_cap.py
@@ -1,74 +1,28 @@
 """
 Queue resource caps: holding work on the summed request vector, not a count.
 
-`caps-mem` is configured with `maxResources: {memory: 600Mi}` and NO integer
-caps, so the only thing that can hold an action is its memory request. Each
-step asks for 200Mi, so three fit and the rest wait for a predecessor to
-finish — the same shape as small_concurrency.py, measured in memory instead
-of actions.
-
-Run it:
+`main` submits six 200Mi steps to `caps-mem`, which caps memory at 600Mi and
+carries no integer caps. Three run at a time; the rest wait for a predecessor
+to release its request.
 
     flyte run examples/queues/resource_cap.py main
 
-Expected timing for `count=6, sleep_seconds=4` and 600Mi / 200Mi = 3:
-ceil(6/3) * 4s = 8s wall time. Under ~6s means the cap is not being enforced;
-far over ~12s means the queue is under-scheduling (a held action should be
-re-picked the tick after its predecessor's demand is released, ~1s).
-
-Watch it from the leasor while it runs:
-
-    curl -s localhost:10254/debug/leasor/queues | jq '.queues[]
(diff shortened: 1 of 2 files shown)


examples/queues/resource_cap.py
```diff
@@ -0,0 +1,123 @@
...
+Run it:
+
+    flyte run examples/queues/resource_cap.py main
+
+Expected timing for `count=6, sleep_seconds=4` and 600Mi / 200Mi = 3:
+ceil(6/3) * 4s = 8s wall time. Under ~6s means the cap is not being enforced;
+far over ~12s means the queue is under-scheduling (a held action should be
+re-picked the tick after its predecessor's demand is released, ~1s).
+
+Watch it from the leasor while it runs:
```
wild-endeavor: let's not include this section.

examples/queues/resource_cap.py
```diff
@@ -0,0 +1,123 @@
...
+
+    curl -s localhost:10254/debug/leasor/queues | jq '.queues[]
+      | select(.Name=="caps-mem") | {max_resources, in_flight, Depth, ActiveActions}'
+
+    curl -s localhost:10254/metrics | grep -E \
+      'queue_resource_(max|in_flight)|schedule_skip_total.*queue_at_resource_cap'
+
+`in_flight.MemoryBytes` should sit at 629145600 (3 x 200Mi) while the queue is
```
wild-endeavor: skip this also

wild-endeavor (commented):
remove part of the comment?  i think we should maybe just redirect the user to the `--watch` on the queue object in flyteplugins-union... but just redirect there, we should be adding docs in our main docs repo. claude needs to stop adding these massive comment blocks here.

examples/queues/resource_cap.py
SVilgelm: Removed in 1b22ce9d.

examples/queues/resource_cap.py
SVilgelm: Removed in 1b22ce9d.

SVilgelm:
Cut the docstring from 123 lines to 75 in 1b22ce9d: both curl blocks and their interpretation are gone, replaced by a pointer to `flyte get queue caps-mem --watch`. Also dropped the timing-diagnostic arithmetic, the cap-retuning notes and the unconfigured-cap caveat — same category, belongs in the docs repo. What's left is what each half does, the two run commands, and the queue creation the example depends on.

wild-endeavor (approved):

PR #1558, opened 2026-09-09
Jan/add dbt plugin

fiedlerNr9:
## Summary

https://github.com/user-attachments/assets/b4b3f010-9192-41b5-b965-37b55d55ff12

Adds a new `flyteplugins-dbt` package for running dbt CLI invocations as Flyte v2 tasks.

The new `DbtTask` maps one `dbtRunner.invoke(cli_args)` call to one Flyte task. CLI arguments are modeled as task inputs so users can define a reusable dbt task and provide exact dbt command arguments at invocation time.

## What changed

- Added `plugins/dbt` package with `flyteplugins.dbt`.
- Added `DbtTask`, which executes dbt through `dbt.cli.main.dbtRunner`.
- Added `DbtNodeResult`, a small serializable summary of dbt node results.
- Added remote task reconstruction support through `DbtTaskResolver`.
- Added default dbt event callback that records finished dbt nodes as Flyte traces.
- Added support for user-provided dbt event callbacks.
- Documented local and remote callback usage, including import-path callback support.
- Added unit tests covering task interface, resolver behavior, dbt result summarization, failure handling, tracing callback behavior, and custom callbacks.

## Notes

The plugin depends on `dbt-core`. Users still need to install the dbt adapter required by their project, such as `dbt-duckdb`, `dbt-bigquery`, or `dbt-snowflake`, in their task image.

For remote execution, custom callbacks must be importable functions. The task accepts either callable functions or import-path strings.

## Testing

```text
cd plugins/dbt
uv run python -m pytest tests
```

```text
14 passed
```

```text
ruff check plugins/dbt
All checks passed
```


cosmicBboy:
this looks good @fiedlerNr9 ! would you mind added a few examples in an ./examples subdirectory?

plugins/dbt/src/flyteplugins/dbt/task.py
```diff
@@ -0,0 +1,128 @@
...
+        from flyte._utils.asyncify import run_sync_in_thread
+
+        return await run_sync_in_thread(
+            invoke_dbt,
+            kwargs["cli_args"],
+            self.callbacks,
+            include_default_callback=self.include_default_callback,
+        )
+
+    def container_args(self, serialize_context: SerializationContext) -> list[str]:
```
kumare3: why do we need to override it, can we not set the resolver at the task level?

kumare3: This is a little fragile as it can break if we change the api

fiedlerNr9:
> this looks good @fiedlerNr9 ! would you mind added a few examples in an ./examples subdirectory?

Sure, I added the jaffle example using duckdb as an example ✅

plugins/dbt/src/flyteplugins/dbt/task.py
fiedlerNr9: I made the dbt task a subclass of AsyncFunctionTaskTemplate. This way we dont have to define container args etc.

Is this better? Its all in this commit 

https://github.com/flyteorg/flyte-sdk/pull/1558/changes/8fb8deab27da1a4f76222ec3c9871212f95d4add

plugins/dbt/src/flyteplugins/dbt/runner.py
```diff
@@ -0,0 +1,235 @@
...
+    )
+
+
+def callback_import_path(callback: DbtEventCallback) -> str:
+    module = getattr(callback, "__module__", None)
+    name = getattr(callback, "__name__", None)
+    qualname = getattr(callback, "__qualname__", None)
+    if not module or not qualname or "<locals>" in qualname or name == "<lambda>":
+        raise ValueError("dbt callbacks used in DbtTask must be importable functions when running remotely.")
+    return f"{module}.{qualname}"
```
samhita-alla, line 165: This builds the path from the raw `__module__`, so a callback defined in a script you run with `python` serializes as `__main__.<name>`. In the task container `__main__` is the flyte entrypoint, so `import_callback` fails with `AttributeError: module '__main__' has no attribute 'print_dbt_event'` before dbt starts.

The example in this PR is fine, because `dbt_build` is submitted from inside `main`, and by then the script has been imported under its real module name. It breaks when the dbt task is the root of the run:

```python
# python run_dbt.py
def print_dbt_event(event): ...

dbt_build = DbtTask(name="dbt-build", task_environment=env, callbacks=[print_dbt_event])

if __name__ == "__main__":
    flyte.init_from_config()
    flyte.run(dbt_build, cli_args=["build", ...])
```

Here the resolver args come out as `callbacks __main__.print_dbt_event`. The SDK already handles this for tasks in `extract_obj_module` (`src/flyte/_module.py`), which maps `__main__` to the script's file stem, and `loader_args` already gets the `root_dir` it needs.

plugins/dbt/src/flyteplugins/dbt/task.py
```diff
@@ -0,0 +1,75 @@
...
+        include_default_callback: bool = True,
+        **kwargs: Any,
+    ):
+        self.callbacks = list(callbacks or [])
+        self.include_default_callback = include_default_callback
+        from flyteplugins.dbt.resolver import DbtTaskResolver
+
+        super().__init__(
+            name=name,
+            func=_dbt_task_placeholder,
```
samhita-alla, line 44: `cache="auto"` won't behave as expected on this task type. The cache version comes from hashing the source of `task.func` (`FunctionBodyPolicy`), and here that's always `_dbt_task_placeholder`, so every `DbtTask` gets the same version and it never changes when the dbt project does. The cache key is inputs + task name + interface + that version, so the SQL never participates.

So with `DbtTask(name="dbt-build", cache="auto", ...)`, if you edit a model, rebuild the image, and rerun with the same `cli_args`, you get a cache hit and dbt doesn't run. Two unrelated `DbtTask`s both came out as version `6f633ec4...` when I checked.

A cache policy that hashes the dbt project would fix it, or rejecting `cache="auto"` until there is one. It's also a little backwards that `--project-dir` in `cli_args` is part of the cache key while the project's contents aren't.

plugins/dbt/src/flyteplugins/dbt/task.py
```diff
@@ -0,0 +1,75 @@
...
+            interface=NativeInterface(
+                inputs={"cli_args": (list[str], inspect.Parameter.empty)},
+                outputs={"results": list[DbtNodeResult]},
+            ),
+            task_type="dbt",
+            _call_as_synchronous=True,
+            parent_env=weakref.ref(task_environment) if task_environment else None,
+            parent_env_name=task_environment.name if task_environment else None,
```
samhita-alla, line 52: `task_environment` only feeds these two fields, which has two side effects.

The task is never added to the environment, so `flyte deploy` skips it. An env with one `@env.task` and one `DbtTask` lists only the `@env.task` in `env.tasks`. `flyte.run(main)` still works because the sub-action spec is sent inline at submit time, but the dbt task can't be deployed on its own, launched by name, or given a trigger. papermill's `NotebookTask` handles this with `task_environment._tasks[name] = self`.

None of the environment's settings carry over either. With `TaskEnvironment(name="dbt", resources=..., env_vars=..., secrets=[flyte.Secret(key="SNOWFLAKE_PASSWORD")])`, a regular `@env.task` inherits all three, but the `DbtTask` ends up with `resources`, `env_vars`, and `secrets` all `None`, since serialization reads them off the task. Putting warehouse credentials on the environment is a pretty natural setup, and here the pod would start without the secret and dbt would fail with a profile error that doesn't point at the real cause. `env.task` copies these onto the template in `_task_environment.py`, so doing the same here would line it up with regular tasks.

plugins/dbt/src/flyteplugins/dbt/task.py
```diff
@@ -0,0 +1,75 @@
...
+
+if TYPE_CHECKING:
+    from flyte import TaskEnvironment
+
+
+def _dbt_task_placeholder(*args: Any, **kwargs: Any) -> list[DbtNodeResult]:
+    raise NotImplementedError
+
+
+@dataclass(kw_only=True)
```
samhita-alla, line 25: `override()` currently raises on a `DbtTask`:

```python
DbtTask(name="dbt-build").override(retries=3)
# TypeError: flyte._task.AsyncFunctionTaskTemplate.__init__() got multiple values for keyword argument 'interface'
```

`override()` is implemented with `dataclasses.replace`, which calls `DbtTask(**all_init_fields)`. Those land in `**kwargs` and collide with the `func`, `interface`, `task_type`, `task_resolver`, and the rest that `__init__` passes explicitly. And since `callbacks` and `include_default_callback` aren't dataclass fields, `replace` would silently drop them even once the collision is fixed.

Making those two real dataclass fields, and letting `__init__` accept the fixed fields when `replace` hands them back, should make `override()` round-trip. A test that calls `override()` and checks the callbacks survive would keep it that way.

plugins/dbt/src/flyteplugins/dbt/runner.py
```diff
@@ -0,0 +1,235 @@
...
+    event_callbacks = []
+    if include_default_callback:
+        event_callbacks.append(_make_on_event_callback())
+    event_callbacks.extend(_make_on_event_callback(callback) for callback in resolve_callbacks(callbacks))
+
+    runner_result = dbtRunner(callbacks=event_callbacks).invoke(args)
+    if not runner_result.success:
+        if runner_result.exception is not None:
+            raise runner_result.exception
+        raise RuntimeError(f"dbt invocation failed for cli_args={args!r}")
```
samhita-alla, line 234: When a model or test fails, dbt returns `success=False` with `exception=None` and a fully populated `result`. `dbtRunner.invoke` catches its own `ResultExit` and returns it that way. So the everyday failure case lands on the `RuntimeError` branch, and the user sees:

```
RuntimeError: dbt invocation failed for cli_args=['build', '--project-dir', 'jaffle_shop', ...]
```

That says nothing about which node failed, even though `summarize_dbt_runner_result(runner_result)` has exactly that one line below. Could we build the message from the failed nodes (unique_id, status, message) and raise something like `flyte.errors.RuntimeUserError`? The flytekit dbt plugin also had an `ignore_handled_error` flag that returned results on model/test failures instead of raising, which might be worth bringing over since partial failures are common with `dbt build`. The current failure test only covers the `exception is not None` path.

samhita-alla (commented):
got my claude to review the PR as i don't have full context on how DBT works. review and incorporate the suggestions as you see fit.

plugins/dbt/src/flyteplugins/dbt/runner.py
```diff
@@ -0,0 +1,235 @@
...
+
+    status = _node_info_value(node_info, "node_status") or getattr(data, "status", None)
+    if status is None:
+        return
+
+    _record_dbt_node_status(
+        str(node_name),
+        _stringify_status(status),
+        trace_name=f"{status}.{node_name}",
+    )
```
samhita-alla, line 156: This one's more of a design question. On remote runs every `NodeFinished` event becomes a full trace action, and each one uploads inputs, uploads outputs, and blocks on `submit_and_wait_for_action` before returning, all on dbt's worker thread. For a project with 2,000 nodes across models, tests, and seeds, that's about 4,000 object-store writes and 2,000 trace actions for one task, with dbt's threads stalling on every round trip. If the watch stream degrades, each trace waits up to 60s (`_F_TRACE_COMPLETION_TIMEOUT`) before moving on.

The traced function doesn't do any work a resumed run could skip, since dbt reruns the whole command anyway, so the durability isn't buying much. One option is a single trace or report at the end built from the `list[DbtNodeResult]` you already produce. Another is making per-node traces opt-in. Right now `include_default_callback=False` is the only switch, and it's all or nothing.

plugins/dbt/src/flyteplugins/dbt/runner.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+def import_callback(import_path: str) -> DbtEventCallback:
+    module_name, _, attr_path = import_path.rpartition(".")
+    if not module_name or not attr_path:
+        raise ValueError(f"Invalid dbt callback import path: {import_path!r}")
+
+    value: Any = importlib.import_module(module_name)
+    for attr in attr_path.split("."):
+        value = getattr(value, attr)
```
samhita-alla, line 175: `rpartition(".")` splits on the last dot, so a path that `callback_import_path` itself produces for a method, like `cbmod.Callbacks.on_event` from a `@staticmethod`, becomes module `cbmod.Callbacks` plus attribute `on_event`, and `import_module` raises `ModuleNotFoundError: No module named 'cbmod.Callbacks'`. It passes validation locally and only fails in the container. The `attr_path.split(".")` loop below never sees more than one segment, which makes me think the intent was to support dotted attributes. Serializing as `module:qualname`, the entry-point convention, would make the split unambiguous, and plain dotted strings from users could keep today's behavior.

plugins/dbt/src/flyteplugins/dbt/runner.py
```diff
@@ -0,0 +1,235 @@
...
+    if raw_results is None:
+        return []
+
+    if hasattr(raw_results, "results"):
+        raw_results = raw_results.results
+
+    if isinstance(raw_results, tuple):
+        raw_results = list(raw_results)
+    elif not isinstance(raw_results, list):
+        raw_results = []
```
samhita-alla, line 69: For `dbt ls` and `dbt list`, `result` is a `List[str]` (it's in the type hint on `dbtRunnerResult.result`), so it passes the list check and each string goes through `_summarize_node_result`. Every `getattr` misses, and `cli_args=["ls", ...]` returns one `DbtNodeResult(unique_id='', name='', resource_type='', status='', ...)` per node, with the node names dropped. Since the README pitches one task for any dbt command, it might be worth only summarizing result objects that look like node results, and returning `[]` or raising a clear error otherwise.

samhita-alla:
One idea for dbt's artifacts. Right now the manifest, run results, and catalog stay in the container and disappear when the pod exits. They map nicely onto Flyte artifacts, which cover most of what dbt teams use them for:

- State for slim CI. A production build publishes its manifest, and a CI run fetches the latest one and passes it to `--state`, so only changed models get built.
- Triggers. An `OnArtifact` trigger can start downstream work whenever a new manifest lands, like retraining a model after the warehouse refreshes.
- Lineage and search. Each artifact records the run that produced it, and attrs like project, target, and dbt version make it filterable.
- A card. Artifacts can carry an HTML card, and `dbt docs generate --static` produces a single self-contained docs page that would fit well there.

Artifacts only work on Union, not open-source Flyte, so publishing should be opt-in. Returning the files as a plain `Dir` output works on both, though, and on open-source Flyte that's what you'd pass to a later run's `--state`. Something like:

```python
dbt_prod = DbtTask(name="dbt-prod", task_environment=env, artifact_name="jaffle-shop-prod")
dbt_ci = DbtTask(name="dbt-ci", task_environment=env)

results, target = await dbt_prod.aio(cli_args=["build", "--target", "prod"])

# CI on Union: build only what changed since the last production run
state = flyte.remote.Artifact.get("jaffle-shop-prod")  # latest version
flyte.run(dbt_ci, cli_args=["build", "--select", "state:modified+", "--defer"], state=state)
```

A few things that shape it:

- An artifact has to be a top-level output, so the task would return `(results, target_dir)`. That changes the return type, so it's worth settling the output shape before merge even if publishing lands in a follow-up.
- The plugin should point dbt's target path at a directory it controls and return just the manifest, run results, and catalog or sources when present. The full `target/` also holds compiled SQL and parse caches, which get big.
- Consuming state needs a `Dir` input on the task, since a path string from the caller won't exist inside the dbt task's container.
- Artifacts are only extracted when the task succeeds, and right now one failed model raises. A failed run is exactly the one `dbt retry` needs, so this depends on returning results with a failure status instead of raising (see the failure-path comment).
- The resolver would need to carry the new setting, so the container-side task declares the same outputs as the parent.

samhita-alla:
On `cli_args: list[str]`: passing dbt's command line straight through is flexible, but the plugin never learns where the project is. That's what blocks a cache policy that hashes the project, and it's what makes returning the target directory awkward. It also puts paths like `--project-dir` into the cache key and the action name, so moving the project inside the image changes both for the same logical work. And every call repeats the same flags, like the example's `dbt_project_args` helper.

Would it make sense to split environment from invocation? Something like:

```python
dbt = DbtTask(name="dbt", task_environment=env, project_dir="jaffle_shop", profiles_dir="dbt-profiles")

await dbt.aio(command="build", select=["stg_orders+"], target="prod", extra_args=["--full-refresh"])
```

With the project known up front, the plugin can fail fast with a clear message when `dbt_project.yml` is missing, reject flags it manages if they show up in `extra_args`, and own `--target-path`. The flytekit dbt plugin had a similar shape, with `project_dir`, `profiles_dir`, `profile`, and `target` as typed fields.

One validation gap this wouldn't fix on its own: a typo in the selection makes dbt succeed with nothing selected, since "Nothing to do" is only a warning. The task then returns an empty list as a success. Failing when a run or build selects zero nodes, or turning that warning into an error through dbt's warn-error options, would catch it.

plugins/dbt/src/flyteplugins/dbt/runner.py
fiedlerNr9: Fixed in this commit: https://github.com/flyteorg/flyte-sdk/pull/1558/changes/1d9528fd7b21c93017dc8291b3096996cd5a6cc8

plugins/dbt/src/flyteplugins/dbt/task.py
fiedlerNr9: I made dbt task now a RunTimeTaskTemplate. All the task configs like cache, retries etc apply now.

In this commit: https://github.com/flyteorg/flyte-sdk/pull/1558/changes/df214fae29cc5df4ff4010d07ab7e8451ebbe7b0

plugins/dbt/src/flyteplugins/dbt/task.py
fiedlerNr9: Good catch. Fixed in this commit: https://github.com/flyteorg/flyte-sdk/pull/1558/changes/3f0e2da9cde16dc398e2649294bcd15b24f91673

All the env settings are now inherited

plugins/dbt/src/flyteplugins/dbt/task.py
fiedlerNr9: Fixed by making it a RunTimeTaskTemplate

plugins/dbt/src/flyteplugins/dbt/runner.py
fiedlerNr9: You are right, we can give better error messages here. Incorporated your feedback in this commit: https://github.com/flyteorg/flyte-sdk/pull/1558/changes/83f52f70ad1ddf0a9035a7087a68613311e4ba5b

plugins/dbt/src/flyteplugins/dbt/runner.py
fiedlerNr9: There is a `trace_node_events` flag that you can turn off and on now.

https://github.com/flyteorg/flyte-sdk/pull/1558/changes/3404260511e38b2b131d36d1f9ecfc693a2d35cb

fiedlerNr9:
> One idea for dbt's artifacts. Right now the manifest, run results, and catalog stay in the container and disappear when the pod exits. They map nicely onto Flyte artifacts, which cover most of what dbt teams use them for:
> 
> * State for slim CI. A production build publishes its manifest, and a CI run fetches the latest one and passes it to `--state`, so only changed models get built.
> * Triggers. An `OnArtifact` trigger can start downstream work whenever a new manifest lands, like retraining a model after the warehouse refreshes.
> * Lineage and search. Each artifact records the run that produced it, and attrs like project, target, and dbt version make it filterable.
> * A card. Artifacts can carry an HTML card, and `dbt docs generate --static` produces a single self-contained docs page that would fit well there.
> 
> Artifacts only work on Union, not open-source Flyte, so publishing should be opt-in. Returning the files as a plain `Dir` output works on both, though, and on open-source Flyte that's what you'd pass to a later run's `--state`. Something like:
> 
> ```python
> dbt_prod = DbtTask(name="dbt-prod", task_environment=env, artifact_name="jaffle-shop-prod")
> dbt_ci = DbtTask(name="dbt-ci", task_environment=env)
> 
> results, target = await dbt_prod.aio(cli_args=["build", "--target", "prod"])
> 
> # CI on Union: build only what changed since the last production run
> state = flyte.remote.Artifact.get("jaffle-shop-prod")  # latest version
> flyte.run(dbt_ci, cli_args=["build", "--select", "state:modified+", "--defer"], state=state)
> ```
> 
> A few things that shape it:
> 
> * An artifact has to be a top-level output, so the task would return `(results, target_dir)`. That changes the return type, so it's worth settling the output shape before merge even if publishing lands in a follow-up.
> * The plugin should point dbt's target path at a directory it controls and return just the manifest, run results, and catalog or sources when present. The full `target/` also holds compiled SQL and parse caches, which get big.
> * Consuming state needs a `Dir` input on the task, since a path string from the caller won't exist inside the dbt task's container.
> * Artifacts are only extracted when the task succeeds, and right now one failed model raises. A failed run is exactly the one `dbt retry` needs, so this depends on returning results with a failure status instead of raising (see the failure-path comment).
> * The resolver would need to carry the new setting, so the container-side task declares the same outputs as the parent.

I really like this idea but lets get the plugin out first and see how it will be used

fiedlerNr9:
> On `cli_args: list[str]`: passing dbt's command line straight through is flexible, but the plugin never learns where the project is. That's what blocks a cache policy that hashes the project, and it's what makes returning the target directory awkward. It also puts paths like `--project-dir` into the cache key and the action name, so moving the project inside the image changes both for the same logical work. And every call repeats the same flags, like the example's `dbt_project_args` helper.
> 
> Would it make sense to split environment from invocation? Something like:
> 
> ```python
> dbt = DbtTask(name="dbt", task_environment=env, project_dir="jaffle_shop", profiles_dir="dbt-profiles")
> 
> await dbt.aio(command="build", select=["stg_orders+"], target="prod", extra_args=["--full-refresh"])
> ```
> 
> With the project known up front, the plugin can fail fast with a clear message when `dbt_project.yml` is missing, reject flags it manages if they show up in `extra_args`, and own `--target-path`. The flytekit dbt plugin had a similar shape, with `project_dir`, `profiles_dir`, `profile`, and `target` as typed fields.
> 
> One validation gap this wouldn't fix on its own: a typo in the selection makes dbt succeed with nothing selected, since "Nothing to do" is only a warning. The task then returns an empty list as a success. Failing when a run or build selects zero nodes, or turning that warning into an error through dbt's warn-error options, would catch it.

I like this a lot. Incorporated here -> https://github.com/flyteorg/flyte-sdk/pull/1558/changes/63d111f8d75dd60c7ddc8150380f11f93731a22e

PR #1562, opened 2026-09-09
feat(app): llama.cpp serving with download / HF / object-store-mount delivery

mhotan:
## Intent
Add `flyteplugins.llamacpp.LlamaCppAppEnvironment` — a Flyte App that serves GGUF models with llama.cpp (OpenAI-compatible `/v1` + Web UI), where llama.cpp fits and vLLM/SGLang don't: quantized GGUF, partial CPU offload, CPU-only serving.

## Delivery model
The weights reach the server one of three mutually-exclusive ways (exactly one):
- **`model_path`** — download an artifact/URI into the pod (default). Binding an `ArtifactValue` records the App→artifact lineage edge.
- **`model_hf_path`** — llama-server pulls from a HuggingFace GGUF repo at startup.
- **`mount=ObjectStoreMount(...)`** — read the weights **in place** from a read-only, object-store-backed PVC (gcsfuse on GKE, Mountpoint-S3 on EKS): no copy, lazy first-touch, releases cleanly on scale-to-zero.

`ObjectStoreMount` groups the mount config (`pvc`, `mount_path`, `model_path`, `draft_model_path`); its **presence is the trigger** — there is no `model_delivery` enum. `mount.model_path` accepts an `ArtifactValue`/`RunOutput` served directly (resolved URI + lineage) *or* a literal subpath string.

## Design notes
- **No `flyte.app` core surface.** An earlier revision put a `ModelDeliveryMixin` in `flyte.app` for cross-engine reuse; vLLM/SGLang stream weights instead and never adopted it, so it had a single consumer. It's collapsed into the plugin and `src/flyte/app/_model_delivery.py` is deleted — removing a speculative public API and the flyte-core version coupling it introduced.
- **`pod_annotations`** is a general (any-mode) knob; the GKE gcsfuse sidecar annotation (`gke-gcsfuse/volumes: "true"`) is set there explicitly.
- Speculative decoding via `draft_model_path` / `draft_model_hf_path` / `mount.draft_model_path`.

## Test Plan
`ty`, `mypy`, `make fmt`, `make check-docstrings` all green; **62 plugin unit tests pass** (`plugins/llamacpp/tests/test_app_environment.py`), covering all three delivery modes, the artifact-over-mount lineage path, `clone_with`, and the pod-template mount attach. Serving validated end-to-end on managed + self-hosted dataplanes (see the example PRs above in the stack).

## Stack
`main` ← #1528 ← #1531 ← **#1562 (this)** ← #1563 ← #1571 ← #1569.


src/flyte/app/_runtime/app_serde.py
```diff
@@ -396,6 +403,7 @@ async def translate_app_env_to_idl(
             app_env,
             app_env.pod_template,
             serialization_context,
+            parameter_overrides=parameters,
```
mhotan, line 406: Pretty sure this was a bug.

mhotan:
Draft: needs commit `996f62f1b` (`docs(llamacpp): reuse the model artifact if it exists; LLAMACPP_FORCE_PREFETCH to re-create`) folded in here. It's a model-delivery/prefetch change that currently sits on #1569 but belongs in this PR.

PR #1569, opened 2026-09-11
docs(llamacpp): union_volume example serves via the uvol mount broker (allow_volumes)

mhotan:
## Intent
Serve a GGUF Model artifact with llama.cpp from a **Union Volume (JuiceFS)**, mounted via the **uvol mount broker** — the zero-privilege delivery mode (inline ephemeral CSI `volumes.union.ai`), as opposed to the read-only object-store PVC in `llamacpp_app_fuse.py`. Repurposes the former `union-volume-exec` PR around the broker path now that it's Knative-serving-viable.

## What changed
- **Rename** `llamacpp_app_union_fuse.py` → `llamacpp_app_union_volume.py` (the name now reflects the Union Volume mechanism, and no longer collides with the object-store `llamacpp_app_fuse.py`).
- **Switch the FUSE mechanism** from the `fuseDevicePlugin` device-plugin (`flyte.PodTemplate().allow_fuse()`: CAP_SYS_ADMIN + `smarter-devices/fuse`) to the **uvol mount broker** (`flyteplugins.union.io.allow_volumes()`: inline ephemeral CSI, broker premounts the FUSE channel and hands the fd over a socket — no CAP_SYS_ADMIN, no `/dev/fuse`, no hostPath, no device-plugin). Applied to both the volume-builder task and the serve App.
- **Drop `fuse3`** from both images (the broker premounts, so the container issues no `mount(2)`).
- **Fix**: the serve App passes a base `flyte.PodTemplate(primary_container_name="app")` to `allow_volumes(...)` so the App-serde's `primary_container_name` check passes.
- Flow is unchanged: `flyte.prefetch.hf_model` → Model artifact → `build_model_volume` (RWVolume → copy artifact → `commit()` → ROVolume locator) → serve via `union-volume-exec` + `build_fserve_command`.

## Also adds
- **`llamacpp_sidecar_union_volume.py`** — a task-pod counterpart to the App: a llama.cpp server as a **native sidecar** (client `primary` + `llama` sidecar, ephemeral or `--reuse` actor, from `llamacpp_sidecar.py`) that serves the model **from the same Union Volume over the broker** (`union-volume-exec`, from this PR's App). The broker wiring is sourced from `allow_volumes()` and attached to the **sidecar** container; all `flyteplugins.*` imports stay lazy so the client container (flyte + openai only) never imports them. A task pod does not traverse Knative, so this variant needs **none** of the Knative volume feature flags the App variant requires.

## How it was tested
Verified **end-to-end on dogfood** (managed, Knative 1.23 gateway + uvol broker): build_model_volume mounted the RWVolume over the broker and committed the ROVolume; the serve pod mounted the ROVolume read-only via the broker (`fuse.uvol` at `/tmp/models`), and llama-server returned a completion (`{"content":"Hello!"}`).

## Deploy dependencies
- App-mode serving of Union Volumes depends on the dataplane serving-gateway update in unionai/cloud#18409. (The task-pod sidecar variant has no such dependency.)
- Requires `flyteplugins-union>=0.11.0b0` (on PyPI); the image builds pin it. Relax the pin once a stable 0.11.x ships.

## Stack
base ← #1531 ← #1562 ← #1563 ← #1571 ← **#1569**


mhotan:
I had to  move this back to draft because have some ongoing changes that remove union-device-plugin to be replaced by uvol broker https://github.com/unionai/helm-charts/pull/585

PR #1570, opened 2026-09-11
feat(app): add cluster to pin an app to a specific cluster

mhotan:
## Summary

Adds a `cluster` field to `AppEnvironment` and a `cluster` argument to `serve()`, setting app `Spec.cluster` so a user can **pin a serving app to a named cluster** instead of load-balancing across a `cluster_pool`.

Motivation: today app placement is `cluster_pool` → first-healthy-cluster-that-pulls (non-deterministic, no capacity awareness). Users who need a serving app on a *specific* cluster (e.g. the one with a particular accelerator) had no way to express it.

## Behavior
- `cluster` and a **non-default** `cluster_pool` are **mutually exclusive** — raises `ValueError` at `AppEnvironment` construction and in `serve()`.
- When `cluster` is set, the serialized `Spec` sends an **empty `cluster_pool`** so the control plane sees an unambiguous pin.
- `cluster` unset → unchanged behavior (pool placement; empty pool → default pool).

## Depends on
flyteorg/flyte#8018 (adds `app.Spec.cluster` to `flyteidl2`). The control plane honors the pin (leases only to the named cluster) and validates mutual exclusion.


mhotan:
### Cross-Stack Review: App Pinning (`Spec.Cluster`)

#### Plan & Design Adherence
- `AppEnvironment` parameters and mutual exclusion validation between `cluster` and non-default `cluster_pool` are cleanly implemented.
- `app_serde.py` correctly translates `app_env.cluster` and clears `cluster_pool = ""` to signal an unambiguous pin to the control plane.
- All 239 app tests pass.

#### ⚠️ Control Plane Dependency Note:
Be aware that in `unionai/cloud#17548`, `AppServiceImpl.Create` resolves `spec.cluster` into `spec.cluster_pool = pool`, but `AppServiceImpl.Update` currently omits this resolution. As a result, updating an existing app via `_serve_remote` will result in `cluster_pool = ""` in the DB and stall the app in `PENDING` until the fix in `cloud#17548` lands (flagged in cloud#17548).

The SDK changes themselves look great and correctly follow the placement spec.

mhotan:
Closing per the multi-cluster app-serving decision (team thread): apps are placed via the existing project/domain → cluster-pool routing (the same path runs use) and the URL is composed from the assigned cluster, rather than an explicit SDK/IDL cluster pin. For a single-cluster pool this is deterministic. We're keeping the addition minimal so cluster-pool + resource-aware scheduling can mature before committing to a new SDK interface. The App Spec.cluster proto field is being removed (flyteorg/flyte #8018), and cloud honors routing-based placement (unionai/cloud #17548). No SDK `cluster=` field for now.

mhotan:
Reopening — the team decided to keep the explicit `cluster` pin after all (multi-cluster cluster pools will be needed, so routing-to-single-cluster-pools doesn't generalize). No code change on this branch; it already carries the `cluster=` field on AppEnvironment/serve. Depends on flyteidl2 carrying `app.cluster` (flyteorg/flyte#8018) + cloud honoring it (unionai/cloud#17548).

kumare3 (approved):

EngHabu (approved):

PR #1598, opened 2026-09-17
fix: ClusteredTaskEnvironment registers empty custom via flyte deploy

AdilFayyaz:
## Motivation

  flyte deploy registered every ClusteredTaskEnvironment task with an empty custom, so each run failed in the backend with [BadTaskSpecification] invalid ClusteredTaskSpec: nil Struct Object passed after ~34 minutes of retries. ClusteredTaskTemplate read its settings back through the parent_env weakref, which copy.copy (used by _deploy._with_local_sys_paths, on by default) and cloudpickle (notebook bundles) both drop. The same copy also blanked TaskSpec.environment.name for every deployed task since v2.5.14.

##  Summary

  - ClusteredSettings: a frozen snapshot of the clustered fields built in ClusteredTaskEnvironment.__post_init__ and carried on _ClusteredPlugin.settings; custom_config/container_args read it instead of parent_env, so custom and the --runtime=jax arg survive copy and pickle. A settings-less marker (the in-container routing case) now raises a clear RuntimeUserError if it ever reaches serialization instead of registering {}.
  - TaskTemplate.__copy__: a shallow copy keeps parent_env; the pickle hooks are unchanged. Restores environment.name/description on deployed specs (cloud persists the name and groups actions on it).
  - _run.py: run-level _F_SYS_PATH is emitted only for ephemeral task specs. For a deployed target (TaskDetails, base.task_spec_source == DEPLOYED, or the trigger-fire path) the registered spec already carries the deploy-time value; re-sending the launcher's copy was the wrong source and duplicated the key, which JobSet (a CRD with env as a name-keyed map) rejects with a 422. Deployed reruns also strip the inherited key.
  - Tests: copy and cloudpickle round-trips keep custom/runtime flag; bare marker fails loudly; clone_with rebuilds settings; deploy dry-run keeps environment.name;  _apply_overrides gate cases (ephemeral / deployed / deployed base / ephemeral base / explicit).

 ## Test Plan

  - uv run pytest tests/flyte/clustered tests/flyte/test_deploy.py tests/flyte/test_task_copy.py tests/flyte/test_run_runspec_chars.py tests/flyte/test_rerun.py tests/flyte/test_run_trigger.py tests/user_api/test_run_python_script.py — 249 passed.
  - Repro from the issue: on main, copy.copy(f).custom_config(None) returns {}; on this branch it matches f.custom_config(None).
  - Dogfood, default settings (no --no-sync-local-sys-paths): flyte deploy --version eng26-1349-t1 examples/clustered/ddp_train.py env registers custom={replicas:2, nprocPerNode:1, runtime:{torchrun:{}}, failurePolicy:{maxRestarts:1}} with environment.name='ddp_env' and _F_SYS_PATH in the container env (the previously failing copy path ran).
  - flyte run deployed-task ddp_env.train_ddp:eng26-1349-t1 --steps 50 → run ud6zq7ssg47lwj8vx6pz SUCCEEDED, 1 attempt, 2×L4 DDP; runSpec.envs has no _F_SYS_PATH. Same run with the SDK's previous behavior (ut9x7v8g6bj7swtvzfrl) failed in 43 ms on the duplicate env name

## Issue 

Related to : https://linear.app/unionai/issue/ENG26-1349/flyte-deploy-fails-with-clustered-task-environments


kumare3:
Verified the root cause end to end against current `main` — `copy.copy()` falling through to the pickle `__getstate__`/`__setstate__` hooks and dropping `parent_env` is real and exactly as described, and the `ClusteredSettings` snapshot is the right fix for the cloudpickle side (a weakref can't survive pickling no matter what `__copy__` does, so avoiding the dependency entirely is correct, not a workaround). Tests hold up — `test_custom_survives_copy`, `test_runtime_flag_survives_copy`, and the `environment.name` assertion in `test_deploy.py` all genuinely fail on unpatched `main`. No must-fixes in the code.

Two things before merge:

- `src/flyte/_run.py:58-64` — the `deployed_target` inference falls back to "ephemeral" (re-injects `_F_SYS_PATH`) whenever `base.task_spec_source == TASK_SPEC_SOURCE_UNSPECIFIED`, not just when it's explicitly `EPHEMERAL`. This PR is the only place in the SDK that reads this field — can you confirm the backend has always populated `task_spec_source` on every `RunSpec`? If there's a rollout gap, a rerun of an old deployed run could still hit the exact duplicate-`_F_SYS_PATH` rejection this PR fixes.
- Heads-up, not a blocker: this conflicts with #1592 (the `ClusteredTaskEnvironment` → `MultiNodeTaskEnvironment` rename) in `clustered/_task.py` and `clustered/_environment.py`. Recommend merging this one first — #1592's `_task.py` hunk retypes the `parent_env`/`cast` code this PR deletes outright, so landing #1592 first leaves that hunk silently inapplicable rather than a clean rebase.

Approving pending an answer on the `task_spec_source` question — happy to merge right after.

AdilFayyaz:
@kumare3 thanks for the review. on cloud `task_spec_source` is always populated. On OSS it's never set, i'll gate it in the SDK so that when the source is `UNSPECIFIED`, we check whether the task spec being rerun already carries `_F_SYS_PATH` (only flyte deploy bakes it in, so it's an exact marker of a deployed spec). Cloud behavior is unchanged. 

On #1592: agreed, ill merge this first.

kumare3 (approved):

